Issue #17, opened 2022-01-27
Add more example code


zhenfeizhang:
should we add more example code? both for downstream user, and for potential contributor to this library?

_Originally posted by @alxiong in https://github.com/SpectrumXYZ/jellyfish/pull/13#discussion_r793546028_

PR #27, opened 2022-02-22
Example: proof of knowledge of exponent

zhenfeizhang:
## Description

part of #17 

---

write a little note why.

- [x] Wrote unit tests


plonk/examples/proof_of_exp.rs
```diff
@@ -0,0 +1,132 @@
...
+//
+// We write the code with generics so that is can be adapted to
+// multiple curves.
+// Specifically, the PoE is associated with an embedded curve with param `P`,
+// that defined twisted-edwards parameters for a curve.
+#[allow(non_snake_case)]
+fn proof_of_exponent_circuit<P>(
+    x: P::ScalarField,
+    X: TEAffine<P>,
+) -> Result<PlonkCircuit<P::BaseField>, PlonkError>
```
alxiong, line 86: can I suggest we change this function to be generic over two parameters (one for twisted edward curve, one for Bls12-381), reason being people's default mindset would be getting a circuit over a scalar field (but here we return a `PlonkCircuit<P::BaseField>`, not immediately clear that the `P` here is for the `TEParam`. 

```rust
fn proof_of_exponent_circuit<P, P2>(
    x: P::ScalarField,
    X: TEAffine<P>,
) -> Result<PlonkCircuit<P2::ScalarField>, PlonkError>
where
    P: TEModelParameters + Clone,
    P2: SWModelParameters<ScalarField = P::BaseField>,
    <P as ModelParameters>::BaseField: PrimeField,
```

of course, renaming `<P, P2>` to `<EmbedParam, CurveParam>` or better name is definitely good.

alxiong (changes requested):

plonk/examples/proof_of_exp.rs
```diff
@@ -0,0 +1,132 @@
...
+    // - public generator `G`;
+    // - public group element `X := xG`
+    // This circuit does not need to have real inputs.
+    // We can simply use a dummy data set.
+    let circuit = proof_of_exponent_circuit(x, X)?;
+
+    // Knowing the circuit size, we are able to simulate the universal
+    // setup and obtain the structured reference string (SRS).
+    //
+    let circuit_size = circuit.num_gates() + 100; // adding 100 as a buffer
```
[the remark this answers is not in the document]
zhenfeizhang: I think this 2 gates are added when the circuit is build, isn't it? @chancharles92

plonk/examples/proof_of_exp.rs
zhenfeizhang: sounds good.

zhenfeizhang: So here is the new one
``` rust
// Specifically, the PoE is associated with
// - an embedded curve with param `EmbedCurve` that defined twisted-edwards
//   parameters for a curve
// - a pairing engine
// - the native field F for the prove system
#[allow(non_snake_case)]
fn proof_of_exponent_circuit<EmbedCurve, PairingCurve, F>
```

plonk/examples/proof_of_exp.rs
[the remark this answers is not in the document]
alxiong: > if there is a way to systematize this though

ah! we should use `circuit.eval_domain_size()` instead

plonk/examples/proof_of_exp.rs
alxiong: is the `F` necessary?

zhenfeizhang: F is to ensure that this is native field evaluation. It is not really `necessary` the same way as `PairingCurve`, but is nice to have IMHO. WDYT? 

``` rust
where
    F: PrimeField,
    EmbedCurve: TEModelParameters<BaseField = F> + Clone,
    PairingCurve: PairingEngine<Fr = F>,
```

alxiong: Okay, sure, make sense. 

(my only minor concern is that my suggestion is making this function looks more complicated than it was, if we were to have a non-generic example, it would looks much simpler which might be a plus for example code)

plonk/examples/proof_of_exp.rs
```diff
@@ -0,0 +1,142 @@
...
+    // - public group element `X := xG`
+    // This circuit does not need to have real inputs.
+    // We can simply use a dummy data set.
+    let circuit = proof_of_exponent_circuit::<EdwardsParameters, Bls12_381>(x, X)?;
+
+    // Knowing the circuit size, we are able to simulate the universal
+    // setup and obtain the structured reference string (SRS).
+    //
+    // The required SRS size can be obtained from the circuit.
+    let srs_size = circuit.eval_domain_size()?;
```
chancharles92, line 52: Should it be `eval_domain_size()` or `eval_domain_size() + 2`?

plonk/src/circuit/basic.rs
```diff
@@ -261,6 +261,12 @@ impl<F: FftField> PlonkCircuit<F> {
     pub fn range_size(&self) -> Result<usize, PlonkError> {
         Ok(1 << self.range_bit_len()?)
     }
+
+    /// The required SRS size for the circuit.
+    pub fn srs_size(&self) -> Result<usize, PlonkError> {
+        // extra 2 degree for masking polynomial to make snark zero-knowledge
+        Ok(self.num_gates() + 2)
+    }
```
alxiong, line 269: - I personally don't know which one of `srs_size()` or `max_degree()` is a better name
- should we add this to `trait Arithmetization`?
- We should also update code here: https://github.com/SpectrumXYZ/jellyfish/blob/main/plonk/src/proof_system/snark.rs#L76-L81

(cc @chancharles92 for input)

zhenfeizhang: > We should also update code here: https://github.com/SpectrumXYZ/jellyfish/blob/main/plonk/src/proof_system/snark.rs#L76-L81

Yeah... seems that that block of code needs its own API too -- input an SRS and a circuit, return whether if the srs is sufficient for that circuit.

plonk/src/circuit/basic.rs
```diff
@@ -261,6 +261,12 @@ impl<F: FftField> PlonkCircuit<F> {
     pub fn range_size(&self) -> Result<usize, PlonkError> {
         Ok(1 << self.range_bit_len()?)
     }
+
+    /// The required SRS size for the circuit.
+    pub fn srs_size(&self) -> Result<usize, PlonkError> {
+        // extra 2 degree for masking polynomial to make snark zero-knowledge
+        Ok(self.num_gates() + 2)
```
chancharles92, line 268: Should this be `self.eval_domain_size() + 2`?

plonk/src/circuit/basic.rs
chancharles92: Agree with @alxiong on point 2 and 3, for point 1, I'm fine with either name.

@zhenfeizhang I don't think it's necessary to have a new API, maybe just write:
``if srs.0.max_degree() < circuit.srs_size() {
            return Err(PlonkError::IndexTooLarge);
   }``

plonk/src/proof_system/snark.rs
```diff
@@ -76,7 +76,7 @@ where
         let domain_size = circuit.eval_domain_size()?;
         let num_inputs = circuit.num_inputs();
         let supported_hiding_bound = 2;
```
chancharles92, line 78: Is `supported_hiding_bound` still used in the function? We can remove it if not.

plonk/src/proof_system/snark.rs
```diff
@@ -76,7 +76,7 @@ where
...
-        if srs.0.max_degree() < domain_size + supported_hiding_bound {
+        if srs.0.max_degree() < circuit.srs_size()? {
```
chancharles92, line 79: In line 99:
`let (commit_key, open_key) = trim(&srs.0, domain_size + supported_hiding_bound);`

Maybe we can define variable `srs_size = circuit.srs_size()` first, and then replace `domain_size + supported_hiding_bound` with `srs_size`, then we don't need to define `supported_hiding_bound` anymore.

chancharles92 (approved):

alxiong (approved):
LGTM as well!

PR #30, opened 2022-02-25
Derive Snafu for TaggedBlobError.

jbearer:
This implements Display, which allows us to take TaggedBlob types on the command line using structopt.


alxiong (approved):
LGTM!

alxiong:
@jbearer Can we just manually implement `Display` without depending on `Snafu`? That would be sufficient for the downsteam?

PR #32, opened 2022-03-01
micro benchmarks

zhenfeizhang:
## Description

closes: #XXXX

---

write a little note why.

This PR adds micro benchmarks for subroutines within proving and verification. It is enabled with `--features=bench`

- [ ] ~Wrote unit tests~

### DIFF
--- a/plonk/benches/bench.rs
+++ b/plonk/benches/bench.rs
@@ -5,7 +5,7 @@
 // along with the Jellyfish library. If not, see <https://mit-license.org/>.
 
 // For benchmark, run:
-//     RAYON_NUM_THREADS=N cargo bench
+//     RAYON_NUM_THREADS=N cargo bench --features bench
 // where N is the number of threads you want to use (N = 1 for single-thread).
 
 use ark_bls12_377::{Bls12_377, Fr as Fr377};
@@ -14,6 +14,7 @@ use ark_bn254::{Bn254, Fr as Fr254};
 use ark_bw6_761::{Fr as Fr761, BW6_761};
 use ark_ff::PrimeField;
 use jf_plonk::{
+    bencher::{init_timers, total_fft_time, total_msm_time, total_poly_eval_time},
     circuit::{Circuit, PlonkCircuit},
     errors::PlonkError,
     proof_system::{PlonkKzgSnark, Snark},
@@ -46,6 +47,8 @@ fn gen_circuit_for_bench<F: PrimeField>(
 
 macro_rules! plonk_prove_bench {
     ($bench_curve:ty, $bench_field:ty, $bench_plonk_type:expr, $num_gates:expr) => {
+        init_timers();
+
         let rng = &mut ark_std::test_rng();
         let cs = gen_circuit_for_bench::<$bench_field>($num_gates, $bench_plonk_type).unwrap();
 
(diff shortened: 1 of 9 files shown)


alxiong (commented):
Question: instead of declaring one bencher function for each (msm, fft, poly_eval), have you tried [`ark_std::start_timer` and `ark_std::end_timer`](https://github.com/arkworks-rs/std/blob/master/src/perf_trace.rs#L2-L9) (another [usage example](https://github.com/arkworks-rs/poly-commit/blob/982e1a1f8b5bce4e05516d3785507b35d41932f7/src/kzg10/mod.rs#L47-L136))?

they use concrete struct instance to avoid the necessity of locking? (need to double check)
Do your bencher profile more accurately under multi-threaded run?

zhenfeizhang:
Ha, good idea! I didn't know about their timer... I will switch to theirs...

zhenfeizhang:
Did some digging in the ark_std's timer. Those are macros that get you the running time __within__ a function. We need to benchmark FFTs and MSMs __across multiple functions__ so we will still need to declare global variables to store the result. Given this I am inclined to keep the current design.

plonk/benches/bench.rs
```diff
@@ -177,7 +299,28 @@ fn bench_batch_verify() {
     plonk_batch_verify_bench!(BW6_761, Fr761, PlonkType::UltraPlonk, 1000);
 }
 
+fn bench_intense() {
+    let mut f = File::create(format!(
+        "../target/{}-threads.txt",
+        rayon::current_num_threads()
+    ))
+    .expect("Unable to create file");
```
alxiong, line 307:
```suggestion
        let mut path = PathBuf::new();
        path.push(env!("CARGO_MANIFEST_DIR"));
        path.push(format!("target/plonkbench/{}-threads", rayon::current_num_threads()));
        path.set_extension("txt");
        
        let mut f = OpenOptions::new()
            .write(true)
            .create(true)
            .truncate(true)
            .open(path.clone())
            .unwrap();
```

You can try using something like this to overwrite the file instead of appending to it in case the file already exists (which [`File::create`](https://doc.rust-lang.org/std/fs/struct.File.html#method.create) does)

plonk/src/bencher.rs
```diff
@@ -0,0 +1,184 @@
+//! Helper functions for micro-benchmarks
+
+use ark_std::{thread_local, time::Instant};
+use core::{cell::RefCell, time::Duration};
+
+thread_local!(static FFT_START_TIME: RefCell<Instant> = RefCell::new(Instant::now()));
+thread_local!(static FFT_TIMER_LOCK: RefCell<bool> = RefCell::new(false));
+thread_local!(static FFT_TOTAL_TIME: RefCell<Duration> = RefCell::new(Duration::ZERO));
```
alxiong, line 8: I'm not sure if there's better method here, basically we are dealing with a _global time counter_ that accumulates time spent across code snippets in many functions. 

1. I wonder if we can get rid of the time lock, and start time -- they can be local variable, right? only the total time needs to be globally accessible.

plonk/src/bencher.rs
```diff
@@ -0,0 +1,184 @@
...
+            panic!("another FFT timer has already started somewhere else");
+        }
+
+        FFT_START_TIME.with(|timer| {
+            *timer.borrow_mut() = Instant::now();
+        });
+
+        FFT_TIMER_LOCK.with(|lock| {
+            *lock.borrow_mut() = true;
+        })
```
alxiong, line 91: this logic is slightly baffling to me, so if FFT is locked, then we panic?
1. it seem to me that our timer will never be used by 2 threads at the same time? because the multi-threaded code is in-between the `fft_start()` and `fft_end()` (e.g. inside `fn compute_selector_polynomials`), so maybe we don't need this locking to begin with?
2. If we have proper concurrency control, then FFT timer should never panic, but rather wait on lock to be released instead?

PR #42, opened 2022-03-10
Add serde and Display/FromStr for schnorr_dsa::KeyPair

jbearer:
I recently created a CLI utility for generating signing keys. I found that I wanted to print the private key using tagged base 64, so that it can be easily copied and pasted to RPass, for example. For this to work we need to add a `#[tagged_blob()]` annotation.

### DIFF
--- a/primitives/src/schnorr_dsa.rs
+++ b/primitives/src/schnorr_dsa.rs
@@ -118,6 +118,7 @@ impl<P: Parameters + Clone> VerKey<P> {
 
 /// Signature secret key pair used to sign messages
 // make sure sk can be zeroized
+#[tagged_blob("SIGNKEYPAIR")]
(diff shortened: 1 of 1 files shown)


primitives/src/schnorr_dsa.rs
```diff
@@ -118,6 +118,7 @@ impl<P: Parameters + Clone> VerKey<P> {
 
 /// Signature secret key pair used to sign messages
 // make sure sk can be zeroized
+#[tagged_blob("SIGNKEY")]
```
zhenfeizhang, line 121: Should the tag be `SIGNKEYPAIR` just in case we may need to tag the real `SIGNKEY` in future?

jbearer: Sure, I'll change it

zhenfeizhang (approved):

Issue #48, opened 2022-04-04
Lookup tables domain separation


chancharles92:
Use the technique in Section 4.1 of Plookup ([https://eprint.iacr.org/2020/315.pdf](url)) to support multiple lookup tables.

PR #47, opened 2022-04-02
a comment on lookup table creation api

chancharles92:
## Description

An important API comment: the key index of the lookup variable should be range-checked before calling `create_table_and_lookup_variable()`, otherwise one may set it outside the bounds of the array, and thus doing a lookup into one of the _other_ tables.

closes: #XXXX

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [ ] Targeted PR against correct branch (main)
- [ ] Linked to GitHub issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [ ] Wrote unit tests
- [ ] Updated relevant documentation in the code
- [ ] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md`
- [ ] Re-reviewed `Files changed` in the GitHub PR explorer

PR #51, opened 2022-04-05
Table domain separation

chancharles92:
## Description

closes: #48   (and a followup of #47 )

- Add an error path test: the circuit should be unsatisfiable if the element to be looked up is in a separate table. (Thanks @tri-joe for spotting it)
- Refactor the code to use Plookup paper's original approach for supporting multiple tables.
- Removed unnecessary range-check gates in MSM.

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/plonk/src/circuit/basic.rs
+++ b/plonk/src/circuit/basic.rs
@@ -314,29 +314,38 @@ impl<F: FftField> Circuit<F> for PlonkCircuit<F> {
             }
             // key-value map lookup gates
             let mut key_val_table = HashSet::new();
-            key_val_table.insert((F::zero(), F::zero(), F::zero()));
-            let mut num_table_elems: u32 = 0;
+            key_val_table.insert((F::zero(), F::zero(), F::zero(), F::zero()));
             let q_lookup_vec = self.q_lookup();
-            for (gate_id, &q_lookup) in q_lookup_vec.iter().enumerate() {
+            let q_dom_sep_vec = self.q_dom_sep();
+            let table_key_vec = self.table_key_vec();
+            let table_dom_sep_vec = self.table_dom_sep_vec();
+            // insert table elements
+            for (gate_id, ((&q_lookup, &table_dom_sep), &table_key)) in q_lookup_vec
+                .iter()
+                .zip(table_dom_sep_vec.iter())
+                .zip(table_key_vec.iter())
+                .enumerate()
+            {
                 if q_lookup != F::zero() {
-                    let key = F::from(num_table_elems);
                     let val0 = self.witness(self.wire_variable(3, gate_id))?;
                     let val1 = self.witness(self.wire_variable(4, gate_id))?;
-                    key_val_table.insert((key, val0, val1));
-                    num_table_elems += 1;
+                    key_val_table.insert((table_dom_sep, table_key, val0, val1));
                 }
             }
-            for (gate_id, &q_lookup) in q_lookup_vec.iter().enumerate() {
+            // check lookups
+            for (gate_id, (&q_lookup, &q_dom_sep)) in
+                q_lookup_vec.iter().zip(q_dom_sep_vec.iter()).enumerate()
+            {
                 if q_lookup != F::zero() {
                     let key = self.witness(self.wire_variable(0, gate_id))?;
                     let val0 = self.witness(self.wire_variable(1, gate_id))?;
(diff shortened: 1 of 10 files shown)


plonk/src/circuit/basic.rs
```diff
@@ -314,29 +314,38 @@ impl<F: FftField> Circuit<F> for PlonkCircuit<F> {
...
+            // insert table elements
+            for (gate_id, ((&q_lookup, &table_dom_sep), &table_key)) in q_lookup_vec
+                .iter()
+                .zip(table_dom_sep_vec.iter())
+                .zip(table_key_vec.iter())
+                .enumerate()
+            {
                 if q_lookup != F::zero() {
-                    let key = F::from(num_table_elems);
                     let val0 = self.witness(self.wire_variable(3, gate_id))?;
```
zhenfeizhang, line 330: Should we define some constants for wire types and place `3`, `4` here to improve code readability?

plonk/src/circuit/gates.rs
```diff
@@ -44,6 +44,18 @@ pub trait Gate<F: Field>: Downcast + GateClone<F> {
...
     }
+    /// UltraPlonk lookup domain separation selector.
+    fn q_dom_sep(&self) -> F {
+        F::zero()
+    }
+    /// UltraPlonk table keys.
+    fn table_key(&self) -> F {
+        F::zero()
+    }
+    /// UltraPlonk table domain sepration ids
```
zhenfeizhang, line 55: sepration -> separation

plonk/src/circuit/basic.rs
```diff
@@ -1171,24 +1195,45 @@ where
...
+            domain.ifft(&self.table_key_vec()),
+        ))
+    }
+
+    fn compute_table_dom_sep_polynomial(&self) -> Result<DensePolynomial<F>, PlonkError> {
+        self.check_plonk_type(PlonkType::UltraPlonk)?;
+        self.check_finalize_flag(true)?;
+        let domain = &self.eval_domain;
+        Ok(DensePolynomial::from_coefficients_vec(
+            domain.ifft(&self.table_dom_sep_vec()),
```
zhenfeizhang, line 1211: Not a comment but just to aid my own understanding. 
So here we are doing two more `ifft`s. 
It seems that the degree `n` is much larger than `dom_sep.len()`.
Is it still better to do iffts (which costs the same regardless `dom_sep.len()`)? Can we KZG-commit a smaller poly here?

plonk/src/circuit/basic.rs
chancharles92: Good point, addressed in f8cf2f64504edf04467fd0704cb1f190c10f085f

plonk/src/circuit/gates.rs
chancharles92: Thanks, fixed in f8cf2f64504edf04467fd0704cb1f190c10f085f

zhenfeizhang (approved):

plonk/src/circuit/customized/ultraplonk/lookup_table.rs
```diff
@@ -42,24 +36,38 @@ impl<F: PrimeField> PlonkCircuit<F> {
...
-            self.insert_gate(&wire_vars, Box::new(LookupGate))?;
+
+            self.insert_gate(
+                &wire_vars,
+                Box::new(LookupGate {
+                    q_dom_sep,
+                    table_dom_sep,
+                    table_key,
+                }),
+            )?;
```
alxiong, line 70: I'm not very convinced that we need `table_key` for each `Gate`.

Our `fn create_table_and_lookup_variables()` is actually not as general as the name suggests, there's a hidden/implicit assumption which is the online lookup table has a consecutive `table_key` field (namely the truely general API should accept query and lookup tuple of the same length:
```rust
        query_vars: &[(Variable, Variable, Variable)],
        table_vars: &[(Variable, Variable, Variable)],
```
)

If I understand correctly, the main reason for us to have this during arithmetization, it's easier to merge tables if we can just read from this "virtual selector column". My point being, even without this `circuit.table_key_vec()`, we can still calculate this with `table_gate_ids: Vec<(GateId, usize)>` and the implicit assumption that these `table_key` are incremental inside each sub-table. 

^^ am I correct? @chancharles92 
(I'm not suggesting we should remove it, cuz it doesn't burden the final proof size, just increase the memory usage by forcing each gate to have this "virtual selector column". )

Issue #52, opened 2022-04-09
implement BLS and BLS-VRF


zhenfeizhang:
- [x] define a generic signature trait and move Schnorr under that trait #53 
- [x] implement BLS signature https://github.com/EspressoSystems/jellyfish/pull/55
- [x] define a generic VRF trait https://github.com/EspressoSystems/jellyfish/pull/55
- [x] implement BLS-VRF https://github.com/EspressoSystems/jellyfish/pull/55
- [ ] (Bonus) implement EC-VRF

PR #53, opened 2022-04-09
signature trait

zhenfeizhang:
## Description

First task of #52 
---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/primitives/src/circuit/schnorr_dsa.rs
+++ b/primitives/src/circuit/schnorr_dsa.rs
@@ -8,7 +8,7 @@
 
 use crate::{
     constants::{challenge_bit_len, field_bit_len},
-    schnorr_dsa::{Signature, VerKey, DOMAIN_SEPARATION},
+    signatures::schnorr::{Signature, VerKey, DOMAIN_SEPARATION},
 };
 use ark_ec::{twisted_edwards_extended::GroupAffine, AffineCurve, TEModelParameters as Parameters};
 use ark_ff::PrimeField;
@@ -64,7 +64,7 @@ where
     /// * `msg` - message variables that have been signed.
     /// * `sig` - signature variable.
     /// * `returns` - a bool variable indicating whether the signature is valid.
-    fn check_valid_signature(
+    fn check_signature_validity(
         &mut self,
         vk: &VerKeyVar,
         msg: &[Variable],
@@ -105,7 +105,7 @@ where
         Ok(())
     }
 
-    fn check_valid_signature(
+    fn check_signature_validity(
         &mut self,
         vk: &VerKeyVar,
         msg: &[Variable],
@@ -196,7 +196,7 @@ where
 mod tests {
     use crate::{
         circuit::schnorr_dsa::*,
-        schnorr_dsa::{KeyPair, Signature, VerKey},
+        signatures::schnorr::{KeyPair, Signature, VerKey},
     };
     use ark_ed_on_bls12_377::EdwardsParameters as Param377;
     use ark_ed_on_bls12_381::EdwardsParameters as Param381;
@@ -311,7 +311,7 @@ mod tests {
             .iter()
             .map(|m| circuit.create_variable(*m))
             .collect::<Result<Vec<_>, PlonkError>>()?;
-        let bit = SignatureGadget::<_, P>::check_valid_signature(
+        let bit = SignatureGadget::<_, P>::check_signature_validity(
             &mut circuit,
             &vk_var,
             &msg_var,
--- a/primitives/src/signatures/mod.rs
+++ b/primitives/src/signatures/mod.rs
@@ -1,0 +1,69 @@
+//! Module for signature primitives.
+
+use crate::errors::PrimitivesError;
+use ark_std::rand::{CryptoRng, RngCore};
+
+pub mod bls;
+pub mod schnorr;
+
+/// Trait definition for a signature scheme.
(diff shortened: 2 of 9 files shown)


chancharles92 (approved):
LGTM

primitives/src/circuit/schnorr_dsa.rs
```diff
@@ -8,7 +8,7 @@
```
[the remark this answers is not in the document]
zhenfeizhang: Yea sounds like a great idea

primitives/src/circuit/schnorr_dsa.rs
```diff
@@ -64,7 +64,7 @@ where
     /// * `msg` - message variables that have been signed.
     /// * `sig` - signature variable.
     /// * `returns` - a bool variable indicating whether the signature is valid.
-    fn is_valid_signature(
+    fn check_signature_validity(
```
alxiong, line 67: I think the (implicit) naming convention in current code is:
for gadgets that "enforces" a relation among input params, we use `check_xxx() -> Result<(), Error>`; 
for gadgets that produces a boolean bit about the SAT of relation among the input params, we use `is_xxx() -> Result<Variable, Error>`

Here maybe we should stick with the existing name?

primitives/src/signatures/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+use ark_std::rand::{CryptoRng, RngCore};
+
+pub mod bls;
+pub mod schnorr;
+
+/// Trait definition for a signature scheme.
+// A signature scheme is associated with a hash function H that is
+// to be used for challenge generation.
+// FIXME: update H bound once hash-api is merged.
+pub trait SignatureScheme<H> {
```
alxiong, line 13: curious, can our current implementation fit into [`trait SignatureScheme` by arkwork](https://docs.rs/ark-crypto-primitives/latest/ark_crypto_primitives/signature/trait.SignatureScheme.html)?

it's kinda weird to see a general Signature Scheme trait generic over a choice of hash.

primitives/src/circuit/schnorr_dsa.rs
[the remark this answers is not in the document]
alxiong: maybe add in rustdoc comment that the circuit gadget is for EdDSA, but the Schnorr signature is generic over any curve, not just Edwards Curve.

primitives/src/circuit/schnorr_dsa.rs
zhenfeizhang: yeah. clippy complaint about the use of `is_xxx()` -- such a function should not mutate `self` which makes sense. Do we have other options than `is_xxx()`?

primitives/src/signatures/mod.rs
zhenfeizhang: Second thought. I think I misunderstood your point. The trait itself does not need to associate with a hash function. We should associate hash function with the actual instantiation.

alxiong: [RustCrypto's trait](https://docs.rs/signature/latest/signature/)

primitives/src/circuit/schnorr_dsa.rs
alxiong: Maybe: change all current `is_xxx()` (such as [here](https://github.com/EspressoSystems/jellyfish/blob/main/plonk/src/circuit/customized/mod.rs#L326)) to `check_xxx()` and those that enforce and will return error if not SAT, we keep using `xxx_gate()` (like [this](https://github.com/EspressoSystems/jellyfish/blob/main/plonk/src/circuit/customized/mod.rs#L316))?
as long as we are consistent, I'm fine.

chancharles92:
> Seem like we have converged to the first candidate? @chancharles92 @alxiong

Agree

zhenfeizhang:
Okay. now this branch implements our own signature trait, and is ready for review. 
FYI: code with arkworks signature trait is stashed here: https://github.com/EspressoSystems/jellyfish/tree/arkwork's-signature-api

primitives/src/signatures/schnorr.rs
```diff
@@ -174,7 +273,7 @@ where
     P: Parameters<BaseField = F> + Clone,
 {
     /// Key-pair generation algorithm
-    pub fn generate<R: CryptoRng + RngCore>(prng: &mut R) -> KeyPair<P> {
+    pub fn generate<R: Rng>(prng: &mut R) -> KeyPair<P> {
```
alxiong, line 276: this triggers me to think if we should replace some `RngCore + CryptoRng` trait bound in our API to just `Rng`?

1. We are not implementing an PRNG ourselves, so we should probably always use `Rng` over `RngCore`, (since the latter is for [backend](https://docs.rs/rand/latest/rand/trait.RngCore.html) implemented by the generator)
2. What are some places where we have to use CSPRNG? probably not in schonrr or transcript?

zhenfeizhang: I think we need to carefully go through all the usage in JF. This is not likely to be handled within this PR

chancharles92:
LGTM, do we want to add this into tag-release 0.0.1 or should we add it after the tag-release PR?

zhenfeizhang:
Perhaps after. There are likely breaking changes after this.

chancharles92 (approved):

PR #55, opened 2022-04-11
naive impl bls signature and vrf

zhenfeizhang:
## Description

closes: #XXXX

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/primitives/src/signatures/bls.rs
+++ b/primitives/src/signatures/bls.rs
@@ -4,7 +4,7 @@ use super::SignatureScheme;
 use crate::{constants::CS_ID_BLS_SIG_NAIVE, errors::PrimitivesError, hash_to_group::HashToGroup};
 use ark_ec::{
     bls12::{Bls12, Bls12Parameters},
-    short_weierstrass_jacobian::{GroupAffine, GroupProjective},
+    short_weierstrass_jacobian::GroupAffine,
     AffineCurve, ModelParameters, ProjectiveCurve,
 };
 use ark_ff::{Fp12, PrimeField};
@@ -62,7 +62,10 @@ where
     type VerificationKey = BLSVerKey<P>;
 
     /// Public Parameter
-    type PublicParameter = GroupAffine<P::G2Parameters>;
+    /// For BLS signatures, we want to use default
+    /// prime subgroup generators. So here we don't need
+    /// to specify which PP it is.
+    type PublicParameter = ();
 
     /// Signature
     type Signature = BLSSignature<P>;
@@ -73,24 +76,25 @@ where
     /// generate public parameters from RNG.
     /// If the RNG is not presented, use the default group generator.
     fn param_gen<R: CryptoRng + RngCore>(
-        prng: Option<&mut R>,
+        _prng: Option<&mut R>,
     ) -> Result<Self::PublicParameter, PrimitivesError> {
-        match prng {
-            None => Ok(GroupAffine::<P::G2Parameters>::prime_subgroup_generator()),
-            Some(prng) => Ok(GroupProjective::<P::G2Parameters>::rand(prng).into_affine()),
-        }
+        Ok(())
     }
 
     /// Sample a pair of keys.
     fn key_gen<R: CryptoRng + RngCore>(
-        pp: &Self::PublicParameter,
+        _pp: &Self::PublicParameter,
         prng: &mut R,
     ) -> Result<(Self::SigningKey, Self::VerificationKey), PrimitivesError> {
         // TODO: absorb ciphersuite_id in prng
         let sk = BLSSignKey(<P::G1Parameters as ModelParameters>::ScalarField::rand(
             prng,
         ));
-        let vk = BLSVerKey(pp.mul(sk.0).into_affine());
+        let vk = BLSVerKey(
+            GroupAffine::<P::G2Parameters>::prime_subgroup_generator()
+                .mul(sk.0)
+                .into_affine(),
+        );
         Ok((sk, vk))
     }
 
@@ -110,7 +114,7 @@ where
 
     /// Verify a signature.
     fn verify<M: AsRef<[Self::MessageUnit]>>(
-        pp: &Self::PublicParameter,
+        _pp: &Self::PublicParameter,
         vk: &Self::VerificationKey,
         msg: M,
         sig: &Self::Signature,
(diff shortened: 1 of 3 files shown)


zhenfeizhang:
This is a naive implementation of bls signatures and vrfs. using an inefficient way to hash element to the group.
There are a couple of todos:
- implement aggregation and batch verification
- implement ieft hash to group


For now we want to know if the APIs are sufficient for your need @nmccarty

primitives/src/hash_to_group/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+        cs_id: B,
+    ) -> Result<GroupProjective<Self>, PrimitivesError> {
+        let mut hasher = Sha256::new();
+        hasher.update([cs_id.as_ref(), data.as_ref()].concat());
+        let mut seed = [0u8; 32];
+        seed.copy_from_slice(hasher.finalize().as_ref());
+        let mut rng = ChaCha20Rng::from_seed(seed);
+        loop {
+            let x = Self::BaseField::rand(&mut rng);
+            let greatest = rng.gen();
```
chancharles92, line 43: Q: what's the usage of `greatest` here?

primitives/src/signatures/bls.rs
```diff
@@ -1 +1,165 @@
...
+    rand::{CryptoRng, RngCore},
+    string::ToString,
+    One, UniformRand,
+};
+use core::marker::PhantomData;
+use jf_utils::{multi_pairing, tagged_blob};
+
+/// BLS signature scheme.
+/// Optimized for signature size, i.e.: PK in G2 and sig in G1
+pub struct BLSSignatureScheme<P: Bls12Parameters> {
```
chancharles92, line 24: `P: Bls12Parameters`: Does this mean that we'll only use BLS curves for BLS signatures? That is, we won't use other pairing-friendly curves?

primitives/src/hash_to_group/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+
+            if let Some(p) = GroupAffine::<Self>::get_point_from_x(x, greatest) {
+                return Ok(p.mul_by_cofactor_to_projective());
+            }
+        }
+    }
+}
+
+impl HashToGroup for ark_bls12_381::g1::Parameters {
+    // todo
```
chancharles92, line 53: Minor: How about capitalizing `todo` to `TODO` so that it's easier to spot? (Same for other places.)

primitives/src/signatures/bls.rs
```diff
@@ -1 +1,165 @@
...
+        ));
+        let vk = BLSVerKey(
+            GroupAffine::<P::G2Parameters>::prime_subgroup_generator()
+                .mul(sk.0)
+                .into_affine(),
+        );
+        Ok((sk, vk))
+    }
+
+    /// Sample a pair of keys.
```
chancharles92, line 97: Fix the comment.

primitives/src/signatures/schnorr.rs
```diff
@@ -50,32 +51,34 @@ where
...
     type Signature = Signature<P>;
 
-    /// message
-    type MessageUnit = P::BaseField;
+    /// A message is &\[MessageUnit\]
+    type MessageUnit = F;
 
     /// generate public parameters from RNG.
-    fn setup<R: CryptoRng + RngCore>(
+    fn param_gen<R: CryptoRng + RngCore>(
```
chancharles92, line 58: Q: In what cases do we need `PublicParameter`? Does a group generator count as part of pp?

primitives/src/vrf/blsvrf.rs
```diff
@@ -0,0 +1,132 @@
...
+};
+use digest::Digest;
+
+/// BLS VRF scheme.
+/// Optimized for signature size, i.e.: PK in G2 and sig in G1
+pub struct BLSVRFScheme<P: Bls12Parameters> {
+    pairing_friend_curve: PhantomData<P>,
+}
+
+impl<H, P> Vrf<H, P> for BLSVRFScheme<P>
```
chancharles92, line 28: Just to be curious: do we need any aggregation APIs regarding VRF?

primitives/src/hash_to_group/mod.rs
zhenfeizhang: A flag to indicate if we want to take positive or negative y cooridinate

primitives/src/signatures/bls.rs
zhenfeizhang: Yeah. I tried to make it generic over `E: PairingEngine` but `PairingEngine` does not expose `ModelParameters` which is needed for hashes, etc.

primitives/src/signatures/schnorr.rs
zhenfeizhang: good question. I was thinking to pub the group generators there -- this allows users to use customized generators other than the one in Arkworks. I haven't thought about the security implications though.

primitives/src/vrf/blsvrf.rs
zhenfeizhang: Not aggregation, but BLS-VRFs can be batch verified. We can use that APIs for acceleration.

primitives/src/hash_to_group/mod.rs
chancharles92: I see, thanks. Is `greatest` a bool? Would `flag` be a better name?

primitives/src/signatures/bls.rs
chancharles92: I see, sounds good for me.

primitives/src/vrf/blsvrf.rs
chancharles92: Yeah, make sense.

primitives/src/hash_to_group/mod.rs
zhenfeizhang: fixed with https://github.com/EspressoSystems/jellyfish/pull/55/commits/19d9c81b42ca8481d8d94c4fcb8f5b4d497e588f

primitives/src/signatures/schnorr.rs
zhenfeizhang: hmmm in https://github.com/EspressoSystems/jellyfish/pull/55/commits/19d9c81b42ca8481d8d94c4fcb8f5b4d497e588f in wrote a potential solution. The api is a bit strange though, when the `default generator` is called.

zhenfeizhang:
Just realized that arkworks has already implemented the `standard` hash_to_curve. So one less thing we need to do.

https://github.com/arkworks-rs/algebra/blob/master/ec/src/hashing/curve_maps/swu/mod.rs

chancharles92 (approved):

Issue #63, opened 2022-05-12
Bugfix: range-gate-with-lookup


chancharles92:
`range-gate-lookup()` should always enforce a linear combination gate after decomposing the witness (even if the witness is supposed to be less than `range_size()`):
https://github.com/EspressoSystems/jellyfish/blob/main/plonk/src/circuit/customized/ultraplonk/range.rs#L51

PR #64, opened 2022-05-12
bugfix range-gate

chancharles92:
## Description

`range-gate-lookup()` should always enforce a linear combination gate after decomposing the witness.

closes: #63


zhenfeizhang (approved):
LGTM

PR #65, opened 2022-05-13
impl H2G for TE curves

zhenfeizhang:
## Description

Implement hash to group features for twisted edwards curves

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 
 - Derive `Debug`, `Snafu` on `enum TaggedBlobError`
 - Updated `tagged-base64` reference url to reflect the Espresso Systems name change
+- Add `HashToGroup` support for both SW and TE curves
 
 ### Bugfixes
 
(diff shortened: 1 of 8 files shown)


alxiong (commented):
- Edward -> Edwards
- add for jubjub over bls12-381
- add to changelog

I will address my own comments in a following commit

another question is: so it seems that the algorithm for both SW and TE is the same? (you mentioned that there might be slight difference yesterday? where (what document) can we double check?)

alxiong (approved):

zhenfeizhang:
Thanks for finishing this up. 

On a related note:
Arkworks at the moment only have the trait definitions for Wahby-Boneh hash-to-group: https://github.com/arkworks-rs/algebra/tree/master/ec/src/hashing/curve_maps
We cannot really use those until the actual `WBParams` are defined.

alxiong:
>  For both SW and TE, y has a sign bit. So they can share a same API.

so basically we can have `HashToGroup: ModelParameters` instead? and unify the two `TEHashToGroup` and `SWHashToGroup`, correct? @zhenfeizhang

PR #70, opened 2022-05-23
fixing a bug for solidity transcript

zhenfeizhang:
## Description

closes: #XXXX

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/plonk/src/proof_system/snark.rs
+++ b/plonk/src/proof_system/snark.rs
@@ -343,6 +343,9 @@ where
 
         // Initialize transcript
         let mut transcript = T::new(b"PlonkProof");
+        // The extra init message makes the system vulnerable to transcript extension
+        // attacks when transcript is not instantiated correctly.
+        // There is a WIP pr to fix this.
         if let Some(msg) = extra_transcript_init_msg {
             transcript.append_message(EXTRA_TRANSCRIPT_MSG_LABEL, &msg)?;
         }
--- a/plonk/src/proof_system/snark.rs
+++ b/plonk/src/proof_system/snark.rs
@@ -343,9 +343,6 @@ where
 
         // Initialize transcript
         let mut transcript = T::new(b"PlonkProof");
-        // The extra init message makes the system vulnerable to transcript extension
-        // attacks when transcript is not instantiated correctly.
-        // There is a WIP pr to fix this.
         if let Some(msg) = extra_transcript_init_msg {
             transcript.append_message(EXTRA_TRANSCRIPT_MSG_LABEL, &msg)?;
         }
--- a/plonk/src/transcript/solidity.rs
+++ b/plonk/src/transcript/solidity.rs
@@ -34,18 +34,19 @@ pub struct SolidityTranscript {
 }
 
 impl<F> PlonkTranscript<F> for SolidityTranscript {
-    /// Create a new plonk transcript. `label` is omitted for efficiency.
-    fn new(_label: &'static [u8]) -> Self {
+    /// Create a new plonk transcript.
+    fn new(label: &'static [u8]) -> Self {
         SolidityTranscript {
-            transcript: Vec::new(),
+            transcript: label.to_vec(),
             state: [0u8; KECCAK256_STATE_SIZE],
         }
     }
 
-    /// Append the message to the transcript. `_label` is omitted for
-    /// efficiency.
-    fn append_message(&mut self, _label: &'static [u8], msg: &[u8]) -> Result<(), PlonkError> {
-        // We remove the labels for better efficiency
+    /// Append the message to the transcript.
+    fn append_message(&mut self, label: &'static [u8], msg: &[u8]) -> Result<(), PlonkError> {
+        self.transcript.extend_from_slice(label);
+        self.transcript
+            .extend_from_slice(msg.len().to_le_bytes().as_ref());
         self.transcript.extend_from_slice(msg);
         Ok(())
     }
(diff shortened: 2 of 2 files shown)


alxiong:
this is a stale PR. iirc, the original motivation was just defense in depth in the light of [trailofbits' disclosure on FrozenHeart ](https://blog.trailofbits.com/2022/04/13/part-1-coordinated-disclosure-of-vulnerabilities-affecting-girault-bulletproofs-and-plonk/)where some FS transcripts forget taking in all the public inputs (especially the vk), but our library has always hashed all PI into our transcript, and adding a domain separator (the label) for each field may not have a security impact, just good-to-have.

Adding this to `SolidityTranscript` will increase the gas cost for PlonkVerifier. 

Just want to bring this out for discussion. @EspressoSystems/jellyfish

mrain:
Is there any reference for old discussion about this issue?

Also is the gas cost increase going to be significant if we add those domain separators? @alxiong

alxiong:
> Is there any reference for old discussion about this issue?

Sorry, I can't recall. @chancharles92 do you have anything to add beyond my comments above?

> Also is the gas cost increase going to be significant if we add those domain separators?

yes, it's non-neglible cuz we are hashing using rescue which is expensive in Solidity (in absolute gas cost, but somewhat manageable in relative gas cost), the more pressing problem is the contract bytecode size will increase. I remember we have a lots of trouble trying to keep CAPE contract under size limit.

chancharles92:
> > Is there any reference for old discussion about this issue?
> 
> Sorry, I can't recall. @chancharles92 do you have anything to add beyond my comments above?

My impression is that Zhenfei mentioned that our SNARK/Circuit transcript implementations are secure, but I don't remember clearly whether the solidity transcript is correct or not.

chancharles92:
Found the context here: https://www.notion.so/espressosys/Potential-vulnerability-on-Jellyfish-s-transcript-91e810dd59c849aba7bab4ef9baaffb9

alxiong:
Thanks for finding the notion page. I have a question, so it's probably a good idea to fix "rescue transcript". But is it necessary to also add label to "solidity transcript"? The scenario described where 3 BN254 points turn out to be identical to 2 BLS12-381 points are not possible in our CAPE contract. 

Also I remember Aztec's plonk verifier contract doesn't add label, neither does [matter-lab](https://github.com/matter-labs/solidity_plonk_verifier/blob/master/Verifier.sol).

chancharles92:
> Thanks for finding the notion page. I have a question, so it's probably a good idea to fix "rescue transcript". But is it necessary to also add label to "solidity transcript"? The scenario described where 3 BN254 points turn out to be identical to 2 BLS12-381 points are not possible in our CAPE contract.
> 
> Also I remember Aztec's plonk verifier contract doesn't add label, neither does [matter-lab](https://github.com/matter-labs/solidity_plonk_verifier/blob/master/Verifier.sol).

Yes, check TL; DR. We didn't need to merge this PR now. OTOH, fixing the "rescue transcript" also has some costs because the SNARK circuit will have more constraints.

mrain:
> > Thanks for finding the notion page. I have a question, so it's probably a good idea to fix "rescue transcript". But is it necessary to also add label to "solidity transcript"? The scenario described where 3 BN254 points turn out to be identical to 2 BLS12-381 points are not possible in our CAPE contract.
> > Also I remember Aztec's plonk verifier contract doesn't add label, neither does [matter-lab](https://github.com/matter-labs/solidity_plonk_verifier/blob/master/Verifier.sol).
> 
> Yes, check TL; DR. We didn't need to merge this PR now. OTOH, fixing the "rescue transcript" also has some costs because the SNARK circuit will have more constraints.

I pushed a commit for fixing the rescue transcript. Please check @alxiong @chancharles92

chancharles92:
> > > Thanks for finding the notion page. I have a question, so it's probably a good idea to fix "rescue transcript". But is it necessary to also add label to "solidity transcript"? The scenario described where 3 BN254 points turn out to be identical to 2 BLS12-381 points are not possible in our CAPE contract.
> > > Also I remember Aztec's plonk verifier contract doesn't add label, neither does [matter-lab](https://github.com/matter-labs/solidity_plonk_verifier/blob/master/Verifier.sol).
> > 
> > 
> > Yes, check TL; DR. We didn't need to merge this PR now. OTOH, fixing the "rescue transcript" also has some costs because the SNARK circuit will have more constraints.
> 
> I pushed a commit for fixing the rescue transcript. Please check @alxiong @chancharles92

If we fix the rescue transcript, should we also sync the change in the circuit? E.g., in `append_variable()`, we should also append the label constant.
https://github.com/EspressoSystems/jellyfish/blob/main/plonk/src/circuit/transcript.rs#L84

mrain:
> > > > Thanks for finding the notion page. I have a question, so it's probably a good idea to fix "rescue transcript". But is it necessary to also add label to "solidity transcript"? The scenario described where 3 BN254 points turn out to be identical to 2 BLS12-381 points are not possible in our CAPE contract.
> > > > Also I remember Aztec's plonk verifier contract doesn't add label, neither does [matter-lab](https://github.com/matter-labs/solidity_plonk_verifier/blob/master/Verifier.sol).
> > > 
> > > 
> > > Yes, check TL; DR. We didn't need to merge this PR now. OTOH, fixing the "rescue transcript" also has some costs because the SNARK circuit will have more constraints.
> > 
> > 
> > I pushed a commit for fixing the rescue transcript. Please check @alxiong @chancharles92
> 
> If we fix the rescue transcript, should we also sync the change in the circuit? E.g., in `append_variable()`, we should also append the label constant. https://github.com/EspressoSystems/jellyfish/blob/main/plonk/src/circuit/transcript.rs#L84
> 
> (It's strange that the tests didn't fail after the changes.)

Sure we should. But it's not easy in the current gadget design. I'll try to do it today.

plonk/src/transcript/rescue.rs
```diff
@@ -93,27 +93,31 @@ where
...
 
-    /// Append the message to the transcript. `_label` is omitted for
-    /// efficiency.
-    fn append_message(&mut self, _label: &'static [u8], msg: &[u8]) -> Result<(), PlonkError> {
-        // We remove the labels for better efficiency
-
-        let mut f = bytes_to_field_elements(&msg);
-        self.transcript.append(&mut f);
+    /// Append the message to the transcript.
+    /// TODO(Chengyu): fix the bug here, current design brings trouble to the
```
mrain, line 97: I updated the gadget implementation for rescue transcript. However, I encountered trouble passing the tests. The reason is described here.
I would also suggest remove this from the current release plan. 
@alxiong @chancharles92

alxiong: Agree, we can include this change in a patch release afterwards.

mrain:
Also could someone check the CI here? It failed test on my local computer. But CI passed.

chancharles92:
> Also could someone check the CI here? It failed test on my local computer. But CI passed.

@Ancient123 Can we do similar changes in Jellyfish as we did to the HyperPlonk repo?

PR #71, opened 2022-06-02
Zprice test vectors and benchmark code

alxiong:
This PR includes:

- Remove non-standard transcripts
- Add MT circuit generation code 
- Add scripts to generate srs, pk, vk
- Add README to explain to participants


zprice/src/serde.rs
```diff
@@ -0,0 +1,160 @@
...
+        Some(dest) => dest,
+        None => default_path(DEFAULT_UNIVERSAL_SRS_FILENAME, "bin"),
+    };
+
+    let now = Instant::now();
+    print!(
+        "Storing universal parameter to: {} ...",
+        dest.to_str().unwrap()
+    );
+    store_data(&universal_param, dest);
```
zhenfeizhang, line 32: do you think we should also store a hash digest of the srs (which we will also put in the readme.md?)

zhenfeizhang (approved):
LGTM. Just a minor comment on the srs digest for which I can try to obtain after running the instance.

alxiong:
Yes, @zhenfeizhang , we should produce sha256sum for all binary files and version controlled them.

an example of this is:
I used to use [this bash script](https://github.com/alxiong/crs/blob/main/scripts/download_transcripts_aztec.sh#L27-L30) to check the Aztec Ignition CRS file's integrity by adding [this checksum file](https://github.com/alxiong/crs/blob/main/data/aztec20/checksums).

PR #72, opened 2022-06-17
Made `tagged_blob` be able to accept constants

VictorKoenders:
## Description

As discussed in zulip, we want to make sure we don't have conflicting `tagged-base64` tags. For this I've made [espresso-systems-common](https://github.com/EspressoSystems/espresso-systems-common)

The `#[tagged_blob(...)]` macro did not support constants. This PR adds this, as well as a test to make sure the new derive macro actually works.

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/utilities/src/serialize.rs
+++ b/utilities/src/serialize.rs
@@ -205,6 +205,39 @@ pub mod tagged_blob {
         let bytes = deserialize_with_tag(T::tag().as_str(), deserializer)?;
         Ok((bytes, Default::default()))
     }
+
+    #[cfg(test)]
+    mod test {
+        use crate as jf_utils;
+        use ark_serialize::*;
+        use jf_utils::tagged_blob;
+        use std::vec::Vec;
+
+        #[tagged_blob("A")]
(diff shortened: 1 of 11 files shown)


utilities/src/serialize.rs
```diff
@@ -205,6 +205,39 @@ pub mod tagged_blob {
...
+
+        #[tagged_blob("A")]
+        #[derive(Debug, Clone, PartialEq, Eq, CanonicalDeserialize, CanonicalSerialize)]
+        pub struct A(Vec<u8>);
+
+        #[test]
+        fn test_tagged_blob_static_str() {
+            let a = A(Vec::new());
+            let str = serde_json::to_string(&a).unwrap();
+            assert_eq!(str, r#""A~AAAAAAAAAABr""#);
```
alxiong, line 224: question: can you help me recall why this would print to this string? Why `AAAAAAAAAABr` ?

alxiong (approved):

utilities/src/serialize.rs
VictorKoenders: I believe `AAAAAAAAAAB` is the length of the buffer (0u64) and `r` is the checksum

PR #76, opened 2022-06-30
mask splitting polys for zk

alxiong:
## Description

Sync with the recent [bug fix of Plonk paper](https://twitter.com/rel_Aztec/status/1542474186664210432?s=20&t=AXa-oonIR5EimKsMyaybWA) on masking the splitting polynomials to ensure successful ZK simulator.

image

The slight modification in our implementation is that our splitting step degree is `n+2` instead of `n`. (since our splitting polynomials are of degree `= n+1`, see [here](https://github.com/EspressoSystems/jellyfish/blob/main/plonk/src/proof_system/prover.rs#L897) and [here](https://github.com/EspressoSystems/jellyfish/blob/main/plonk/src/proof_system/prover.rs#L897))
@chancharles92 Do you remember why we decide to do this instead of `split_lo + split_mid * X^n + ...` as in the paper?

---

write a little note why.

- [x] ~Wrote unit tests~

### DIFF
--- a/plonk/src/proof_system/prover.rs
+++ b/plonk/src/proof_system/prover.rs
@@ -896,6 +896,14 @@ impl<E: PairingEngine> Prover<E> {
 
     /// Split the quotient polynomial into `num_wire_types` polynomials.
     /// The first `num_wire_types`-1 polynomials have degree `domain_size`+1.
+    ///
+    /// Let t(X) be the input quotient polynomial, t_i(X) be the output
+    /// splitting polynomials. t(X) = \sum_{i=0}^{num_wire_types}
+    /// X^{i*(n+2)} * t_i(X)
+    ///
+    /// NOTE: we have a step polynomial of X^(n+2) instead of X^n as in the
+    /// GWC19 paper to achieve better balance among degrees of all splitting
+    /// polynomials (especially the highest-degree/last one).
     fn split_quotient_polynomial<R: CryptoRng + RngCore>(
         &self,
         prng: &mut R,
@@ -907,6 +915,8 @@ impl<E: PairingEngine> Prover<E> {
             return Err(WrongQuotientPolyDegree(quot_poly.degree(), expected_degree).into());
         }
         let n = self.domain.size();
+        // compute the splitting polynomials t'_i(X) s.t. t(X) =
+        // \sum_{i=0}^{num_wire_types} X^{i*(n+2)} * t'_i(X)
         let mut split_quot_polys: Vec<DensePolynomial<E::Fr>> = (0..num_wire_types)
             .into_par_iter()
             .map(|i| {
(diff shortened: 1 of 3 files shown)


chancharles92:
> 

Because we also mask witness polynomials and Z polynomials at the beginning (so their degrees are increased by 1 or 2 initially), and I believe `(n+1)` is the best parameter to make each polynomial's degree balanced. If we use `n`, the last splitted polynomial will have much higher degree.

chancharles92:
@alxiong Do you know why it's insecure even if we've masked both witness polynomials and Z(X)? Also I saw this from Ariel's post:

`Since this seems to not be an exploitable issue, and it seems infeasible to use this to extract any user information;`

plonk/src/proof_system/prover.rs
```diff
@@ -919,6 +921,29 @@ impl<E: PairingEngine> Prover<E> {
...
+        // mask splitting polynomials
+        // t'(X) = t(X) - b_last + b_now * X^(n+2)
+        // with t'_lowest(X) = t_lowest(X) - 0 + b_now * X^(n+2)
+        // and t'_highest(X) = t_highest(X) - b_last
+        let mut last_randomizer = E::Fr::zero();
+        split_quot_polys
+            .iter_mut()
+            .take(num_wire_types - 1)
+            .for_each(|poly| {
+                let now_randomizer = E::Fr::rand(prng);
```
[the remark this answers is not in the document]
alxiong: This is masking blinding factors that supposed to be only known to prover, not public coins from the transcript, otherwise we lost zk. 

So yeah it should be randomly sampled. (You can also look at how we mask wire polynomial as a reference)

plonk/src/proof_system/prover.rs
```diff
@@ -919,6 +921,29 @@ impl<E: PairingEngine> Prover<E> {
                 )
             })
             .collect();
+
+        // mask splitting polynomials
+        // t'(X) = t(X) - b_last + b_now * X^(n+2)
+        // with t'_lowest(X) = t_lowest(X) - 0 + b_now * X^(n+2)
+        // and t'_highest(X) = t_highest(X) - b_last
```
chancharles92, line 928: Make notation consistent with the paper. `t'_lo`, `t'_mid`, `t'_high` are the polynomials before masking, `t_lo`, `t_mid`, `t_high` are the polynomials after masking.

plonk/src/proof_system/prover.rs
```diff
@@ -919,6 +921,29 @@ impl<E: PairingEngine> Prover<E> {
                 )
             })
             .collect();
+
+        // mask splitting polynomials
+        // t'(X) = t(X) - b_last + b_now * X^(n+2)
```
chancharles92, line 926: maybe write `t'_j(X) = t_j(X) - b_last_j + b_now_j * X^{n+2}`

plonk/src/proof_system/prover.rs
```diff
@@ -919,6 +921,29 @@ impl<E: PairingEngine> Prover<E> {
...
+        split_quot_polys
+            .iter_mut()
+            .take(num_wire_types - 1)
+            .for_each(|poly| {
+                let now_randomizer = E::Fr::rand(prng);
+
+                let mut coeffs = vec![E::Fr::zero(); n + 3];
+                coeffs[0] = last_randomizer.neg();
+                coeffs[n + 2] = now_randomizer;
+                *poly += &DensePolynomial::from_coefficients_vec(coeffs);
```
chancharles92, line 939: Is it possible to directly change `poly`'s 2 coefficients rather than adding an O(n)-sized polynomial? Same as the last splitting poly.

chancharles92 (changes requested):

plonk/src/proof_system/prover.rs
alxiong: very good idea. Mine is easier to read but very wasteful in memory.

chancharles92 (dismissed):
LGTM

chancharles92 (approved):

Issue #84, opened 2022-07-25
unsatisified no_std 


alxiong:
Currently we thought we had  proper `no_std` support since we use `ark_std` by default and only use std library with `features(std)`. 

I run into a bug today on `cap-rollup` branch that suggests otherwise.

## How to reproduce

Check out to [`cap-rollup` branch](https://github.com/EspressoSystems/jellyfish/tree/cap-rollup), run `cargo check` the code should compile successfully, but when running `cargo check --features std`, then you should see this error:

```
    Checking jf-plonk v0.1.2 (/Users/alex/work/jellyfish/plonk)
error[E0119]: conflicting implementations of trait `std::error::Error` for type `errors::PlonkError`
  --> plonk/src/errors.rs:50:1
   |
47 | impl ark_std::error::Error for PlonkError {}
   | ----------------------------------------- first implementation here
...
50 | impl std::error::Error for PlonkError {}
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `errors::PlonkError`
```

This is unexpected since `ark_std::error::Error` should be a different trait than `std::error::Error`. 

## Analysis

@zhenfeizhang and I spent some time digging around. 
- We first eliminate the possibility of leaky no_std from upsteam `ark_std` library by creating a minimal reproducible example.
- We also double checked that our syntax `#![cfg_attr(not(feature = "std"), no_std)]` is a correct `no_std` declaration
- we come to temporary conclusion that we could have accidentally imported a version of `ark_std` with `features = ["std"]`, this could be through `parallel` features of `ark-ff`, `ark-ec`, `ark-bls` or any curve impl crates. 
  - this suspicion gets confirmed by removing the `parallel` feature flag on `ark-ff` and `ark-ec` in our `utilities` crate, and we get the following error when only trying to compile `jf-utils`:
    ```
       --> utilities/src/serialize.rs:119:17
        |
    119 | #[derive(Debug, Snafu)]
        |                 ^^^^^ method cannot be called on `&SerializationError` due to unsatisfied trait bounds
        |
       ::: /Users/alex/.cargo/registry/src/github.com-1ecc6299db9ec823/ark-serialize-0.3.0/src/error.rs:5:1
        |
    5   | pub enum SerializationError {
        | ---------------------------
        | |
        | doesn't satisfy `SerializationError: AsErrorSource`
        | doesn't satisfy `SerializationError: StdError`
        |
[...]

PR #85, opened 2022-07-26
Restoring no_std compliance

alxiong:
## Description

Major changes include: 
- adjust feature flag dep of arkworks so that by default we don't use any `std` components
- remove unnecessary feature flag declaration (or `default-features = false`) on arkworks
- use `hashbrown`'s HashSet, HashMap 
- put `rayon` accelerated code to `parallel` feature flag which is on by default

closes: #84 

---

write a little note why.

- [x] ~Targeted PR against correct branch (main)~
- [x] ~Wrote unit tests~
- [x] ~Updated relevant documentation in the code~

### DIFF
--- a/plonk/Cargo.toml
+++ b/plonk/Cargo.toml
@@ -30,8 +30,8 @@ derivative = { version = "2", features = ["use_core"] }
 num-bigint = { version = "0.4", default-features = false}
 rand_chacha = { version = "0.3.1" }
 sha3 = "^0.10"
-espresso-systems-common = { git = "https://github.com/espressosystems/espresso-systems-common", tag = "0.1.1" }
-
+espresso-systems-common = { git = "https://github.com/espressosystems/espresso-systems-common", branch = "main" }
+hashbrown = "0.12.3"
 
 [dependencies.ark-poly-commit]
 git = "https://github.com/arkworks-rs/poly-commit/"
--- a/plonk/Cargo.toml
+++ b/plonk/Cargo.toml
@@ -11,13 +11,13 @@ jf-utils = { path = "../utilities" }
 jf-rescue = { path = "../rescue" }
 
 ark-std = { version = "0.3.0", default-features = false }
-ark-serialize = { version = "0.3.0", default-features = false }
-ark-ff = { version = "0.3.0", default-features = false, features = ["asm", "parallel"] }
-ark-ec = { version = "0.3.0", default-features = false, features = ["parallel"] }
-ark-poly = { version = "0.3.0", default-features = false,  features = ["parallel"] }
-ark-bn254 = { version = "0.3.0", default-features = false, features = ["curve"] }
-ark-bls12-377 = { git = "https://github.com/arkworks-rs/curves", features = ["curve"], rev = "677b4ae751a274037880ede86e9b6f30f62635af" }
-ark-bls12-381 = { version = "0.3.0", default-features = false, features = ["curve"] }
+ark-serialize = "0.3.0"
+ark-ff = "0.3.0"
+ark-ec = "0.3.0"
+ark-poly = "0.3.0"
+ark-bn254 = "0.3.0"
+ark-bls12-377 = { git = "https://github.com/arkworks-rs/curves", rev = "677b4ae751a274037880ede86e9b6f30f62635af" }
+ark-bls12-381 = "0.3.0"
 ark-bw6-761 = { git = "https://github.com/arkworks-rs/curves", rev = "677b4ae751a274037880ede86e9b6f30f62635af" }
 
 merlin = { version = "3.0.0", default-features = false }
@@ -40,10 +40,10 @@ default-features = false
 
 [dev-dependencies]
 bincode = "1.0"
-ark-ed-on-bls12-381 = { version = "0.3.0", default-features = false }
+ark-ed-on-bls12-381 = "0.3.0"
 ark-ed-on-bls12-377 = { git = "https://github.com/arkworks-rs/curves", rev = "677b4ae751a274037880ede86e9b6f30f62635af" }
-ark-ed-on-bls12-381-bandersnatch = { git = "https://github.com/arkworks-rs/curves", default-features = false, rev = "677b4ae751a274037880ede86e9b6f30f62635af" }
-ark-ed-on-bn254 = { version = "0.3.0", default-features = false }
+ark-ed-on-bls12-381-bandersnatch = { git = "https://github.com/arkworks-rs/curves", rev = "677b4ae751a274037880ede86e9b6f30f62635af" }
+ark-ed-on-bn254 = "0.3.0"
 hex = "^0.4.3"
 
 # Benchmarks
@@ -53,6 +53,7 @@ path = "benches/bench.rs"
 harness = false
 
 [features]
-std = []
-# exposing apis for testing purpose
-test_apis = []
+default = [ "ark-ff/asm" ] # no_std support
+std = [ "ark-std/std", "ark-serialize/std", "ark-ff/std", "ark-ec/std", "ark-poly/std"]
+test_apis = [] # exposing apis for testing purpose
+parallel = [ "ark-ff/parallel", "ark-ec/parallel", "ark-poly/parallel" ]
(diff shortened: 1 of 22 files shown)


plonk/src/testing_apis.rs
```diff
@@ -68,7 +68,7 @@ impl<F: Field> From<Challenges<F>> for structs::Challenges<F> {
 /// A wrapper of crate::proof_system::structs::ScalarsAndBases
 #[derive(Debug, Clone)]
 pub struct ScalarsAndBases<E: PairingEngine> {
-    pub base_scalar_map: HashMap<E::G1Affine, E::Fr>,
+    pub base_scalar_map: Vec<(E::G1Affine, E::Fr)>,
```
zhenfeizhang, line 71: Hmmm. so will this slightly decrease the performance? In the case there are identical bases

plonk/Cargo.toml
```diff
@@ -11,13 +11,13 @@ jf-utils = { path = "../utilities" }
...
 
 ark-std = { version = "0.3.0", default-features = false }
-ark-serialize = { version = "0.3.0", default-features = false }
-ark-ff = { version = "0.3.0", default-features = false, features = ["asm", "parallel"] }
-ark-ec = { version = "0.3.0", default-features = false, features = ["parallel"] }
-ark-poly = { version = "0.3.0", default-features = false,  features = ["parallel"] }
-ark-bn254 = { version = "0.3.0", default-features = false, features = ["curve"] }
-ark-bls12-377 = { git = "https://github.com/arkworks-rs/curves", features = ["curve"], rev = "677b4ae751a274037880ede86e9b6f30f62635af" }
-ark-bls12-381 = { version = "0.3.0", default-features = false, features = ["curve"] }
+ark-serialize = "0.3.0"
```
zhenfeizhang, line 14: iiuc. this means we import ark-serialize etc. with default features, is that right?

alxiong: yes, I checked these, default features is no_std. ([see here](https://github.com/arkworks-rs/algebra/blob/master/serialize/Cargo.toml#L27))

The slightly confusing part is that, except `ark-std`, most of `ark-*` are no_std by default.

plonk/src/testing_apis.rs
alxiong: That's right, this will be worse performance. I was kind of forced to choose this since: 
1. we can't use `HashMap` since that's not available from `alloc`
2. we can't use `BTreeMap` since `G1Affine` doesn't implement `Ord` which is required by [`.entry()`](https://doc.rust-lang.org/alloc/collections/btree_map/struct.BTreeMap.html#method.entry) or any search in the tree.

Can you think of other alternatives?

alxiong:
Currently output from [`cargo-nono`](https://github.com/hobofan/cargo-nono):

```
└─○ cargo nono check --package jf-utils
ALL GOOD

└─○ cargo nono check --package jf-utils-derive
quote: ❌  
  - Did not find a #![no_std] attribute or a simple conditional attribute like #[cfg_attr(not(feature = "std"), no_std)] in the crate source. Crate most likely doesn't support no_std without changes.
  - Source code contains an explicit `use std::` statement.
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: StripPrefixError(())', /Users/alex/.cargo/registry/src/github.com-1ecc6299db9ec823/cargo-nono-0.1.9/src/check_source.rs:88:18

└─○ cargo nono check --package jf-rescue      
ALL GOOD!

└─○ cargo nono check --package jf-primitives
espresso-systems-common: ❌  
  - Did not find a #![no_std] attribute or a simple conditional attribute like #[cfg_attr(not(feature = "std"), no_std)] in the crate source. Crate most likely doesn't support no_std without changes.
rayon: ❌  
  - Did not find a #![no_std] attribute or a simple conditional attribute like #[cfg_attr(not(feature = "std"), no_std)] in the crate source. Crate most likely doesn't support no_std without changes.
  - Source code contains an explicit `use std::` statement.
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: StripPrefixError(())', /Users/alex/.cargo/registry/src/github.com-1ecc6299db9ec823/cargo-nono-0.1.9/src/check_source.rs:88:18

└─○ cargo nono check --package jf-plonk     
espresso-systems-common: ❌  
  - Did not find a #![no_std] attribute or a simple conditional attribute like #[cfg_attr(not(feature = "std"), no_std)] in the crate source. Crate most likely doesn't support no_std without changes.
rayon: ❌  
  - Did not find a #![no_std] attribute or a simple conditional attribute like #[cfg_attr(not(feature = "std"), no_std)] in the crate source. Crate most likely doesn't support no_std without changes.
  - Source code contains an explicit `use std::` statement.
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: StripPrefixError(())', /Users/alex/.cargo/registry/src/github.com-1ecc6299db9ec823/cargo-nono-0.1.9/src/check_source.rs:88:18
```

plonk/src/testing_apis.rs
nyospe: What about https://crates.io/crates/hashbrown/0.12.3

nyospe (approved):
It would be nice to find a way to make the parallel/not(parallel) per_iter/iter branches inline, so that we don't have potentially diverging duplicate code, but otherwise, this looks great.

alxiong:
> It would be nice to find a way to make the parallel/not(parallel) per_iter/iter branches inline, so that we don't have potentially diverging duplicate code, but otherwise, this looks great.

completely agree! I haven't figured a way yet, but I will try a few things tmr.

alxiong:
I come up with this solution, it should be much cleaner now @nyospe : 

```rust
/// this function helps with slice iterator creation that optionally use
/// `par_iter()` when feature flag `parallel` is on.
///
/// # Usage
/// let v = [1, 2, 3, 4, 5];
/// let sum = parallelizable_slice_iter(&v).sum();
///
/// // above code is a shorthand for (thus equivalent to)
/// #[cfg(feature = "parallel")]
/// let sum = v.par_iter().sum();
/// #[cfg(not(feature = "parallel"))]
/// let sum = v.iter().sum();
#[cfg(feature = "parallel")]
pub(crate) fn parallelizable_slice_iter<T: Sync>(data: &[T]) -> rayon::slice::Iter<T> {
    use rayon::iter::IntoParallelIterator;
    data.into_par_iter()
}

#[cfg(not(feature = "parallel"))]
pub(crate) fn parallelizable_slice_iter<T>(data: &[T]) -> ark_std::slice::Iter<T> {
    use ark_std::iter::IntoIterator;
    data.iter()
}
```

nyospe:
> I come up with [04a6f06](https://github.com/EspressoSystems/jellyfish/pull/85/commits/04a6f06501a4643e8aac1793349725c7bae93783), it should be much cleaner now @nyospe:
> 
> ```rust
> /// this function helps with slice iterator creation that optionally use
> /// `par_iter()` when feature flag `parallel` is on.
> ///
> /// # Usage
> /// let v = [1, 2, 3, 4, 5];
> /// let sum = parallelizable_slice_iter(&v).sum();
> ///
> /// // above code is a shorthand for (thus equivalent to)
> /// #[cfg(feature = "parallel")]
> /// let sum = v.par_iter().sum();
> /// #[cfg(not(feature = "parallel"))]
> /// let sum = v.iter().sum();
> #[cfg(feature = "parallel")]
> pub(crate) fn parallelizable_slice_iter<T: Sync>(data: &[T]) -> rayon::slice::Iter<T> {
>     use rayon::iter::IntoParallelIterator;
>     data.into_par_iter()
> }
> 
> #[cfg(not(feature = "parallel"))]
> pub(crate) fn parallelizable_slice_iter<T>(data: &[T]) -> ark_std::slice::Iter<T> {
>     use ark_std::iter::IntoIterator;
>     data.iter()
> }
> ```
> 
> cc @chancharles92, @zhenfeizhang this commit could need acknowledgement.

🎆

alxiong:
I come across another (even more comprehensive) solution: [`maybe_rayon`](https://github.com/mir-protocol/plonky2/blob/main/maybe_rayon/src/lib.rs) by plonky2.

This is more complete than and basically the generalization of my `parallelizable_slice_iter()`, we should switch to that later.

Issue #89, opened 2022-08-01
[optimization] Improve H2G with BLST


zhenfeizhang:
Current BLS and VRF uses a very naive H2G function. We may consider using BLST for BLS and VRF.

zhenfeizhang:
BLS signature
https://github.com/EspressoSystems/jellyfish/blob/main/primitives/src/signatures/bls.rs
VRF
https://github.com/EspressoSystems/jellyfish/blob/main/primitives/src/vrf/blsvrf.rs


BLST https://github.com/supranational/blst

cc @mrain

PR #90, opened 2022-08-02
Chengyu/blst migration

mrain:
## Description

closes: #89 

 - Use BLS signature/VRF implementation from [blst](https://github.com/supranational/blst) library

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/primitives/src/signatures/bls.rs
+++ b/primitives/src/signatures/bls.rs
@@ -1,77 +1,36 @@
-//! Placeholder for BLS signature.
+//! BLS Signature Scheme
 
 use super::SignatureScheme;
-use crate::{
-    constants::CS_ID_BLS_SIG_NAIVE, errors::PrimitivesError, hash_to_group::SWHashToGroup,
-};
-use ark_ec::{
-    bls12::{Bls12, Bls12Parameters},
-    short_weierstrass_jacobian::GroupAffine,
-    AffineCurve, ModelParameters, ProjectiveCurve,
-};
-use ark_ff::{Fp12, PrimeField};
-use ark_serialize::{CanonicalDeserialize, CanonicalSerialize, SerializationError, Write};
+use crate::{constants::CS_ID_BLS_SIG_NAIVE, errors::PrimitivesError};
+
 use ark_std::{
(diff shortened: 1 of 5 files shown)


primitives/src/vrf/blsvrf.rs
```diff
@@ -76,16 +65,18 @@ where
...
     fn evaluate(
         _pp: &Self::PublicParameter,
         proof: &Self::Proof,
     ) -> Result<Self::Output, PrimitivesError> {
-        let mut proof_serialized = Vec::new();
-        proof.0.serialize_uncompressed(&mut proof_serialized)?;
+        let proof_serialized = proof.serialize();
+        // let mut proof_serialized = Vec::new();
+        // proof.serialize()
+        // proof.0.serialize_uncompressed(&mut proof_serialized)?;
```
alxiong, line 79: remove these

alxiong (commented):
Changes LGTM.

One problem is `blst`'s rust binding is not `no_std` compliant, we can merge this PR first, and will try to submit PR to `blst` to fix that.

primitives/src/signatures/bls.rs
```diff
@@ -89,15 +48,13 @@ where
...
-        let vk = BLSVerKey(
-            GroupAffine::<P::G2Parameters>::prime_subgroup_generator()
-                .mul(sk.0)
-                .into_affine(),
-        );
+        let mut ikm = [0u8; 32];
+        prng.fill_bytes(&mut ikm);
+        let sk = match SecretKey::key_gen(&ikm, &[]) {
+            Ok(sk) => sk,
+            Err(e) => return Err(PrimitivesError::InternalError(format!("{:?}", e))),
```
zhenfeizhang, line 55: +80

can we add a wrapper in `error.rs` and define a new error type for this?

zhenfeizhang (approved):

primitives/src/signatures/bls.rs
mrain: Here the error shall never occur. The only case where it will happen, is when the length of ikm is less than 32.

Issue #91, opened 2022-08-03
[Constraint] Introducing BoolVariable


alxiong:
we could introduce a `struct BoolVariable` which on initialization would enforce it's binary value. This would makes our API for gates like `unpack()` or `logic_gate()` clearer and less error-prone when it comes to deciding when to enforce/check `{0,1}` on the wire variable.

cc @zhenfeizhang

PR #94, opened 2022-08-08
Introduce BoolVar struct

alxiong:
## Description

Introducing `struct BoolVar` whose creation automatically enforces a `bool_gate()`, distinguishing from normal `Variable` to avoid repeated checks and make API clearer.

closes: #91 

---

write a little note why.

- [x] ~Wrote unit tests~

### DIFF
--- a/plonk/src/circuit/basic.rs
+++ b/plonk/src/circuit/basic.rs
@@ -269,7 +269,10 @@ impl<F: FftField> PlonkCircuit<F> {
     /// You should absolutely sure about what you are doing.
     /// You should normally only use this API if you already enforce `v` to be a
     /// boolean value using other constraints.
-    pub(crate) fn create_bool_variable_unchecked(&mut self, a: F) -> Result<BoolVar, PlonkError> {
+    pub(crate) fn create_boolean_variable_unchecked(
+        &mut self,
+        a: F,
+    ) -> Result<BoolVar, PlonkError> {
         let var = self.create_variable(a)?;
         Ok(BoolVar::new_unchecked(var))
     }
--- a/plonk/src/circuit/basic.rs
+++ b/plonk/src/circuit/basic.rs
@@ -715,23 +715,6 @@ impl<F: FftField> PlonkCircuit<F> {
         Ok(())
     }
 
-    // Check whether the variable `var` is a boolean value
-    // this is used to return error to invalid parameter early in the circuit
-    // building development lifecycle, it should NOT be used as a circuit constraint
-    // for which you should use bool_gate() instead
-    #[inline]
-    pub(crate) fn check_bool(&self, var: Variable) -> Result<(), PlonkError> {
-        let val = self.witness(var)?;
-        if val != F::zero() && val != F::one() {
-            Err(ParameterError(
-                "Expecting a boolean value, something is wrong with your circuit logic".to_string(),
-            )
-            .into())
-        } else {
-            Ok(())
-        }
-    }
-
     // Return the variable that maps to a wire `(i, j)` where i is the wire type and
     // j is the gate index. If gate `j` is a padded dummy gate, return zero
     // variable.
--- a/plonk/src/circuit/basic.rs
+++ b/plonk/src/circuit/basic.rs
@@ -5,7 +5,7 @@
 // along with the Jellyfish library. If not, see <https://mit-license.org/>.
 
 //! Basic instantiations of Plonk-based constraint systems
-use super::{Arithmetization, Circuit, GateId, Variable, WireId};
+use super::{Arithmetization, BoolVar, Circuit, GateId, Variable, WireId};
 use crate::{
     circuit::{gates::*, SortedLookupVecAndPolys},
     constants::{compute_coset_representatives, GATE_WIDTH, N_MUL_SELECTORS},
@@ -263,6 +263,15 @@ impl<F: FftField> PlonkCircuit<F> {
     /// The range size of UltraPlonk range gates.
     pub fn range_size(&self) -> Result<usize, PlonkError> {
         Ok(1 << self.range_bit_len()?)
+    }
+
+    /// creating a `BoolVar` without checking if `v` is a boolean value!
+    /// You should absolutely sure about what you are doing.
+    /// You should normally only use this API if you already enforce `v` to be a
+    /// boolean value using other constraints.
+    pub(crate) fn create_bool_variable_unchecked(&mut self, a: F) -> Result<BoolVar, PlonkError> {
+        let var = self.create_variable(a)?;
+        Ok(BoolVar::new_unchecked(var))
     }
 }
 
--- a/plonk/src/circuit/customized/ecc/glv.rs
+++ b/plonk/src/circuit/customized/ecc/glv.rs
@@ -7,7 +7,7 @@
 use crate::{
     circuit::{
         customized::ecc::{MultiScalarMultiplicationCircuit, PointVariable},
-        Circuit, PlonkCircuit, Variable,
+        BoolVar, Circuit, PlonkCircuit, Variable,
     },
     errors::PlonkError,
 };
@@ -93,17 +93,10 @@ where
         self.check_var_bound(scalar)?;
         self.check_point_var_bound(base)?;
 
-        let k_vars = scalar_decomposition_gate::<_, P, _>(self, &scalar)?;
+        let (s1_var, s2_var, s2_sign_var) = scalar_decomposition_gate::<_, P, _>(self, &scalar)?;
 
         let endo_base_var = endomorphism_circuit::<_, P>(self, base)?;
-        multi_scalar_mul_circuit::<_, P>(
-            self,
-            base,
-            k_vars[0],
-            &endo_base_var,
-            k_vars[1],
-            k_vars[2],
-        )
+        multi_scalar_mul_circuit::<_, P>(self, base, s1_var, &endo_base_var, s2_var, s2_sign_var)
     }
 }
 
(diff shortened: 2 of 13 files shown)


chancharles92 (changes requested):
@alxiong Nice work. There is one subtlety that needs to be addressed though.

plonk/src/circuit/customized/ecc/mod.rs
```diff
@@ -852,10 +851,12 @@ mod test {
         let p3 = GroupAffine::<P>::rand(&mut rng);
 
         let mut circuit: PlonkCircuit<F> = PlonkCircuit::new_turbo_plonk();
+        let false_var = circuit.create_bool_variable(false)?;
```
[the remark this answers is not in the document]
alxiong: I'm not sure I fully sympathize with the concern. 
these variables are created by circuit dev, he/she can name it however they want, all `create_bool_variable()` does is making sure the created variable has an underlying value of either `F::zero()` or `F::one()`.

I currently didn't offer API for _constant_ true/false var, which is maybe your `circuit.false()` refers to.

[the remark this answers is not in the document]
chancharles92: Imagine a naive circuit that returns true only if the secret witness of `var` is false. Suppose the circuit is created as follows:
`let false_var = circuit.create_bool_variable(false);`
`let var = circuit.create_variable(F::zero);`
`circuit.equal_gate(false_var, var);`
This circuit will return true no matter whether `wit[var] = wit[false_var] = true` or `wit[var] = wit[false_var] = false`. The right way to build this circuit is
`let false_var = circuit.zero()`
`let var = circuit.create_variable(F::zero);`
`circuit.equal_gate(false_var, var);`.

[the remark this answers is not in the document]
alxiong: sry, I still don't find your example problematic. maybe I miss something.

> This circuit will return true no matter whether wit[var] = wit[false_var] = true or wit[var] = wit[false_var] = false. 

1. `wit[false_var]` was set by the circuit dev (during `create_bool_variable(false)`), not by malicious users who try to provide (instance, witness) to derive a satisfying assignment. 
2. if the circuit dev accidentally change `wit[false_var]`, by the same fashion, he can change `wit[0]` and change the value of `circuit.zero()`

That said, I do see one single difference in the two implementations above:
the latter uses 1 less variable 

iiuc, the analogy of the two approaches is: 
```rust
fn is_positive_gate(cs: &mut Circuit, a: Variable) {
  let zero_var = cs.create_variable(F::zero()); // create a new variable, but of fixed value
  cs.less_than_gate(a, zero_var);
}

fn is_positive_gate(cs: &mut Circuit, a: Variable) {
  let zero_var = cs.zero(); // utilize a existing value of fixed value that I already know
  cs.less_than_gate(a, zero_var);
}
```
both are correct imo.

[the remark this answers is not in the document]
alxiong: we can give circuit dev the option:

```rust
    /// Return a default variable with value `false` (namely zero).
    fn false_var(&self) -> BoolVar {
        BoolVar::new_unchecked(self.zero())
    }

    /// Return a default variable with value `true` (namely one).
    fn true_var(&self) -> BoolVar {
        BoolVar::new_unchecked(self.one())
    }
```

plonk/src/circuit/customized/ecc/mod.rs
```diff
@@ -1198,43 +1168,30 @@ mod test {
         let mut circuit: PlonkCircuit<F> = PlonkCircuit::new_turbo_plonk();
         let p0_var = circuit.create_point_variable(Point::from(p0))?;
         let p1_var = circuit.create_point_variable(Point::from(p1))?;
-        let select_p0 = circuit.binary_point_vars_select(circuit.zero(), &p0_var, &p1_var)?;
+        let true_var = circuit.create_bool_variable(true)?;
+        let false_var = circuit.create_bool_variable(false)?;
```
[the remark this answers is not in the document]
alxiong: update in [6a19265](https://github.com/EspressoSystems/jellyfish/pull/94/commits/6a1926505fbfa3f2b0e065afe3d853708d9c997c) see if you like it.

plonk/src/circuit/customized/ecc/glv.rs
```diff
@@ -521,7 +514,8 @@ where
     };
 
     //  (f.3) either f.1 or f.2 is satisfied
-    let sat = circuit.conditional_select(k2_sign_var, k2_is_neg_sat, k2_is_pos_sat)?;
+    let sat =
+        circuit.conditional_select(k2_sign_var, k2_is_neg_sat.into(), k2_is_pos_sat.into())?;
```
zhenfeizhang, line 518: looks like `k2_is_pos_sat` is an output of `circuit.check_equal(left_var, right_var)?` which makes me think that `circuit.check_equal(left_var, right_var)?` should output a BoolVar instead of a `Variable`. WDYT?

zhenfeizhang:
In general I want to advocate my "circuit ref" idea. For example 
``` rust
struct BoolVar<'a> {
  index: Variable,
  cs: &'a ConstraintSystem
}
```

On top of overloading operators here are some additional thoughts:
- to access the variable, we should use an API `get_index(&self)->Variable` -- i think it is a bit cleaner than `into()`
  - it makes more sense to me to implement Into/From for `Vec<BoolVar>` and `Variable`, in both directions, which are binary (de)composition
- to create new variables we should use associated function `BoolVar::new(cs: &ConstraintSystem, value: bool)` rather than `Circuit::create_bool_variable(...)` and so on... 
- to check variable bound we can simply call `bool_var.check_bound()`, and we do not need to pass parameters

plonk/src/circuit/basic.rs
```diff
@@ -264,6 +264,15 @@ impl<F: FftField> PlonkCircuit<F> {
     pub fn range_size(&self) -> Result<usize, PlonkError> {
         Ok(1 << self.range_bit_len()?)
     }
+
+    /// creating a `BoolVar` without checking if `v` is a boolean value!
+    /// You should absolutely sure about what you are doing.
+    /// You should normally only use this API if you already enforce `v` to be a
+    /// boolean value using other constraints.
+    pub(crate) fn create_bool_variable_unchecked(&mut self, a: F) -> Result<BoolVar, PlonkError> {
```
zhenfeizhang, line 272: slightly prefer `boolean_variable` over `bool_variable`

plonk/src/circuit/customized/ecc/mod.rs
```diff
@@ -263,12 +263,18 @@ where
...
         self.check_point_var_bound(point0)?;
         self.check_point_var_bound(point1)?;
         let x_eq = self.check_equal(point0.0, point1.0)?;
         let y_eq = self.check_equal(point0.1, point1.1)?;
-        self.mul(x_eq, y_eq)
+
+        let res = self.create_bool_variable_unchecked(
+            self.witness(x_eq.into())? * self.witness(y_eq.into())?,
+        )?;
+        self.mul_gate(x_eq.into(), y_eq.into(), res.into())?;
```
chancharles92, line 275: Can we replace `self.mul(x_eq, y_eq)` (in the olde commit) with `self.logic_and(x_eq: BoolVar, y_eq : BoolVar) -> BoolVar`? We can simplify the code here by doing that.

plonk/src/circuit/customized/ecc/glv.rs
alxiong: yeah exactly, currently we have: `pub fn check_equal(&mut self, a: Variable, b: Variable) -> Result<BoolVar, PlonkError> ` which is why you see here `k2_is_pos_sat.into()` since `k2_is_pos_sat` is `BoolVar` type.

plonk/src/circuit/customized/ecc/mod.rs
alxiong: oh yes, good suggestion

plonk/src/circuit/customized/mod.rs
```diff
@@ -831,28 +834,31 @@ pub(crate) mod test {
 
     fn test_logic_or_helper<F: PrimeField>() -> Result<(), PlonkError> {
         let mut circuit: PlonkCircuit<F> = PlonkCircuit::new_turbo_plonk();
-        let zero_var = circuit.zero();
-        let one_var = circuit.one();
+        let false_var = circuit.create_bool_variable(false)?;
+        let true_var = circuit.create_bool_variable(true)?;
```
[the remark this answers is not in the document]
chancharles92: oh i see, right

plonk/src/circuit/customized/ecc/mod.rs
alxiong: done in [5f0cd66](https://github.com/EspressoSystems/jellyfish/pull/94/commits/5f0cd663417b46145cedbd3f620aa8963808ba00)

plonk/src/circuit/basic.rs
alxiong: done in [5f0cd66](https://github.com/EspressoSystems/jellyfish/pull/94/commits/5f0cd663417b46145cedbd3f620aa8963808ba00)

chancharles92 (approved):
LGTM, we can wait for Zhenfei's approval before merging.

plonk/src/circuit/customized/ecc/glv.rs
zhenfeizhang: cool. Sorry I somehow thought the `into()` here was the reversed direction

zhenfeizhang:
LGTM2

Issue #81, opened 2022-07-18
[Constraints]: Comparison gate


alxiong:
Add `lt`, `gt` gates to the constraint system. 

Reference: `r1cs-std`'s [`cmp.rs`](https://github.com/arkworks-rs/r1cs-std/blob/master/src/fields/fp/cmp.rs)

alxiong:
Note: we use `range_check(a, bit_len) -> Vec<BoolVar>` that ensure variable `a \in [0, 2^bit_len)` and return the binary representation of `a` as a building block, this gate costs (`bit_len` boolean constraint + `bit_len / 3` linear combination) constraints 
(see "Range check" section in cap-spec for details, or [code here](https://github.com/EspressoSystems/jellyfish/blob/main/plonk/src/circuit/customized/mod.rs#L680-L684)). 

iiuc, arkworks' [`cmp.rs` constraint](https://github.com/arkworks-rs/r1cs-std/blob/master/src/fields/fp/cmp.rs) logic for checking `a<b` goes as follows: 
1. check MSB of `a` and `b` is `0`, by checking `a<(p-1)/2 && b<(p-1)/2` -- takes 2 `range_check`
2. compute `c = (a-b)*2` in circuit, and decompose into binary representation and return the MSB of `c` -- takes 1 subtraction gate and a `range_check` gate

the constraint works because if `a<b`, then after subtraction, the MSB would be flipped from `0` to `1`, 

_However, the limitation of arkworks' approach is that, they won't allow comparison between two variables lies within `[(p-1)/2, p)`._

---

Fortunately, with a little tweak, we can compare any two field elements.

Idea: **either the `a.MSB = 0 AND b.MSB = 1` or the `|p|-1`-least-significant bits of `a` and `b` follow the constraints above.**

We denote $\ell$ as the bit length of $p$, and assume that to be a multiple of three (if not, we pad with zero)

- first decompose $a, b$ as: 
  ```math
  a= (a_0, a_1, \ldots, a_{\ell-1}), b= (b_0, b_1, \ldots, b_{\ell-1})
  ```
- to check $a_0=0 \land b_0 = 1$: constrain $a_0 * (b_0-1) = 0 \land a_0 + (b_0-1) = 0$
- to check the second half of the statement: reuse the second last `acc` value calculated during decomposition in the first step and compute $a', b'$, essentially 
  ```math
  \begin{matrix}
[...]

PR #96, opened 2022-08-10
Comparison gate

mrain:
## Description

closes: #81 

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/plonk/src/circuit/basic.rs
+++ b/plonk/src/circuit/basic.rs
@@ -1859,10 +1859,12 @@ pub(crate) mod test {
         should_also_check_equality: bool,
     ) -> Result<(), PlonkError> {
         let mut circuit: PlonkCircuit<F> = PlonkCircuit::new_turbo_plonk();
-        let a = circuit.create_variable(F::from(3u32))?;
-        let b = circuit.create_variable(F::from(4u32))?;
+        let val_a = F::from(F::modulus_minus_one_div_two()).add(F::one());
+        let val_b = F::from(F::modulus_minus_one_div_two()).mul(F::from(2u32));
+        let ord = val_a.cmp(&val_b);
+        let a = circuit.create_variable(val_a)?;
+        let b = circuit.create_variable(val_b)?;
         circuit.enforce_cmp(a, b, ordering, should_also_check_equality)?;
-        let ord = 3u32.cmp(&4u32);
         let expected_ok = ord == ordering;
         // Check circuits.
         if expected_ok {
@@ -1934,10 +1936,12 @@ pub(crate) mod test {
         should_also_check_equality: bool,
     ) -> Result<(), PlonkError> {
         let mut circuit: PlonkCircuit<F> = PlonkCircuit::new_turbo_plonk();
-        let a = circuit.create_variable(F::from(3u32))?;
-        let b = circuit.create_variable(F::from(4u32))?;
+        let val_a = F::from(F::modulus_minus_one_div_two()).add(F::one());
+        let val_b = F::from(F::modulus_minus_one_div_two()).mul(F::from(2u32));
+        let ord = val_a.cmp(&val_b);
+        let a = circuit.create_variable(val_a)?;
+        let b = circuit.create_variable(val_b)?;
         let c = circuit.is_cmp(a, b, ordering, should_also_check_equality)?;
-        let ord = 3u32.cmp(&4u32);
         let expected_result = F::from(if ord == ordering { 1u32 } else { 0u32 });
         // Check circuits.
         assert_eq!(circuit.witness(c)?, expected_result);
--- a/plonk/src/circuit/basic.rs
+++ b/plonk/src/circuit/basic.rs
@@ -1559,26 +1559,24 @@ impl<F: PrimeField> PlonkCircuit<F> {
 
         // Iterating from LSB to MSB. Skip the front consecutive 1's.
         // Put an OR gate for bit 0 and an AND gate for bit 1.
-        const_bits_le
+        let mut zipped = const_bits_le
             .into_iter()
             .chain(ark_std::iter::repeat(false))
             .take(a_bits_le.len())
             .zip(a_bits_le.iter())
-            .skip_while(|(b, _)| *b)
-            .try_fold(
-                BoolVar(self.one()),
-                |current, (b, a)| -> Result<BoolVar, PlonkError> {
-                    if b {
-                        self.logic_and(*a, current)
-                    } else {
-                        if current.0 == self.one() {
-                            Ok(*a)
-                        } else {
-                            self.logic_or(*a, current)
-                        }
-                    }
-                },
-            )
+            .skip_while(|(b, _)| *b);
+        if let Some((_, &var)) = zipped.next() {
+            zipped.try_fold(var, |current, (b, a)| -> Result<BoolVar, PlonkError> {
+                if b {
+                    self.logic_and(*a, current)
+                } else {
+                    self.logic_or(*a, current)
+                }
+            })
+        } else {
+            // the constant is all one
+            Ok(BoolVar(self.zero()))
+        }
     }
 }
 
@@ -1807,6 +1805,7 @@ pub(crate) mod test {
 
     fn test_cmp_helper<F: PrimeField>() -> Result<(), PlonkError> {
         let list = [
+            (F::from(5u32), F::from(5u32)),
             (F::from(1u32), F::from(2u32)),
             (
                 F::from(F::modulus_minus_one_div_two()).add(F::one()),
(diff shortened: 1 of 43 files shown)


mrain:
Still incorporating recent changes of Boolean variables.

plonk/src/circuit/basic.rs
```diff
@@ -1416,6 +1488,105 @@ impl<F: PrimeField> PlonkCircuit<F> {
...
+        let msb_check = self.logic_and(a_le_const, b_ge_const)?;
+
+        // Check whether `a` and `b` are both <= (q-1)/2 or
+        // are both > (q-1)/2
+        let msb_eq = self.check_equal(a_le_const.0, b_le_const.0)?;
+        // If so, check whether (b-a) <= (q-1)/2
+        let c = self.sub(b, a)?;
+        let cmp_result =
+            self.is_leq_constant_internal(c, &F::from(F::modulus_minus_one_div_two()))?;
+        let cmp_result = self.logic_and(msb_eq, cmp_result)?;
```
chancharles92, line 1532: Looks like that except for the last line, all other codes are equal to `is_leq_internal()`, can we remove the repetition?

plonk/src/circuit/basic.rs
```diff
@@ -1416,6 +1488,105 @@ impl<F: PrimeField> PlonkCircuit<F> {
...
+    /// than a given constant. Let N = F::size_in_bits()
+    /// It uses N negation gates and N AND/OR gates.
+    fn is_leq_constant_internal(
+        &mut self,
+        a: Variable,
+        constant: &F,
+    ) -> Result<BoolVar, PlonkError> {
+        let a_bits_le = self.unpack(a, F::size_in_bits())?;
+        let const_bits_le = constant.into_repr().to_bits_le();
+        let bits_length = const_bits_le.len().max(a_bits_le.len());
```
chancharles92, line 1547: Is it max or min?

plonk/src/circuit/basic.rs
```diff
@@ -1416,6 +1488,105 @@ impl<F: PrimeField> PlonkCircuit<F> {
...
+                    current_run.clear();
+                    current_run.push(c);
+                } else {
+                    let c = self.logic_neg(*a)?;
+                    current_run.push(c);
+                }
+                Ok(())
+            })?;
+        // If bits length of `a` is too long, check whether the most significant
+        // bits of `a` is 0.
```
chancharles92, line 1573: `is 0` -> `are all zeroes`.

plonk/src/circuit/basic.rs
```diff
@@ -1416,6 +1488,105 @@ impl<F: PrimeField> PlonkCircuit<F> {
...
+
+    /// Helper function to check whether `a` is no greater
+    /// than a given constant. Let N = F::size_in_bits()
+    /// It uses N negation gates and N AND/OR gates.
+    fn is_leq_constant_internal(
+        &mut self,
+        a: Variable,
+        constant: &F,
+    ) -> Result<BoolVar, PlonkError> {
+        let a_bits_le = self.unpack(a, F::size_in_bits())?;
```
chancharles92, line 1545: If we already know that all bit variables are bool, can we use a1 + ... + an = 0 to constrain that all {ai} are zeroes? This will save many gates below.

plonk/src/circuit/basic.rs
```diff
@@ -1416,6 +1488,105 @@ impl<F: PrimeField> PlonkCircuit<F> {
...
+        let const_bits_le = constant.into_repr().to_bits_le();
+        let bits_length = const_bits_le.len().max(a_bits_le.len());
+
+        // Stack for constructing the comparison gate
+        let mut current_run = Vec::new();
+        // Iterate from LSB to MSB. If current bit of constant is 0
```
chancharles92, line 1551: Do we need a stack? Can we just maintain a variable `c` checking whether the prefix is smaller or eq right now? (c = 1 if currently it's larger than the constant's prefix)
When the new b_i = 1, we update `c = a_i &&  c`; when b_i = 0, update `c = a_i || c`

The circuit saves many negation gates and require 1 gate per bit.

plonk/src/circuit/basic.rs
```diff
@@ -1416,6 +1488,105 @@ impl<F: PrimeField> PlonkCircuit<F> {
...
+                }
+                Ok(())
+            })?;
+        // If bits length of `a` is too long, check whether the most significant
+        // bits of `a` is 0.
+        a_bits_le
+            .iter()
+            .skip(bits_length)
+            .try_for_each(|a| -> Result<(), PlonkError> {
+                let c = self.logic_neg(*a)?;
```
chancharles92, line 1578: We can save this negation gates by checking ak + ... + an = 0

chancharles92 (changes requested):

plonk/src/circuit/basic.rs
```diff
@@ -1416,6 +1488,105 @@ impl<F: PrimeField> PlonkCircuit<F> {
...
+                current_run.push(c);
+                Ok(())
+            })?;
+        if current_run.len() == 1 {
+            Ok(current_run[0])
+        } else {
+            self.logic_and_all(&current_run)
```
chancharles92, line 1585: Instead of using a stack, we only need to obtain 2 vars finally, `c` that as mentioned before, and `d` used to check whether `ak, a_{k+1}, ..., a_n` are all zeroes. Finally we return `d && NOT(c)`

plonk/src/circuit/basic.rs
chancharles92: Actually we may still be able to use a stack and save further.

plonk/src/circuit/basic.rs
mrain: True. I'm duplicating this to save one gate. Let me dig into it to see whether we can save more.

plonk/src/circuit/customized/mod.rs
```diff
@@ -361,6 +361,14 @@ where
         Ok(y)
     }
 
+    /// Constrain a variable to be non-zero.
```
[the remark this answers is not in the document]
mrain: Zero. Thanks.

plonk/src/circuit/basic.rs
mrain: Was thinking about it but not sure.

mrain:
Test failed because `F::zero()` is not zero for some parameters.
Try `cargo test test_cmp_gates` here.

@chancharles92 @alxiong

plonk/src/circuit/basic.rs
```diff
@@ -1490,100 +1500,85 @@ impl<F: PrimeField> PlonkCircuit<F> {
...
+            .take(a_bits_le.len())
+            .zip(a_bits_le.iter())
+            .skip_while(|(b, _)| *b)
+            .try_fold(
+                BoolVar(self.one()),
+                |current, (b, a)| -> Result<BoolVar, PlonkError> {
+                    if b {
+                        self.logic_and(*a, current)
+                    } else {
+                        if current.0 == self.one() {
```
chancharles92, line 1574: This circuit implementation is input/witness dependent.

mrain: I don't understand why it's witness dependent here. `b` here is from some constant.

chancharles92: It seems that when b == 0, the gate depends on the value of `current.0`, or did I miss sth?

mrain: `current` is a BoolVar. So `current.0` is just a variable. It's just checking whether `current` is the initial value `BoolVar(self.one())`

chancharles92: Oh you're right, this line will only be triggered at the beginning. But for readability, maybe we can directly write `self.logic_or()`.

chancharles92 (commented):
LGTM, let's wait for @alxiong 's review

plonk/src/circuit/mod.rs
```diff
@@ -150,6 +151,46 @@ pub trait Circuit<F: Field> {
...
+
+    /// Constrain that `a` > `b`.
+    fn enforce_ge(&mut self, a: Variable, b: Variable) -> Result<(), PlonkError>
+    where
+        F: PrimeField;
+
+    /// Constrain that `a` >= `b`.
+    fn enforce_geq(&mut self, a: Variable, b: Variable) -> Result<(), PlonkError>
+    where
+        F: PrimeField;
```
alxiong, line 172: 1. do you think this granular set of API is better than a single `enforce_cmp(a: Variable, b: Variable, ordering: Ordering, enforce_equality: bol)`? (I tend to prefer the compact single API)
2. even if we were to keep these, we should provide default implementation for everything except `enforce_le`, because the rest can be build just using available gate of `trait Circuit`?

mrain: 1. I think the current API set is more clear. Additionally, when you are using it, it's more convenient to write `enforce_le(a, b)` rather than `enforce_cmp(a, b, Ordering::Less, false)` and the code will be more understandable.
2. That's a good idea. Will do.

plonk/src/circuit/customized/mod.rs
```diff
@@ -361,6 +361,14 @@ where
         Ok(y)
     }
 
+    /// Constrain a variable to be zero.
+    /// Return error if the variable is invalid.
+    pub fn zero_gate(&mut self, var: Variable) -> Result<(), PlonkError> {
+        self.check_var_bound(var)?;
+        let wire_vars = [var, 0, 0, 0, 0];
+        self.insert_gate(&wire_vars, Box::new(ZeroGate))
```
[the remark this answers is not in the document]
mrain: 1. Yes
2. Didn't see it before. Will do.

alxiong (approved):
LGTM, @chancharles92 maybe you double check the last commit

chancharles92 (approved):
LGTM

Issue #109, opened 2022-08-16
[Dependency] Remove dependency of jf-plonk from jf-primitives


mrain:
As suggested by @alxiong 
- split out the constraint system definition into its own crate: jf-relation, which includes the PlonkCircuit (aka our constraint system) and our gates over
- then, we keep the proof system (jf_plonk::proof module) as jf-plonk which depends on jf-relation
- our jf-primitives would depend on jf-relation

PR #110, opened 2022-08-17
Dependency refactor

mrain:
## Description

closes: #109 

- Remove `jf-rescue` crate, rescue hash function now resides in `jf-primitives/rescue`.
- Plonk constraint system definition and concrete constructions now live in a standalone crate `jf-relation`.
    - Basic and customized circuit gates are defined in `jf-relation`.
    - Customized/advanced circuit implementations are located in their own crates.
        - Plonk verifier related gadgets, `transcript` and `plonk-verifier` are now in `jf-plonk/circuit`.
        - Primitive gadgets, including `commitment`, `el gamal` etc. remains in `jf-primitives/circuit`.
        - Circuit for rescue hash function is now in `jf-primitives/circuit/rescue`.
- `par-utils` is moved to `jf-utils`.
- New dependency graph
    - `jf-relation` depends on `jf-utils`.
    - `jf-primitives` depends on `jf-relation` and `jf-utils`.
    - `jf-plonk` depends on all other crates.

---

write a little note why.

- [ ] Wrote unit tests


alxiong (commented):
👍 nice work.

here's some comments:
- `plonk/src/circuit/*` should be moved to `jf-relation`, those are not related to Plonk proof system, they are gadgets for plonk verifier (when you try to do proof recursion and prove that a Plonk proof is correct in circuit, therefore they belong with all the other gadgets)
- within `jf-relation`:
  - no need to have `customized/` folder,
  - instead, put all gate impl (including `basic.rs` into `relation/src/gadgets/*`
  - merged `relation/src/customized/gates.rs` into `relation/src/gates.rs`
  - rename `relation/src/customized/mod.rs` to `customized.rs` ? (@chancharles92, @zhenfeizhang any better idea?) or we can also split them up and group them like [arkwork](https://github.com/arkworks-rs/r1cs-std/tree/master/src) or like [plonky2](https://github.com/mir-protocol/plonky2/tree/main/plonky2/src/gates)
  - gadgets moved from `plonk/` should be under `/gadgets/transcript/` and `gadgets/plonk_verifier/`, just like `gadgets/ecc/`

mrain:
> 👍 nice work.
> 
> here's some comments:
> 
> * `plonk/src/circuit/*` should be moved to `jf-relation`, those are not related to Plonk proof system, they are gadgets for plonk verifier (when you try to do proof recursion and prove that a Plonk proof is correct in circuit, therefore they belong with all the other gadgets)

Currently, circuit implementation of plonk verifier depends on struct `VerifyingKey` provided by `jf-plonk`. Moving it to `jf-relation` will cause circular dependency.

> * within `jf-relation`:
>   
>   * no need to have `customized/` folder,
>   * instead, put all gate impl (including `basic.rs` into `relation/src/gadgets/*`
>   * merged `relation/src/customized/gates.rs` into `relation/src/gates.rs`
>   * rename `relation/src/customized/mod.rs` to `customized.rs` ? (@chancharles92, @zhenfeizhang any better idea?) or we can also split them up and group them like [arkwork](https://github.com/arkworks-rs/r1cs-std/tree/master/src) or like [plonky2](https://github.com/mir-protocol/plonky2/tree/main/plonky2/src/gates)
>   * gadgets moved from `plonk/` should be under `/gadgets/transcript/` and `gadgets/plonk_verifier/`, just like `gadgets/ecc/`

Good suggestion. Will do.

plonk/Cargo.toml
```diff
@@ -57,4 +58,4 @@ harness = false
 default = [ "parallel" ]
 std = [ "ark-std/std", "ark-serialize/std", "ark-ff/std", "ark-ec/std", "ark-poly/std"]
 test_apis = [] # exposing apis for testing purpose
-parallel = [ "ark-ff/parallel", "ark-ec/parallel", "ark-poly/parallel", "rayon" ]
+parallel = [ "ark-ff/parallel", "ark-ec/parallel", "ark-poly/parallel", "jf-utils/parallel", "jf-relation/parallel", "jf-primitives/parallel", "rayon" ]
```
zhenfeizhang, line 61: nits: can wen separate this into multiple lines?

plonk/src/constants.rs
```diff
@@ -14,10 +14,8 @@ use rand_chacha::ChaChaRng;
...
 // ==========================
 
-/// The number of input wires.
-pub const GATE_WIDTH: usize = 4;
-/// The number of multiplication selectors.
-pub(crate) const N_MUL_SELECTORS: usize = 2;
+// Gate Width
+pub use jf_relation::constants::GATE_WIDTH;
 /// The number of TurboPlonk selectors.
 pub(crate) const N_TURBO_PLONK_SELECTORS: usize = 13;
```
zhenfeizhang, line 20: should this be also in `jf_relation`? it seems that this and `GATE_WIDTH` should be in a same crate

primitives/Cargo.toml
```diff
@@ -60,4 +65,4 @@ harness = false
 [features]
 default = [ "parallel" ]
 std = []
-parallel = [ "jf-plonk/parallel", "rayon" ]
+parallel = [ "ark-ff/parallel", "ark-ec/parallel", "jf-utils/parallel", "jf-relation/parallel", "rayon" ]
```
zhenfeizhang, line 68: nits: multiple lines

relation/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+bincode = "1.0"
+ark-ed-on-bls12-381 = "0.3.0"
+ark-ed-on-bls12-377 = { git = "https://github.com/arkworks-rs/curves", rev = "677b4ae751a274037880ede86e9b6f30f62635af" }
+ark-ed-on-bls12-381-bandersnatch = { git = "https://github.com/arkworks-rs/curves", rev = "677b4ae751a274037880ede86e9b6f30f62635af" }
+ark-ed-on-bn254 = "0.3.0"
+hex = "^0.4.3"
+
+[features]
+default = [ "parallel" ]
+std = [ "ark-std/std", "ark-serialize/std", "ark-ff/std", "ark-ec/std", "ark-poly/std"]
```
zhenfeizhang, line 42: nit: multi lines

relation/src/constants.rs
```diff
@@ -0,0 +1,82 @@
...
+use rand_chacha::ChaChaRng;
+
+// ==========================
+// Circuit-related constants.
+// ==========================
+
+/// The number of input wires.
+pub const GATE_WIDTH: usize = 4;
+/// The number of multiplication selectors.
+pub(crate) const N_MUL_SELECTORS: usize = 2;
```
zhenfeizhang, line 20: expose this outside of the crate?

relation/src/customized/ecc/mod.rs
```diff
@@ -138,15 +135,18 @@ where
...
         let x = self.witness(point_var.0)?;
         let y = self.witness(point_var.1)?;
         Ok(Point(x, y))
     }
 
     /// Add a new EC point (as witness) to the circuit
-    pub fn create_point_variable(&mut self, point: Point<F>) -> Result<PointVariable, PlonkError> {
+    pub fn create_point_variable(
+        &mut self,
+        point: Point<F>,
```
zhenfeizhang, line 148: `point: &Point<F>`

zhenfeizhang:
> * rename `relation/src/customized/mod.rs` to `customized.rs` ? (@chancharles92, @zhenfeizhang any better idea?) or we can also split them up and group them like [arkwork](https://github.com/arkworks-rs/r1cs-std/tree/master/src) or like [plonky2](https://github.com/mir-protocol/plonky2/tree/main/plonky2/src/gates)

Good idea... now the `customized` is a bit large and it makes sense to group its components

chancharles92:
>   * rename `relation/src/customized/mod.rs` to `customized.rs` ? (@chancharles92, @zhenfeizhang any better idea?) or we can also split them up and group them like [arkwork](https://github.com/arkworks-rs/r1cs-std/tree/master/src) or like [plonky2](https://github.com/mir-protocol/plonky2/tree/main/plonky2/src/gates)
Agree with @zhenfeizhang , I prefer the approach used by arkworks and plonky2

alxiong:
> Currently, circuit implementation of plonk verifier depends on struct VerifyingKey provided by jf-plonk. Moving it to jf-relation will cause circular dependency.

Ah, I see, then just keep them where there are.

relation/src/gadgets/ecc/mod.rs
```diff
@@ -133,20 +129,20 @@ impl PointVariable {
...
         let x = self.witness(point_var.0)?;
         let y = self.witness(point_var.1)?;
         Ok(Point(x, y))
     }
 
     /// Add a new EC point (as witness) to the circuit
-    pub fn create_point_variable(&mut self, point: Point<F>) -> Result<PointVariable, PlonkError> {
+    pub fn create_point_variable(
+        &mut self,
+        point: &Point<F>,
```
alxiong, line 144: I think move the ownership of `point` (consume the input) is a good idea generally. 
IMO, the workflow generally goes like you construct a point through whatever step, then finally create a variable and from then on, you should only deal with `PointVar`. 
I can't think of a reason why would you still want the original `Point`. 

^^ maybe opinionated, would you agree? @zhenfeizhang @chancharles92

relation/src/gadgets/logic.rs
```diff
@@ -0,0 +1,424 @@
...
+        self.check_var_bound(a.into())?;
+        self.check_var_bound(b.into())?;
+        let wire_vars = &[a.into(), b.into(), 0, 0, 0];
+        self.insert_gate(wire_vars, Box::new(LogicOrGate))?;
+        Ok(())
+    }
+
+    /// Obtain a bool variable representing whether two input variables are
+    /// equal. Return error if variables are invalid.
+    pub fn is_equal(&mut self, a: Variable, b: Variable) -> Result<BoolVar, CircuitError> {
```
alxiong, line 30: I remember in the very beginning we name it `is_equal()` but later either because of clippy or some other reason we change to `check_equal()`

@zhenfeizhang @chancharles92  can you remember why we make the change previously? 
(altho this is minor imo, doesn't hurt changing back)

alxiong (commented):
I really like your reorg! great work!  🎉 
Thank you for the big migration!

Some minor comments: 
- `relation/gadgets/basic.rs` seems a bit strange, I would merge this file with `relation/circuit.rs` and call it `relation/constraint_system.rs` (which contains both the `trait Circuit` and `struct PlonkCircuit`) <-- I agree our usage of circuit/constraint_system/gadget/gates need to be re-checked later.

relation/src/gadgets/ecc/mod.rs
chancharles92: The scenario I can think of is when the point is a public variable, then we'd still like keep the point so that we can later pass it as a public input. Do we also have an API `create_public_point_variable()`?

mrain: Yes there's a `create_public_point_variable()`. And in the original codebase, it still consumes the ownership.

relation/src/gadgets/logic.rs
mrain: This is for constancy with naming of comparison gates, like `is_lt`. 
The same issue occurs between `*_gate` and `enforce_*`. They essentially represent the same type of constraints. IMO the later one is more straightforward. But it doesn't make sense to rename `logic_or_gate` to `enforce_logic_or`. Or we could leave `logic_or_gate` as is, and change other ones to `enforce_*`?

relation/src/gadgets/ecc/mod.rs
alxiong: > when the point is a public variable, then we'd still like keep the point so that we can later pass it as a public input. Do we also have an API create_public_point_variable()?

when a point is supposed to be a public input, you will consume it using `create_public_point_variable()`, usually (i mean most of the time) you won't create both a witness point and a PI point using the _same_ point. 
in the rare case that you do, you can always clone. 

> According to https://github.com/EspressoSystems/jellyfish/pull/107, can we still pass in the reference here?

that PR is about passing in `rng: impl RngCore + CryptoRng`, not other struct, so won't affect any decision here.

relation/src/gadgets/logic.rs
alxiong: consistency is good and indeed some of our gadget naming is not. 

I also prefer `is_equal()`, just can't recall why we change it in the first place, I'd say go ahead and use `is_equal()`.

I also prefer `enforce_*()` over `*_gate()`

mrain: Okay let me work on this change. Using `is_*` and `enforce_*`. Except for some arithmetic operations like `add_gate`.

alxiong (dismissed):
LGTM!

plonk/src/circuit/plonk_verifier/mod.rs
```diff
@@ -829,4 +832,24 @@ mod test {
 
         Ok(())
     }
+
+    // two circuit with the same statement should have the same extended permutation
+    // polynomials even with different variable assignment
+    // TODO(Chengyu): Copied from crate `jf-relation`, can we remove this
```
chancharles92, line 838: Maybe we can set this API public in `jf-relation` so that we can call it in other crates? WDYT? @alxiong @zhenfeizhang

chancharles92 (commented):
Good job! Only 1 minor comment

plonk/src/circuit/plonk_verifier/mod.rs
mrain: I tried but failed before. Maybe because this function lies in `mod test`?

chancharles92: I see, one approach is to move the API out of `mod test`, and use sth like this:
```
#[cfg(test)]
pub fn foo() { ... }

#[cfg(not(test))]
fn foo() { ... }
```
I'm not sure if this is the best practice though.

relation/Cargo.toml
```diff
@@ -0,0 +1,46 @@
...
+ark-ed-on-bls12-377 = { git = "https://github.com/arkworks-rs/curves", rev = "677b4ae751a274037880ede86e9b6f30f62635af" }
+ark-ed-on-bls12-381-bandersnatch = { git = "https://github.com/arkworks-rs/curves", rev = "677b4ae751a274037880ede86e9b6f30f62635af" }
+ark-ed-on-bn254 = "0.3.0"
+hex = "^0.4.3"
+
+[features]
+default = [ "parallel" ]
+std = [ "ark-std/std", "ark-serialize/std", "ark-ff/std", 
+        "ark-ec/std", "ark-poly/std"]
+test_apis = [] # exposing apis for testing purpose
```
zhenfeizhang, line 44: seems like those features are no longer needed

plonk/src/circuit/plonk_verifier/mod.rs
mrain: Moved out. Thanks.

chancharles92 (approved):

plonk/src/circuit/plonk_verifier/mod.rs
alxiong: yes, Binyi's solution is good for now. 
in the future, when we want to expose bunch of helper functions for testing, we can put them under:

```rust
#[cfg(test)]
mod test_utils;

// in test_utils
pub fn reuseable_test_helpers() {}
```

so that these functions can only be used in testing modules of other crates.

Issue #62, opened 2022-05-03
audit/improve KZG10 code


zhenfeizhang:
the main bottleneck of Jellyfish currently are:
- FFT
- KZG commitment

And KZG dominants.

It appears that arkwork's KZG implementation may not be optimized, for example:
https://github.com/arkworks-rs/poly-commit/pull/98

The non-optimization in the example is not critical to us yet because it is for the verifier, and CAPE uses solidity verifier rather than rust verifier (cc @philippecamacho).

zhenfeizhang:
we will use our own PCS APIs for this

zhenfeizhang:
- move HP PCS to Jellyfish/primitive
- update Jellyfish/plonk to new PCS
- implement univariate KZG batching for PCS (different PR)
- update jellyfish/plonk with pcs batching (different PR)

PR #116, opened 2022-08-26
New PCS trait and implementation

mrain:
## Description

closes: #62 

- Introduct new `PolynomialCommitmentScheme` trait and basic implementations
- Now `PlonkKzgSnark` use our own KZG10 implementation

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/primitives/src/pcs/mod.rs
+++ b/primitives/src/pcs/mod.rs
@@ -1,0 +1,149 @@
+// Copyright (c) 2022 Espresso Systems (espressosys.com)
+// This file is part of the Jellyfish library.
+
+// You should have received a copy of the MIT License
+// along with the Jellyfish library. If not, see <https://mit-license.org/>.
+
+//! Polynomial Commitment Scheme
+pub mod errors;
+mod multilinear_kzg;
+pub mod prelude;
+mod structs;
+mod transcript;
+mod univariate_kzg;
+
+use ark_ec::PairingEngine;
+use ark_serialize::CanonicalSerialize;
+use ark_std::{rand::RngCore, vec::Vec};
+use errors::PCSError;
+
+/// This trait defines APIs for polynomial commitment schemes.
+/// Note that for our usage of PCS, we do not require the hiding property.
+pub trait PolynomialCommitmentScheme<E: PairingEngine> {
+    /// Prover parameters
+    type ProverParam;
+    /// Verifier parameters
+    type VerifierParam;
+    /// Structured reference string
+    type SRS;
+    /// Polynomial and its associated types
+    type Polynomial;
+    /// Polynomial input domain
+    type Point;
+    /// Polynomial Evaluation
+    type Evaluation;
+    /// Commitments
+    type Commitment: CanonicalSerialize;
+    /// Batch commitments
+    type BatchCommitment: CanonicalSerialize;
+    /// Proofs
+    type Proof;
+    /// Batch proofs
+    type BatchProof;
+
+    /// Build SRS for testing.
+    ///
+    /// - For univariate polynomials, `log_size` is the log of maximum degree.
+    /// - For multilinear polynomials, `log_size` is the number of variables.
+    ///
+    /// WARNING: THIS FUNCTION IS FOR TESTING PURPOSE ONLY.
+    /// THE OUTPUT SRS SHOULD NOT BE USED IN PRODUCTION.
+    fn gen_srs_for_testing<R: RngCore>(rng: &mut R, log_size: usize)
+        -> Result<Self::SRS, PCSError>;
+
+    /// Trim the universal parameters to specialize the public parameters.
+    /// Input both `supported_log_degree` for univariate and
+    /// `supported_num_vars` for multilinear.
+    fn trim(
+        srs: &Self::SRS,
+        supported_log_degree: usize,
+        supported_num_vars: Option<usize>,
+    ) -> Result<(Self::ProverParam, Self::VerifierParam), PCSError>;
+
+    /// Generate a commitment for a polynomial
+    fn commit(
+        prover_param: &Self::ProverParam,
+        poly: &Self::Polynomial,
+    ) -> Result<Self::Commitment, PCSError>;
+
+    /// Generate a commitment for a list of polynomials
+    fn multi_commit(
+        prover_param: &Self::ProverParam,
+        polys: &[Self::Polynomial],
+    ) -> Result<Self::BatchCommitment, PCSError>;
+
+    /// On input a polynomial `p` and a point `point`, outputs a proof for the
+    /// same.
+    fn open(
+        prover_param: &Self::ProverParam,
+        polynomial: &Self::Polynomial,
+        point: &Self::Point,
+    ) -> Result<(Self::Proof, Self::Evaluation), PCSError>;
+
+    /// Input a list of MLEs, and a same number of points, and a transcript,
+    /// compute a multi-opening for all the polynomials.
+    fn multi_open(
+        prover_param: &Self::ProverParam,
+        multi_commitment: &Self::Commitment,
+        polynomials: &[Self::Polynomial],
+        points: &[Self::Point],
+    ) -> Result<(Self::BatchProof, Vec<Self::Evaluation>), PCSError>;
+
+    /// Verifies that `value` is the evaluation at `x` of the polynomial
+    /// committed inside `comm`.
+    fn verify(
+        verifier_param: &Self::VerifierParam,
+        commitment: &Self::Commitment,
+        point: &Self::Point,
+        value: &E::Fr,
+        proof: &Self::Proof,
+    ) -> Result<bool, PCSError>;
+
+    /// Verifies that `value_i` is the evaluation at `x_i` of the polynomial
+    /// `poly_i` committed inside `comm`.
+    fn batch_verify<R: RngCore>(
+        verifier_param: &Self::VerifierParam,
+        multi_commitment: &Self::BatchCommitment,
+        points: &[Self::Point],
+        values: &[E::Fr],
+        batch_proof: &Self::BatchProof,
+        rng: &mut R,
+    ) -> Result<bool, PCSError>;
+}
+
+/// API definitions for structured reference string
+pub trait StructuredReferenceString<E: PairingEngine>: Sized {
+    /// Prover parameters
+    type ProverParam;
+    /// Verifier parameters
+    type VerifierParam;
+
+    /// Extract the prover parameters from the public parameters.
+    fn extract_prover_param(&self, supported_log_size: usize) -> Self::ProverParam;
+    /// Extract the verifier parameters from the public parameters.
+    fn extract_verifier_param(&self, supported_log_size: usize) -> Self::VerifierParam;
+
+    /// Trim the universal parameters to specialize the public parameters
+    /// for polynomials to the given `supported_log_size`, and
+    /// returns committer key and verifier key.
+    ///
+    /// - For univariate polynomials, `supported_log_size` is the log of maximum
+    ///   degree.
+    /// - For multilinear polynomials, `supported_log_size` is the number of
+    ///   variables.
+    ///
+    /// `supported_log_size` should be in range `1..=params.log_size`
+    fn trim(
+        &self,
+        supported_log_size: usize,
+    ) -> Result<(Self::ProverParam, Self::VerifierParam), PCSError>;
+
+    /// Build SRS for testing.
+    ///
+    /// - For univariate polynomials, `log_size` is the log of maximum degree.
+    /// - For multilinear polynomials, `log_size` is the number of variables.
+    ///
+    /// WARNING: THIS FUNCTION IS FOR TESTING PURPOSE ONLY.
+    /// THE OUTPUT SRS SHOULD NOT BE USED IN PRODUCTION.
+    fn gen_srs_for_testing<R: RngCore>(rng: &mut R, log_size: usize) -> Result<Self, PCSError>;
+}
--- a/primitives/src/pcs/mod.rs
+++ b/primitives/src/pcs/mod.rs
@@ -13,9 +13,10 @@ mod transcript;
 mod univariate_kzg;
 
 use ark_ec::PairingEngine;
-use ark_serialize::CanonicalSerialize;
+use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 use ark_std::{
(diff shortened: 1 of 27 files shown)


primitives/src/pcs/mod.rs
```diff
@@ -0,0 +1,155 @@
...
+    /// Build SRS for testing.
+    ///
+    /// - For univariate polynomials, `supported_size` is the maximum degree.
+    /// - For multilinear polynomials, `supported_size` is the number of
+    ///   variables.
+    ///
+    /// WARNING: THIS FUNCTION IS FOR TESTING PURPOSE ONLY.
+    /// THE OUTPUT SRS SHOULD NOT BE USED IN PRODUCTION.
+    fn gen_srs_for_testing<R: RngCore>(
+        rng: &mut R,
```
alxiong, line 53: we should use marker trait `CryptoRng` (even tho it didn't affect anything, it's at least marking and indicating that users should pass in a cryptographic RNG.
i.e. change to `R: RngCore + CryptoRng`, similar to other occurrences.

primitives/src/pcs/mod.rs
```diff
@@ -0,0 +1,155 @@
...
+    ) -> Result<Self::SRS, PCSError>;
+
+    /// Trim the universal parameters to specialize the public parameters.
+    /// Input both `supported_degree` for univariate and
+    /// `supported_num_vars` for multilinear.
+    fn trim(
+        srs: &Self::SRS,
+        supported_degree: usize,
+        supported_num_vars: Option<usize>,
+    ) -> Result<(Self::ProverParam, Self::VerifierParam), PCSError>;
```
alxiong, line 64: imo, input parameters that are potentially huge (exceed stack capacity of resource limited machine) should use type like: 

```rust
fn trim(srs: impl Borrow<Self::SRS>, ...);
fn commit(prover_param: impl Borrow<Self::ProverParam>, ...);
```

so this primarily referring to SRS and prover_param. other structs should be fine.
the syntax above allow for any types of borrow, from normal in-memory reference, to smart pointer, to atomic reference counters etc.

here's a simple illustration:
```rust
use std::borrow::Borrow;
use std::rc::Rc;
use std::sync::Arc;

fn test(_srs: impl Borrow<String>){}

fn main() {
    let srs = String::from("hi");
    let srs_boxed = Box::new(srs.clone());
    let srs_rc = Rc::new(srs.clone());
    let srs_arc = Arc::new(srs.clone());
    
    test(&srs);
    test(srs_boxed);
    test(srs_arc);
    test(srs_rc);
}
```

primitives/src/pcs/prelude.rs
```diff
@@ -0,0 +1,21 @@
...
+// You should have received a copy of the MIT License
+// along with the Jellyfish library. If not, see <https://mit-license.org/>.
+
+//! Prelude
+pub use crate::pcs::{
+    errors::PCSError,
+    multilinear_kzg::{
+        srs::{MultilinearProverParam, MultilinearUniversalParams, MultilinearVerifierParam},
+        util::{compute_qx_degree, get_batched_nv, merge_polynomials},
+        BatchProof, KZGMultilinearPCS, Proof,
```
alxiong, line 13: `BatchProof` and `Proof` are not exactly intelligible for the downstream.

either we rename them to `MultilinearKZGBatchProof` and `MultilinearKZGProof` 
or we use alias re-export `pub use multilinear_kzg::BatchProof as MultilinearKZGBatchProof`.

primitives/src/pcs/prelude.rs
```diff
@@ -0,0 +1,21 @@
...
+    },
+    structs::Commitment,
+    univariate_kzg::{
+        srs::{UnivariateProverParam, UnivariateUniversalParams, UnivariateVerifierParam},
+        KZGUnivariateOpening, KZGUnivariatePCS,
```
alxiong, line 18: Speaking of which, I find the naming slightly different from my preference (obviously it's a matter of preference, nothing major)

I find `UniveriateKZGxxx` easier to read than `KZGUniveriatexxx`

primitives/src/pcs/structs.rs
```diff
@@ -0,0 +1,24 @@
...
+    Clone(bound = ""),
+    Copy(bound = ""),
+    Debug(bound = ""),
+    PartialEq(bound = ""),
+    Eq(bound = "")
+)]
+/// A commitment is an Affine point.
+pub struct Commitment<E: PairingEngine>(
+    /// the actual commitment is an affine point.
+    pub E::G1Affine,
```
alxiong, line 23: feel like I asked this question before, can @zhenfeizhang reminds me why we need this wrapper struct again?

primitives/src/pcs/multilinear_kzg/mod.rs
```diff
@@ -0,0 +1,488 @@
...
+    /// openings of q(x) at 1, omega, ..., and r
+    pub q_x_opens: Vec<KZGUnivariateOpening<E>>,
+}
+
+impl<E: PairingEngine> PolynomialCommitmentScheme<E> for KZGMultilinearPCS<E> {
+    // Parameters
+    type ProverParam = (
+        MultilinearProverParam<E>,
+        UnivariateProverParam<E::G1Affine>,
+    );
```
alxiong, line 66: Q: is `powers_of_g` in `MultilinearProverParam` and in `UnivariateProverParam` the same thus sharable? or are they used for different purposes?

(maybe I should re-review these implementations when I digest multilinear kzg scheme later.)

primitives/src/pcs/mod.rs
```diff
@@ -0,0 +1,155 @@
...
+    ) -> Result<(Self::Proof, Self::Evaluation), PCSError>;
+
+    /// Input a list of MLEs, and a same number of points, and a transcript,
+    /// compute a multi-opening for all the polynomials.
+    fn multi_open(
+        prover_param: &Self::ProverParam,
+        multi_commitment: &Self::Commitment,
+        polynomials: &[Self::Polynomial],
+        points: &[Self::Point],
+    ) -> Result<(Self::BatchProof, Vec<Self::Evaluation>), PCSError>;
```
alxiong, line 93: 1. what is `MLEs` in the code comments?
2. what is a `multi_commitment`? should explain in comment, and explain
   - what API is used to generate such a `multi_comm`
   - how is different from `Self::BatchCommitment`
3. why `multi_open(multi_commitment: &Self::Commitment)`, but `batch_verify(multi_commitment: &Self::BatchCommitment)` ?

primitives/src/pcs/mod.rs
```diff
@@ -0,0 +1,155 @@
...
+    /// Polynomial Evaluation
+    type Evaluation;
+    /// Commitments
+    type Commitment: CanonicalSerialize;
+    /// Batch commitments
+    type BatchCommitment: CanonicalSerialize;
+    /// Proofs
+    type Proof;
+    /// Batch proofs
+    type BatchProof;
```
alxiong, line 42: Can add more trait bounds such as: 
`type Proof: Clone + Serialize + CanonicalSerialize + Debug + PartialEq...`

can take a look at [PairingEngine's associated type](https://docs.rs/ark-ec/0.3.0/ark_ec/trait.PairingEngine.html), they inherented a lot of basic traits.

primitives/src/pcs/mod.rs
mrain: One question: shall we make a similar change for other potentially large types? Like `Self::Polynomial`, `Self::BatchProof` etc?

chancharles92: I'm neutral for `Self::Polynomial`. It can be large when the degree is high, possibly ~|SRS|/2 because it consists of field elements not group elements. On the other hand putting it in heaps might affect prover's efficiency.

For `Self::BatchProof` I'd say no because it won't be huge.

mrain: Alright. I'll just update it for `SRS` and `ProverParam`.

primitives/src/pcs/multilinear_kzg/mod.rs
```diff
@@ -25,40 +25,48 @@ use ark_ff::PrimeField;
...
+    vec::Vec,
+    One, Zero,
 };
 use batching::{batch_verify_internal, multi_open_internal};
 use srs::{MultilinearProverParam, MultilinearUniversalParams, MultilinearVerifierParam};
 use util::merge_polynomials;
 
 /// KZG Polynomial Commitment Scheme on multilinear polynomials.
-pub struct KZGMultilinearPCS<E: PairingEngine> {
+pub struct MultilinearKZGPCS<E: PairingEngine> {
```
alxiong, line 44: sorry to be picky, what about camel-case some abbreviation when there's multiple of them?

`MultilinearKZGPCS` -> `MultilinearKzgPCS` or `MultilinearKzgPcs` (I prefer the first.
(similar comment to other names)

primitives/src/pcs/mod.rs
```diff
@@ -58,36 +62,36 @@ pub trait PolynomialCommitmentScheme<E: PairingEngine> {
...
     fn trim(
-        srs: &Self::SRS,
+        srs: impl Borrow<Self::SRS>,
         supported_degree: usize,
         supported_num_vars: Option<usize>,
     ) -> Result<(Self::ProverParam, Self::VerifierParam), PCSError>;
 
     /// Generate a commitment for a polynomial
     fn commit(
-        prover_param: &Self::ProverParam,
```
alxiong: maybe fore the first occurrence of `srs` and `prover_param`, let's add some note in the code comments that: 

```
\\\ ## Note on function signature 
\\\ Usually, data structure like SRS and ProverParam are huge and users might wish to keep them in heap using different kinds of smart pointers (instead of only in stack)
\\\ therefore our `impl Borrow<_>` interface allows for passing in any pointer type, e.g.:
\\\ `trim(srs: &Self::SRS, ..)` or `trim(srs: Box<Self::SRS>, ..)` or `trim(srs: Arc<Self::SRS>, ..)` etc.
```

primitives/src/pcs/mod.rs
alxiong: @mrain, @chancharles92  maybe use this?
```rust
    /// Prover parameters
    type ProverParam: Clone;
    /// Verifier parameters
    type VerifierParam: Clone + CanonicalSerialize;
    /// Structured reference string
    type SRS: Clone + Debug;
    /// Polynomial and its associated types
    type Polynomial: ark_poly::polynomial::Polynomial<E::Fr>;
    /// Polynomial input domain
    type Point: Clone + Copy + Debug + PartialEq + Eq;
    /// Polynomial Evaluation
    type Evaluation: Field;
    /// Commitments
    type Commitment: Clone + CanonicalSerialize + CanonicalDeserialize + Debug + PartialEq + Eq;
    /// Batch commitments
    type BatchCommitment: Clone + CanonicalSerialize + CanonicalDeserialize + Debug + PartialEq + Eq;
    /// Proofs
    type Proof: Clone + CanonicalSerialize + CanonicalDeserialize + Debug + PartialEq + Eq;
    /// Batch proofs
    type BatchProof: Clone + CanonicalSerialize + CanonicalDeserialize + Debug + PartialEq + Eq;
```

primitives/src/pcs/multilinear_kzg/mod.rs
alxiong: A: (from Binyi) they can be the same, but conceptually different, plus `univeriate_prover_param` will be relatively small

primitives/src/pcs/mod.rs
alxiong: - `type Polynomial` should be greatest common subset of traits bound from [`Polynomial`](https://docs.rs/ark-poly/0.3.0/ark_poly/polynomial/trait.Polynomial.html) and [`MultilinearExtension`](https://docs.rs/ark-poly/0.3.0/ark_poly/evaluations/multivariate/multilinear/trait.MultilinearExtension.html).
- `type Point` maybe exclude `Copy`

primitives/src/pcs/transcript.rs
```diff
@@ -0,0 +1,107 @@
+// Copyright (c) 2022 Espresso Systems (espressosys.com)
```
chancharles92, line 1: Why do we need transcript in the PCS crate?

mrain: In `jf-plonk` there's a more general transcript implementation. However some transcript sub-modules depend on things defined in `jf-plonk` so we cannot use it in `jf-primitives`.
We could put a TODO here.

chancharles92: I see. On the other hand, where is the place that we use transcript in PCS? I'm wondering if we can remove the use of transcript in PCS.

mrain: The batching part. Right now only for multilinear settings.

chancharles92: I see, thanks. Let's put a TODO here then.

primitives/src/pcs/mod.rs
mrain: Thanks @alxiong . But for `Polynomial`, `Rc` used in multilinear setting could not fit in this trait bound.
I think it make sense to put polynomial in heap because it's usually quite large.

alxiong: I agree, but similar to the argument for `impl Borrow<>`, how do you know what type of smart pointer one wants?

I don't think we should fix smart pointer type at the associate type level, (namely `MultilinearKzgPCS::Polynomial = Rc<DenseMultilinearExtension<E::Fr>>` is not as flexible as just `DenseMultilinearExtension<E::Fr>`. 

and you can always wrap this MLE into some smart pointer, such as `Rc`, `RefCell`, or `Arc` etc.

@chancharles92 what's your opinion?

alxiong (approved):
I give a provisional approval on my comments https://github.com/EspressoSystems/jellyfish/pull/116#discussion_r960226029 being settled.

primitives/src/pcs/mod.rs
chancharles92: I don't have a strong opinion. Actually we were using `Polynomial = DenseMultilinearExtension`, but later Zhenfei changed it to `Polynomial = Rc<Dense...>` which has a few advantages:
1. It significantly simplifies the code, o/w we need to add Rc everywhere.
2. It make sure that the developer never forget to use `Rc`, otherwise the code performance can be accidentally effected.
3. When we want to use other smart pointers, we can e.g. use `Polynomial = Arc<...>` in associated type rather than change it everywhere.

Issue #104, opened 2022-08-12
[Constraint] compare to constant gate


mrain:
Need to implement public APIs for comparing to constants. E.g. `is_lt_const`, `enforce_gt_const`

zjw110:
good luck

PR #117, opened 2022-08-30
Gadgets for comparing to constant

mrain:
## Description

closes: #104 
Some unfinished work for #96 and #81 

---

write a little note why.

- [x] Wrote unit tests
- [ ] ~~Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md`~~

### DIFF
--- a/relation/src/gadgets/cmp.rs
+++ b/relation/src/gadgets/cmp.rs
@@ -265,6 +265,7 @@ mod test {
     use ark_ed_on_bn254::Fq as FqEd254;
     use ark_ff::PrimeField;
     use ark_std::cmp::Ordering;
+    use itertools::multizip;
 
     #[test]
     fn test_cmp_gates() -> Result<(), CircuitError> {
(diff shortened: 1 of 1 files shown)


relation/src/gadgets/cmp.rs
```diff
@@ -316,4 +419,92 @@ mod test {
...
+        let mut circuit = PlonkCircuit::<F>::new_turbo_plonk();
+        let expected_result = if a < b { F::one() } else { F::zero() };
+        let a = circuit.create_variable(*a)?;
+
+        let c = circuit.is_lt_constant(a, *b)?;
+        assert!(circuit.witness(c.into())?.eq(&expected_result));
+        assert!(circuit.check_circuit_satisfiability(&[]).is_ok());
+        Ok(())
+    }
+    fn test_is_leq_constant<F: PrimeField>(a: &F, b: &F) -> Result<(), CircuitError> {
```
alxiong, line 432: can we unify these to a single function instead?
`test_cmp(a: &F, b: &F, order: Ordering, also_check_equality: bool, is_b_constant: bool)`

alxiong (commented):
Generally LGTM, just a small comment for smaller code repetition in test.

relation/src/gadgets/cmp.rs
chancharles92: Agree, now I also see the advantage of having a unified API `is_cmp(...)` and `enforce_cmp()` as it can remove code repetition. Not sure if we want to change the APIs, but agree that we should do it at least in tests.

relation/src/gadgets/cmp.rs
```diff
@@ -200,115 +288,134 @@ mod test {
...
+                },
+                Ordering::Greater => {
+                    if should_also_check_equality {
+                        circuit.is_geq_constant(a, *b)?
+                    } else {
+                        circuit.is_gt_constant(a, *b)?
+                    }
+                },
+                Ordering::Equal => {
+                    let b = circuit.create_constant_variable(*b)?;
```
chancharles92, line 339: Do we need this test path?

relation/src/gadgets/cmp.rs
```diff
@@ -200,115 +288,134 @@ mod test {
...
+                    }
+                },
+                Ordering::Greater => {
+                    if should_also_check_equality {
+                        circuit.enforce_geq_constant(a, *b)?
+                    } else {
+                        circuit.enforce_gt_constant(a, *b)?
+                    }
+                },
+                Ordering::Equal => {
```
chancharles92, line 394: Do we need this test path? It's basically a constant gate.

chancharles92 (commented):
Generally LGTM

relation/src/gadgets/cmp.rs
mrain: No. Added here only for completeness. Test will not enter here.

chancharles92 (approved):

Issue #88, opened 2022-07-29
[API] Refactor Merkle tree and its gadget 


alxiong:
Current API for MT and its gadgets need some refactoring, e.g. here's the code we use in ZPrice to generate a circuit of lots of Merkle Proof verification:

```rust
    // construct circuit constraining membership proof check
    let mut circuit = PlonkCircuit::new();
    // add root as a public input
    let root_var = circuit.create_public_variable(root)?;
    for (uid, proof) in merkle_proofs.iter().enumerate() {
        let leaf_var = circuit.create_variable(leaves[uid])?;
        let proof_var = AccMemberWitnessVar::new(&mut circuit, proof)?;
        let acc_elem_var = AccElemVars {
            uid: proof_var.uid,
            elem: leaf_var,
        };

        let claimed_root_var = circuit.compute_merkle_root(acc_elem_var, &proof_var.merkle_path)?;

        // enforce matching merkle root
        circuit.equal_gate(root_var, claimed_root_var)?;
    }

    // sanity check: the circuit must be satisfied.
    assert!(circuit.check_circuit_satisfiability(&[root]).is_ok());
```

- mixed style of `Struct::new()` and field filling to instantiate structs
- numerous `MerklePath`, `MerkleNode` etc are slightly confusing
- gadget APIs are not intuitive at all

alxiong:
we should define something like `trait AppendOnlyMerkleTree` (or `ForgettableMerkleTree` or whatever, name subject to change), so that the interface of our MT is clear from the outset.

alxiong:
I feel that the following is more ergonomic:

```rust
use ark_ff::PrimeField;
use jf_relation::{errors::CircuitError, BoolVar, PlonkCircuit};

use crate::{
    merkle_tree::{
        prelude::{RescueMerkleTree, RescueSparseMerkleTree},
        MerkleTreeScheme, UniversalMerkleTreeScheme,
    },
    rescue::RescueParameter,
};

/// Gadgets for rescue-based merkle tree
pub trait MerkleTreeGadget<F: PrimeField + RescueParameter> {
    /// given an element and its merkle proof in a `RescueMerkleTree`,
    /// return `BoolVar` indicating the correctness of its membership proof
    fn is_member(
        &mut self,
        elem: &<RescueMerkleTree<F> as MerkleTreeScheme>::Element,
        merkle_proof: &<RescueMerkleTree<F> as MerkleTreeScheme>::MembershipProof,
[...]

PR #120, opened 2022-09-15
Refactor MT & Generic Implementation

mrain:
## Description

Part of: #88 

- [x] Generic merkle tree traits
- [x] Generic merkle tree implementation [WIP]
    - [x] Merkle tree
    - [x] Updatable / Appendable Merkle tree
    - [x] Forgetable Merkle tree 
    - [x] Unit tests
- [ ] Merkle tree gadgets

---

write a little note why.

- [x] Wrote unit tests


primitives/src/merkle_tree/merkle_tree_impl.rs
```diff
@@ -0,0 +1,413 @@
...
+    LeafArity: Unsigned,
+    TreeArity: Unsigned,
+    F: Field,
+{
+    type ElementType = E;
+    type Hasher = H;
+    type LeafArity = LeafArity;
+    type TreeArity = TreeArity;
+    type Proof = MerkleProof<E, F>;
+    type BatchProof = MerkleNode<E, F>;
```
chancharles92, line 46: Is it better to define a struct `MerkleBatchProof<E,F>` rather than using `MerkleNode`?

primitives/src/merkle_tree/merkle_tree_impl.rs
```diff
@@ -0,0 +1,413 @@
...
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct MerkleProof<E, F: Field> {
+    /// Proof of inclusion for element at index `pos`
+    pub pos: u64,
+    /// Branch index
+    pub path: Vec<usize>,
+    /// root of proof path
```
chancharles92, line 409: nodes of proof path?

primitives/src/merkle_tree/merkle_tree_impl.rs
mrain: I haven't implement the batch proof part. This is just a placeholder. 
Naturally for a batch proof, it contains a sub-tree. (Single proof is just a tree path, we can collect it as a vector).
I'm still deciding on the right representation for a sub-tree.

primitives/src/merkle_tree/merkle_tree_traits.rs
```diff
@@ -0,0 +1,158 @@
...
+pub trait AppendableMerkleTree<F: Field>: MerkleTree<F> {
+    /// Insert a new value at the leftmost available slot
+    /// * `elem` - element to insert in the tree
+    /// * `returns` - Ok(()) if successful
+    fn push(&mut self, elem: &Self::ElementType) -> Result<(), PrimitivesError>;
+
+    /// Insert a list of new values at the leftmost available slots
+    /// * `elems` - elements to insert
+    /// * `returns` - Ok(()) if successful
+    fn emplace(
```
jbearer, line 126: Is there a reason we went with `emplace`? I think it's more idiomatic to call this `extend`. `emplace` reminds me of C++ emplacement `new`, but that's not what this is at all.

primitives/src/merkle_tree/merkle_tree_impl.rs
```diff
@@ -0,0 +1,413 @@
...
+            self.capacity,
+            elem,
+        )
+    }
+
+    fn emplace(
+        &mut self,
+        elems: impl Iterator<Item = Self::ElementType>,
+    ) -> Result<(), PrimitivesError> {
+        // TODO(Chengyu): efficient batch insert
```
jbearer, line 176: This will be very important for the wallet. We currently make use of [FilledMTBuilder](https://jellyfish.docs.espressosys.com/jf_primitives/merkle_tree/struct.FilledMTBuilder.html) (e.g. [here](https://github.com/EspressoSystems/seahorse/blob/main/src/lw_merkle_tree.rs#L250))

primitives/src/merkle_tree/merkle_tree_traits.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    /// Return the current root value
+    fn value(&self) -> F;
+
+    /// Returns the leaf value given a position
+    /// * `pos` - zero-based index of the leaf in the tree
+    /// * `returns` - Leaf value at the position. LookupResult::EmptyLeaf if the
+    ///   leaf position is empty or invalid, LookupResult::NotInMemory if the
+    ///   leaf position has been forgotten.
+    fn lookup(&self, pos: u64) -> LookupResult<(), Self::Proof>;
```
jbearer, line 99: Why return `()` and not `Self::ElementType`?

primitives/src/merkle_tree/merkle_tree_traits.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    /// Verify an element is a leaf of a Merkle tree given the proof
+    /// * `pos` - zero-based index of the leaf in the tree
+    /// * `proof` - a merkle tree proof
+    /// * `returns` - Err() if something is wrong with this proof, Ok(result)
+    ///   otherwise
+    fn verify(&self, pos: u64, proof: impl Borrow<Self::Proof>) -> Result<bool, PrimitivesError>;
```
jbearer, line 106: What does `true`/`false` indicate in the return value?

jbearer (commented):
Just commenting on the interface, haven't looked at the implementation yet.

One thing I think is missing, that we use extensively in wallet/EsQS/validator interactions, is the ability to extract a [comment](https://jellyfish.docs.espressosys.com/jf_primitives/merkle_tree/struct.MerkleCommitment.html) and [frontier](https://jellyfish.docs.espressosys.com/jf_primitives/merkle_tree/enum.MerkleFrontier.html) from a `MerkleTree`, and then later rebuild a `MerkleTree` from the commitment/frontier. The actual commitment/frontier combo can be opaque (e.g. an associated type), the important thing is that there is some commitment operation which gives you something concise which you can then use to reinstantiate a `MerkleTree`.

primitives/src/merkle_tree/merkle_tree_traits.rs
mrain: No particular reason. Will return to `extend`.

primitives/src/merkle_tree/merkle_tree_traits.rs
mrain: I'm not sure whether it's useful to return the element. But let me add this.

primitives/src/merkle_tree/merkle_tree_impl.rs
mrain: Will implement soon.

primitives/src/merkle_tree/merkle_tree_traits.rs
jbearer: Then what does `Err` indicate? I thought `Err` would mean the proof is not valid and `Ok` would mean it was. I think at least this function needs better documentation on how to interpret the result.

mrain: `Err` indicates that the proof is not well structured (E.g. not for this given MT).
Sure I'll try to document it better.

jbearer: Oh, I see. Makes sense now

mrain:
> Good work to kickstart more discussion and refactoring work. 👍
> 
> One thing I struggle with, is the unclear abstraction of our `trait MT`, should we incline towards a Set accumulator or a Vector accumulator?
> 
> * API like `append()` or `push()` is more on the "vector accumulator" side, whereas (unordered) `insert` is leaning towards "set accumulator"

IMO it's more like a vector/map accumulator. To access an element you need to specify a position to it.

> * Our Interval Merkle Tree is insertion order dependent, thus should more behave like a Vector accumulator. (quite honestly, so is our current merkle_tree impl as an append-only MT)

UptablableMerkleTree is just implemented. See latest commitment.

primitives/src/merkle_tree/merkle_tree_traits.rs
```diff
@@ -0,0 +1,158 @@
...
+    ) -> Result<(), PrimitivesError>;
+}
+
+/// Merkle tree that allows modification
+pub trait UpdatableMerkleTree<F: Field>: MerkleTree<F> {
+    /// Update the leaf value at a given position
+    /// * `pos` - zero-based index of the leaf in the tree
+    /// * `elem` - newly updated element
+    fn update(&mut self, pos: u64, elem: &Self::ElementType) -> Result<(), PrimitivesError>;
+}
```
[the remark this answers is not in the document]
mrain: Or should we call it `RandomAccessMerkleTree`? 
With 2 interfaces provided.
- `insert(pos, elem)` will error if given position is occupied.
- `overwrite(pos, elem)` will ignore whether it's occupied.

primitives/src/merkle_tree/merkle_tree_traits.rs
```diff
@@ -0,0 +1,261 @@
...
+    /// Verify an element is a leaf of a Merkle tree given the proof
+    /// * `pos` - zero-based index of the leaf in the tree
+    /// * `proof` - a merkle tree proof
+    /// * `returns` - Ok(true) if the proof is accepted, Ok(false) if not. Err()
+    ///   if the proof is not well structured, E.g. not for this merkle tree.
+    fn verify(
+        &self,
+        pos: Self::IndexType,
+        proof: impl Borrow<Self::MembershipProof>,
+    ) -> Result<bool, PrimitivesError>;
```
alxiong, line 181: Should we have `pos` as input param? Intuitively just a proof is enough and practically, the pos info is embedded in `proof`.

primitives/src/merkle_tree/merkle_tree_traits.rs
```diff
@@ -0,0 +1,158 @@
...
+/// Merkle tree hash function
+pub trait Hasher<F: Field> {
+    /// Digest a list of values
+    fn digest(data: &[F]) -> F;
+}
+
+/// Basic functionalities for a merkle tree implementation
+pub trait MerkleTree<F: Field>: Sized {
+    /// Merkle tree element type
+    type ElementType: ElementType<F>;
```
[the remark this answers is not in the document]
alxiong: again, 
- inline this and also `IndexType`
- add a `trait ToFields` or `AsFieldSlice` or sth to capture the extra interface requirement
- rename to `type Element`, similarly `type Index`

Issue #129, opened 2022-11-08
Generic Universal Merkle tree implementation


mrain:
In branch `mt-refactor` first.

PR #132, opened 2022-11-08
Universal Merkle tree implementation

mrain:
## Description

closes: #129 

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/primitives/src/merkle_tree/mod.rs
+++ b/primitives/src/merkle_tree/mod.rs
@@ -7,6 +7,7 @@
 //! Merkle Tree traits and implementations
 pub mod append_only;
 pub mod examples;
+pub mod macros;
 pub mod sparse_merkle_tree;
 
 mod internal;
--- a/primitives/src/merkle_tree/mod.rs
+++ b/primitives/src/merkle_tree/mod.rs
@@ -224,7 +224,8 @@ pub trait UniversalMerkleTreeScheme: MerkleTreeScheme {
     /// Update the leaf value at a given position
     /// * `pos` - zero-based index of the leaf in the tree
     /// * `elem` - newly updated element
-    fn update(&mut self, pos: Self::Index, elem: &Self::Element) -> Result<(), PrimitivesError>;
+    fn update(&mut self, pos: Self::Index, elem: &Self::Element)
+        -> LookupResult<Self::Element, ()>;
 
     // TODO(Chengyu): non-membership proof interfaces
 }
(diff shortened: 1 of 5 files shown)


primitives/src/merkle_tree/examples.rs
```diff
@@ -16,35 +19,37 @@ use sha3::{Digest, Sha3_256};
...
 }
 
 /// A standard merkle tree using RATE-3 rescue hash function
-pub type RescueMerkleTree<F> = MerkleTree<F, RescueHash<F>, u64, U3, F>;
+pub type RescueMerkleTree<F> = MerkleTree<F, RescueHash<u64, F>, u64, U3, F>;
 
+impl Index for BigUint {}
 /// Example instantiation of a SparseMerkleTree indexed by BigUInt
-pub type SparseMerkleTree<E, F> = MerkleTree<E, RescueHash<F>, BigUint, U3, F>;
+pub type SparseMerkleTree<E, F> = UniversalMerkleTree<E, RescueHash<BigUint, F>, BigUint, U3, F>;
```
alxiong, line 45: Why the `Index` type is `BigUint` instead of `F`?

(plus `Field` impl `Copy`)

primitives/src/merkle_tree/mod.rs
```diff
@@ -223,8 +227,23 @@ pub trait UniversalMerkleTreeScheme: MerkleTreeScheme {
...
+    /// * `height` - height of the merkle tree
+    /// * `data` - an iterator of key-value pairs. Could be a std hashmap or
+    ///   simply an array or a slice of (key, value) pairs
+    fn from_kv_set<BI, BE>(
+        height: usize,
+        data: impl IntoIterator<Item = impl Borrow<(BI, BE)>>,
+    ) -> Result<Self, PrimitivesError>
+    where
+        BI: Borrow<Self::Index>,
+        BE: Borrow<Self::Element>;
```
alxiong, line 246: if `data` can already be turned into an iterator of tuple of references, 
we don't need the iterator item to be borrowed.

i.e. 
`data: impl IntoIterator<Item = (BI, BE)>` should be fine.

mrain: It doesn't work for `&[(1,2)]`. In this case, you get references of tuples instead of tuple of references.
I was haunted by this before.

primitives/src/merkle_tree/examples.rs
mrain: Good point. I don't know before that `Field` already bound for various Ops. Let me work on it.

PR #135, opened 2022-11-14
Mt refactor

mrain:
## Description

closes: #88

---

write a little note why.

- [ ] Wrote unit tests


primitives/src/merkle_tree/examples.rs
```diff
@@ -0,0 +1,117 @@
...
+        let mut hasher = Sha3_256::new();
+        for value in data {
+            hasher.update(value);
+        }
+        Sha3Node(hasher.finalize().into())
+    }
+
+    fn digest_leaf(_pos: &I, _elem: &E) -> Sha3Node {
+        // Serialize and hash
+        todo!()
```
fkrell, line 112: is there a task for this TODO?

primitives/src/merkle_tree/macros.rs
```diff
@@ -0,0 +1,127 @@
...
+            I: Index + From<u64>,
+            Arity: Unsigned,
+            T: NodeValue,
+        {
+            type Element = E;
+            type Digest = H;
+            type Index = I;
+            type NodeValue = T;
+            type MembershipProof = MerkleProof<E, I, T>;
+            // TODO(Chengyu): implement batch membership proof
```
fkrell, line 44: Again, make sure there is task for it

fkrell (commented):
Looks good over all. Reply to comments, then I'll approve

primitives/benches/merkle_path.rs
```diff
@@ -7,47 +7,47 @@
...
-    benchmark_group.finish();
+// use jf_primitives::merkle_tree::{
+//     MerkleLeafProof, MerklePath, MerklePathNode, MerkleTree, NodePos,
+// NodeValue, };
+// use std::time::Duration;
+
+// const BENCH_NAME: &str = "merkle_path_height_20";
+
+fn twenty_hashes(_c: &mut Criterion) {
+    // let mut benchmark_group = c.benchmark_group(BENCH_NAME);
```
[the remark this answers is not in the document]
mrain: This bench is current broken after refactoring. Need a new one.

primitives/src/merkle_tree/append_only.rs
```diff
@@ -0,0 +1,292 @@
...
+            },
+            LookupResult::NotInMemory => LookupResult::NotInMemory,
+            LookupResult::EmptyLeaf(_) => LookupResult::EmptyLeaf(()),
+        }
+    }
+
+    fn remember(
+        &mut self,
+        pos: Self::Index,
+        _element: impl Borrow<Self::Element>,
```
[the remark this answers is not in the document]
mrain: Current proof already includes the element.

primitives/src/merkle_tree/examples.rs
mrain: No, this is for demo only.

primitives/src/merkle_tree/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+/// Typically, F for element type, P for membership proof type and N for
+/// non-membership proof type
+pub enum LookupResult<F, P, N> {
+    /// The value at the given index, and a proof of validity
+    Ok(F, P),
+    /// The index is valid but we do not have the leaf in memory
+    NotInMemory,
+    /// The index is outside the occupied range in the tree, and a
+    /// non-membership proof
+    EmptyLeaf(N),
```
jbearer, line 32: I feel like `Leaf` is a bit of an implementation detail. If this is going to be in a public, generic interface, maybe something higher-level like `NotInRange` or `NotFound` would be better?

primitives/src/merkle_tree/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+impl<T: Default + Eq + PartialEq + Copy + Clone + Debug> NodeValue for T {}
+
+/// Merkle tree hash function
+pub trait DigestAlgorithm<E, I, T>
+where
+    E: Element,
+    I: Index,
+    T: NodeValue,
+{
+    // Possible improvement: adding digest_element() and digest_index()
```
jbearer, line 83: 👍

primitives/src/merkle_tree/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+
+    /// Digest a leaf (an indexed element)
+    fn digest_leaf(pos: &I, elem: &E) -> T;
+}
+
+/// An trait for Merkle tree index type.
+pub trait ToTraversalPath {
+    /// Convert the given index to a vector of branch indices given tree height
+    /// and arity.
+    fn to_traverse_path(&self, height: usize, arity: usize) -> Vec<usize>;
```
jbearer, line 95: Should name this `to_traversal_path` to be consistent with the name of the trait

primitives/src/merkle_tree/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+    // Possible improvement: adding digest_element() and digest_index()
+    /// Digest a list of values
+    fn digest(data: &[T]) -> T;
+
+    /// Digest a leaf (an indexed element)
+    fn digest_leaf(pos: &I, elem: &E) -> T;
+}
+
+/// An trait for Merkle tree index type.
+pub trait ToTraversalPath {
```
jbearer, line 92: I wonder if this trait should be generic on arity, instead of making it an input of `to_traverse_path`. Fixing arity in the type system would allow optimizations like a specialized type where the path is determined by bit shifting, for 2-arity.

primitives/src/merkle_tree/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+        }
+    }
+}
+
+/// An element of a Merkle tree.
+pub trait Element: Clone + Eq + PartialEq {}
+impl<T: Clone + Eq + PartialEq> Element for T {}
+
+/// An index type of a leaf in a Merkle tree.
+pub trait Index: Debug + Eq + PartialEq + Ord + PartialOrd + Clone + ToTraversalPath {}
```
jbearer, line 69: `ToTraversalPath` seems like it would be used by the implementor of a Merkle tree, not the user. If true, it might be better to require that on the impl (e.g. `impl<I: ToTraversalPath> AppendableMerkleTreeScheme for MerkleTree<I>`) since the `ToTraversalPath` requirement might be stricter than a particular implementation really needs. (As a concrete example, `ToTraversalPath`, as currently specified, requires a conversion to a path for any arity, but any particular Merkle tree implementation probably has a fixed arity, see my comment below)

primitives/src/merkle_tree/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+            ret.push((&pos % (arity as u64)).to_usize().unwrap());
+            pos /= arity as u64;
+        }
+        ret
+    }
+}
+
+/// A merkle commitment consists a root hash value, a tree height and number of
+/// leaves
+#[derive(Eq, PartialEq, Clone, Copy, Serialize, Deserialize)]
```
jbearer, line 112: I would request a few extra traits: `CanonicalSerialize`, `CanoniclDeserialize`, `Hash` and `Ord`, plus replace `Serialize, Deserialize` with `#[tagged_blob("MERKLECOMM")]` so we also get `Display`/`FromStr`

primitives/src/merkle_tree/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+pub struct MerkleCommitment<T: NodeValue> {
+    /// Root of a tree
+    pub root_value: T,
+    /// Height of a tree
+    pub height: usize,
+    /// Number of leaves in the tree
+    pub num_leaves: u64,
+}
```
jbearer, line 120: This seems like it would be very specific to an implementation. If the intent is for the trait below to be implementable for a variety of accumulators, I think this should be replaced by a `trait MerkleCommitment<T: NodeValue>` and an associated type on `MerkleTreeScheme` like `type Commitment: MerkleCommitment`. Interface-wise, I don't think the exact structure of this type is important for users, as long as it satisfies the following properties:
* concise (e.g. `Copy`)
* can be used to identify a tree (`Eq`, `Ord`, `Hash`)
* exposes the size of the tree (`num_leaves`, but we should expose it via a trait function with a more abstract name like `size` or something)

primitives/src/merkle_tree/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+
+/// Basic functionalities for a merkle tree implementation. Abstracted as an
+/// accumulator for fixed-length array. Supports generate membership proof at a
+/// given position and verify a membership proof.
+pub trait MerkleTreeScheme: Sized {
```
jbearer, line 125: I wonder if we should call this (and the whole module, maybe) something like `AccumulatorScheme`, to be a bit more generic

primitives/src/merkle_tree/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+    fn to_traverse_path(&self, height: usize, arity: usize) -> Vec<usize> {
+        let mut pos: BigUint = <F as Into<BigUint>>::into(self.clone());
+        let mut ret = vec![];
+        for _i in 0..height {
+            ret.push((&pos % (arity as u64)).to_usize().unwrap());
+            pos /= arity as u64;
+        }
+        ret
+    }
+}
```
jbearer, line 108: This blanket impl makes it impossible to implement `ToTraversalPath` for types other than integer-like types. For example, if it weren't for this, we could implement `ToTraversalPath` for `Vec<usize>` (trivially) or arbitrary user-defined types by serializing. I think it would be slightly better to manually implement this for each specific number type (e.g. `u8` through `u128`, `BigUint`, `U256`, field elements). It is probably not too bad to just turn this impl into a macro that gets applied for each numeric type.

primitives/src/merkle_tree/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+
+/// Basic functionalities for a merkle tree implementation. Abstracted as an
+/// accumulator for fixed-length array. Supports generate membership proof at a
+/// given position and verify a membership proof.
+pub trait MerkleTreeScheme: Sized {
+    /// Merkle tree element type
+    type Element: Element;
+    /// Hash algorithm used in merkle tree
+    type Digest: DigestAlgorithm<Self::Element, Self::Index, Self::NodeValue>;
```
jbearer, line 129: Why is this an associated type if it's not used in the interface of the trait? Shouldn't it just be up to the implementation to pick a digest, without it being exposed in the trait at all?

primitives/src/merkle_tree/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+
+    /// Return the height of this merkle tree
+    fn height(&self) -> usize;
+    /// Return the maximum allowed number leaves
+    fn capacity(&self) -> BigUint;
+    /// Return the current number of leaves
+    fn num_leaves(&self) -> u64;
+
+    /// Return the current root value
+    fn root(&self) -> Self::NodeValue;
```
jbearer, line 156: Not sure this really belongs in the trait. For an abstract interface, `commitment` serves the purpose of giving the tree a concise identifier, and I'm not sure what other functionality the user would expect from `root` -- especially with the return value being an opaque type. If we were to get rid of this, and implement the change I suggested to make `Commitment` an associated type, we could remove the `NodeValue` associated type, since it would no longer appear in the interface.

primitives/src/merkle_tree/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+
+    /// Returns the leaf value given a position
+    /// * `pos` - zero-based index of the leaf in the tree
+    /// * `returns` - Leaf value at the position along with a proof.
+    ///   LookupResult::EmptyLeaf if the leaf position is empty or invalid,
+    ///   LookupResult::NotInMemory if the leaf position has been forgotten.
+    fn lookup(
+        &self,
+        pos: impl Borrow<Self::Index>,
+    ) -> LookupResult<Self::Element, Self::MembershipProof, ()>;
```
jbearer, line 169: We could have `Self::NonMembershipProof` be an associated type that defaults to `()`. That way if we were to implement this trait for some accumulator that does support non-membership proofs, this function would automatically become slightly more useful when using that specific implementation.

primitives/src/merkle_tree/mod.rs
jbearer: This is especially important for key-value type accumulators

primitives/src/merkle_tree/mod.rs
mrain: Right now I integrate it to `digest_leaf`. Users can self-define their behaviors on how to handle index and element.

primitives/src/merkle_tree/mod.rs
mrain: There's a `universal_lookup` function in `UniversalMerkleTreeScheme`. Are you suggesting to merge this two?
@alxiong WDYT

primitives/src/merkle_tree/mod.rs
mrain: Good suggestion. Will do.

primitives/src/merkle_tree/mod.rs
mrain: Had the same trouble before. Let me change it to a macro first.

primitives/src/merkle_tree/mod.rs
jbearer: I see. I guess there probably is some advantage to having `universal_lookup` be separate. I could imagine an implementation that supports non-membership proofs but creating them incurs some extra cost. Then `lookup` shouldn't pay that penalty if the caller doesn't care about the non-membership proof. If they want to pay the cost then they can call `universal_lookup`. So I guess keeping it as is is probably for the best

chancharles92:
@mrain Let's change this PR from a draft to a ready-for-review PR.

primitives/src/merkle_tree/append_only.rs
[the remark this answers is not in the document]
fkrell: Sure, but we need to make sure the API isn't misused.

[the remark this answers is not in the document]
mrain: Additional check is added.

primitives/src/merkle_tree/mod.rs
mrain: Now I have a trait named `MerkleCommitment` and a struct implements it called `MerkleTreeCommitment`.
It still may bring confusions to users. Are there any naming suggestion?

primitives/src/merkle_tree/mod.rs
mrain: Good point. If we do this, then everything would be in a different flavor. Probably we should also change `AppendableMerkleTreeScheme` to `VectorAccumulatorScheme` and `UniversalMerkleTreeScheme` to `MapAccumulatorScheme`? 

cc @alxiong @chancharles92 @tessico @fkrell @tessico

chancharles92: 1. If we want to make it an `AccumulatorScheme` trait, then we should remove `height()` API, remove ARITY constant, and rename some APIs as well (e.g. root(), num_leaves(), etc). Also, we should move the `AccumulatorScheme` trait to another file I think.
2. I still prefer to keep `MerkleTreeScheme`, `AppendableMerkleTreeScheme`, and `UniversalMerkleTreeScheme` traits, as their possible instantiations are rather different from other flavors of instantiations (e.g. bilinear-map or RSA accumulators). If we want `AccumulatorScheme` trait, what we can do is add `trait MerkleTreeScheme : AccumulatorScheme` after defining the `AccumulatorScheme` trait.

primitives/src/merkle_tree/mod.rs
```diff
@@ -80,53 +112,84 @@ where
...
     fn digest_leaf(pos: &I, elem: &E) -> T;
 }
 
 /// An trait for Merkle tree index type.
-pub trait ToTraversalPath {
+pub trait ToTraversalPath<Arity: Unsigned> {
     /// Convert the given index to a vector of branch indices given tree height
     /// and arity.
-    fn to_traverse_path(&self, height: usize, arity: usize) -> Vec<usize>;
+    fn to_traversal_path(&self, height: usize) -> Vec<usize>;
```
jbearer, line 126: We could enforce the arity constraint at compile time by changing `Vec<usize>` to `GenericArray<usize, Arity>`

primitives/src/merkle_tree/macros.rs
```diff
@@ -120,7 +108,140 @@ macro_rules! impl_merkle_tree_scheme {
...
+                    ))
+                }
+            }
+        }
+    };
+}
+
+/// Macros for implementing ToTreversalPath for primitive types
+#[macro_export]
+macro_rules! impl_to_treversal_path_primitives {
```
jbearer, line 216: `treversal` -> `traversal`

primitives/src/merkle_tree/macros.rs
```diff
@@ -120,7 +108,140 @@ macro_rules! impl_merkle_tree_scheme {
...
+                }
+                ret
+            }
+        }
+    };
+}
+
+/// Macros for implementing ToTreversalPath for BigUint types
+#[macro_export]
+macro_rules! impl_to_treversal_path_biguint {
```
jbearer, line 234: `treversal` -> `traversal`

primitives/src/merkle_tree/mod.rs
mrain: Here the length is height + 1. Is it too complicated to make height as a trait as well?

jbearer: Oh right, the length of the output is not based on `Arity`. I was confused. I think making `height` a type parameter would be a good thing but might be more trouble than it's worth...currently we assume that a tree will be instantiated with a certain height at runtime and changing that to instantiate the type parameter at compile time might be troublesome. Never mind this comment, then

jbearer (approved):
This looks good to me from a systems/interface perspective

primitives/src/merkle_tree/mod.rs
```diff
@@ -0,0 +1,354 @@
...
+//     }
+
+//     fn height(&self) -> usize {
+//         self.height
+//     }
+
+//     fn size(&self) -> u64 {
+//         self.num_leaves
+//     }
+// }
```
alxiong, line 185: these should be removed?

primitives/src/merkle_tree/mod.rs
```diff
@@ -0,0 +1,354 @@
...
+    /// Index type for this merkle tree
+    type Index: Index;
+    /// Internal and root node value
+    type NodeValue: NodeValue;
+    /// Merkle proof
+    type MembershipProof: Clone;
+    /// Batch proof
+    type BatchMembershipProof: Clone;
+    /// Merkle tree commitment
+    type Commitment: Clone;
```
alxiong, line 202:
```suggestion
    /// Merkle tree commitment
    type Commitment: MerkleCommitment;
```

primitives/src/merkle_tree/internal.rs
```diff
@@ -0,0 +1,672 @@
...
+
+#[allow(clippy::type_complexity)]
+pub(crate) fn build_tree_internal<E, H, I, Arity, T>(
+    height: usize,
+    elems: impl IntoIterator<Item = impl Borrow<E>>,
+) -> Result<(Box<MerkleNode<E, I, T>>, u64), PrimitivesError>
+where
+    E: Element,
+    H: DigestAlgorithm<E, I, T>,
+    I: Index + From<u64>,
```
alxiong, line 146: There's a lot of `Index + From<u64>`, maybe we should add to `trait Index: ... + From<u64>` instead?

alxiong (commented):
Really awesome job on this big PR! way to go! 💯

primitives/src/merkle_tree/macros.rs
```diff
@@ -0,0 +1,248 @@
...
+        }
+    };
+}
+
+/// Macros for implementing ToTreversalPath for BigUint types
+#[macro_export]
+macro_rules! impl_to_traversal_path_biguint {
+    ($t: ty) => {
+        impl<Arity: Unsigned> ToTraversalPath<Arity> for $t {
+            fn to_traversal_path(&self, height: usize) -> Vec<usize> {
```
alxiong, line 237: curious, do you want to merge these two into one macro to save code duplication?

impl_to_traversal_path(type, is_primitive) ?

flake.nix
```diff
@@ -58,15 +59,15 @@
...
-              };
+              # license-header-c-style = {
+              #   enable = true;
+              #   description =
+              #     "Ensure files with c-style comments have license header";
+              #   entry = ''
+              #     insert_license --license-filepath .license-header  --comment-style "//"'';
+              #   types_or = [ "rust" ];
+              #   pass_filenames = true;
+              # };
```
alxiong, line 70: apology this is my mess, pls just remove these lines.

primitives/src/merkle_tree/internal.rs
jbearer: @alxiong @mrain this prevents using a complex type as `Index` (e.g. for key-value Merkle tree) are we ok with that?

mrain: I agree with @jbearer . It's not changed.

Issue #125, opened 2022-10-19
VRF API refactor


alxiong:
There're a few things to follow-up on VRF refactoring:

- possible to remove `CSID` domain separator? (how necessary)
- should add the `output` to `vrf_verify(pk, input, output, proof)` API or follow this [IRTF draft](https://datatracker.ietf.org/doc/pdf/draft-irtf-cfrg-vrf-15) and leaving it out?
- should the `<H: Hasher>` be generic over `VrfScheme` or `VrfScheme::evaluate()`?

cc @chancharles92 @mrain @jbearer @DieracDelta @nyospe @philippecamacho

DieracDelta:
A couple of other things:
- we may want to require some of the associated types of VRF implement: `Sync + Send + PartialEq + Clone + Serialize + for<'a> Deserialize<'a>`
- we need to expose a bytes representation of the keys so we aren't doing a cursed serialization [here](https://github.com/EspressoSystems/HotShot/blob/main/src/traits/election/vrf.rs#L204)

nyospe:
Neither `VrfSchema` nor `VrfSchema::evaluate` should be generic over `H: Hasher`.

The use of a hasher is specific to some (many) implementations, but not an inherent property of a `VrfSchema`, and only the structs that implement `VrfSchema` with a hasher should be generic. `VrfSchema` should not be generic. `BLSVrfSchema` should be.

PR #137, opened 2022-11-17
VRF API refactor

tessico:
## Description

closes: #125 

---

write a little note why.

- [x] Wrote unit tests


chancharles92:
@tessico Let me know when the PR is ready, and you can add me and @alxiong as reviewers.

primitives/src/vrf/blsvrf.rs
```diff
@@ -2,26 +2,51 @@
...
+impl BLSVRFScheme {
+    /// Creates a new BLS VRF instance with the given ciphersuite.
+    pub fn new(cs_id: BLSVRFCipherSuite) -> Self {
+        match cs_id {
+            BLSVRFCipherSuite::VRF_BLS_12_381_SHA256 => Self {
+                hasher: Box::new(Sha256::new()),
+            },
+            BLSVRFCipherSuite::VRF_BLS_12_381_SHA512 => Self {
+                hasher: Box::new(Sha512::new()),
+            },
```
alxiong, line 44: I'm slightly concerned about the tradeoff of using trait objects here (i.e. `Box<dyn>`)

I understood it as a way around not having to pass in as a generic param on the trait`VrfScheme<H: Hasher>`, but dynamic dispatch come at a runtime cost ([some reports](https://medium.com/digitalfrontiers/rust-dynamic-dispatching-deep-dive-236a5896e49b) say about 20% slowdown), and our VRF is quite performance sensitive.

imo, the following idea might be better: 
```rust
// trait is not generic, as some VRF instantiation uses PRF instead of Hash to get the output e.g.
trait Vrf {
    fn prove() {}
}

// the concrete impl will be generic
struct BlsVrf<H: Digest>(PhantomData<H>);

impl<H: Digest> Vrf for BlsVrf<H> {
    fn prove() {
        let hasher = H::new();
    }
}
```

primitives/src/vrf/mod.rs
```diff
@@ -54,47 +54,17 @@ pub trait Vrf<VrfHasher> {
...
     ) -> Result<Self::Output, PrimitivesError>;
 
     /// Verifies a VRF proof.
     fn verify(
+        &self,
         pp: &Self::PublicParameter,
         proof: &Self::Proof,
         public_key: &Self::PublicKey,
         input: &Self::Input,
     ) -> Result<bool, PrimitivesError>;
```
alxiong, line 69: @philippecamacho suggested previously that we should add `output: Self::Output` to the input param to `verify()` so that the users won't accidentally use the wrong (proof, output) combo. 

as you could potentially use a proof, then `H_1(proof, "dom sep: app1")` and `H_2(proof, "dom sep: app2)` to get different VRF output. 

I believe we knew it's not the most standard-compliant API but it's less error-prone. 

cc @fkrell @chancharles92

alxiong (commented):
Good job 👍 ! @tessico , only a few comments

- maybe you can just start a branch from this repo instead of from your fork, (I believe you have write access, right?) so that it's slightly easier for us to switch to your branch without adding another git remote. (thanks!)

primitives/src/vrf/blsvrf.rs
tessico: Your proposed solution was actually my first attempt at this. The dynamic trait is indeed added to avoid the generic `H`. Also it allows for a fine control of which concrete instantiations are actually supported via the `BLSVRFCipherSuite` enum.
The concern about the performance is a good point though. I will run some benchmarks on both approaches.

primitives/src/vrf/mod.rs
tessico: Do you mean that the `verify` function will take `proof, output, ...` and internally check `evaluate(proof) =? output` as an extra safety check?

I'm against that. It would mean that a user (potentially) runs the `evaluate` sub-routing twice - first to actually get the VRF output and second to verify that a proof is valid.

Unless the **only** scenario when someone calls `verify(proof, output, ...)` is when they are provided **both** `proof` and `output` by a prover - then the "safe" API would make sense.
But I can imagine that a more likely scenario is when the prover produces a proof, runs evaluate for themselves, and only posts/shares a `proof` without the `output` - leaving it for the verifier to run `evaluate` for themselves (once!) and then finally `verify(proof, ...)`.

primitives/src/vrf/blsvrf.rs
```diff
@@ -98,14 +127,33 @@ where
...
+            .unwrap());
+    }
 
     #[test]
     fn test_bls_vrf() {
-        let message = [0u8; 32];
-        let message_bad = [1u8; 32];
-        sign_and_verify::<BLSVRFScheme, Sha256>(&message);
-        failed_verification::<BLSVRFScheme, Sha256>(&message, &message_bad);
+        let message = vec![0u8; 32];
```
chancharles92, line 151: Should we also test the correctness of the hash output somewhere?

primitives/src/vrf/mod.rs
chancharles92: I think we can have 2 verify APIs, the new set of APIs is as follows:
1. prove(pp, sec_key, input) -> proof
2. proof_to_hash(pp, proof) -> output
3. evaluate(pp, sec_key, input) -> output (i.e. proof_to_hash(pp, prove(pp, sec_key, input)))
4. verify(pp, pub_key, proof, input) -> (bit, output) (this API also outputs hash as the hash is much cheaper than sig verification and the user doesn't need to call `proof_to_hash` again to obtain the hash)
5. verify_hash_consistency(pp, proof, output) -> bit

The APIs 1..4 are exactly identical to the standard (https://datatracker.ietf.org/doc/draft-irtf-cfrg-vrf/), the API 5 is used to address the issue mentioned above where the prover and verifier may use inconsistent hashing scheme so that their outputs are inconsistent.

chancharles92: Actually now I prefer to remove API 5 as the work can be done in API 4: so we just modify the API 4 as follows:
`verify(pp, pub_key, proof, input, output') -> (bit, output)`
Since API 4 anyway computes `output` itself, it can add an additional equality check that `output' ?= output`, which is almost for free.

primitives/src/vrf/blsvrf.rs
tessico: Good point, will add

primitives/src/vrf/mod.rs
tessico: I agree that we should return `(bit, output)` from the `verify()` function. I missed that, sorry.

I would argue, however, that there might be cases where `output` is not known beforehand and cannot be used as input to the `verify` routine (unless the user explicitly calls `proof_to_hash` first). Please correct me if that's never the case.

If it's so, we would have these alternatives:
1. We give `output'` an `Option<Self::Output>` type, i.e. `verify(pp, pub_key, proof, input, option(output')) -> (bit, output)`. If available, we perform the check, and otherwise just return the result.
2. We delegate correctness checking to the user. This option is fully [IRTF draft](https://datatracker.ietf.org/doc/draft-irtf-cfrg-vrf/) compatible.

I've got an example of alternative 2 [here](https://github.com/EspressoSystems/jellyfish/tree/vrf-refactor-v2). The nice thing is that Rust can help with ensuring that output is used, e.g. by the `#[must_use]` attribute, as shown in the v2 branch. This is my preferred approach. There are numerous ways that users can screw up the protocol, I'm not sure whether we should try to prevent against such fundamental mistakes as using a wrong hash function.

Also as @alxiong suggested I've just created a new branch instead of a branch in a fork. For the sake of keeping these discussions I guess we can keep this PR as is now? There's a branch with the same name and I intend to keep them in-sync if people wanna check it out directly.

chancharles92: `For the sake of keeping these discussions I guess we can keep this PR as is now? `
Yeah, ofc!

Regarding the new changes in the new branch:
1. Let's still have `evaluate` API but provide a default implementation by composing `prove` and `proof_to_hash`
2. If we use alternative 2 (over alternative 1), I'm not fully sure how the user would be able to check that it's in sync with the prover. E.g., both the prover and the verifier were using hash v1.0. before; later prover updates to hash v2.0.while the verifier still uses hash v1.0. How would the verifier be able to know that he's now using the wrong hash functions now?

cc @alxiong @mrain for more opinions.

tessico: > 2. If we use alternative 2 (over alternative 1), I'm not fully sure how the user would be able to check that it's in sync with the prover.

If the prover shares only `proof`, and doesn't provide the output (e.g. only proof is posted on-chain, or some other way), then the verifier has no way to ensure sync anyway.

If the prover shares `(proof, output')`, then the verifier can call `verify(proof, ...) = (bit, output)` and for themselves check if `output' == output`.

tessico: > Let's still have `evaluate` API but provide a default implementation by composing `prove` and `proof_to_hash`

So currently I just renamed `evaluate` -> `proof_to_hash`.

You saying that we add an extra `evaluate(secret_key, input)` function corresponding to `VRF_hash` in the draft as a convenience, right? Would you add that for the `Vrf` trait or specifically for the BLS VRF as a struct method?

chancharles92: `You saying that we add an extra evaluate(secret_key, input) function corresponding to VRF_hash in the draft as a convenience, right? Would you add that for the Vrf trait or specifically for the BLS VRF as a struct method?`
Yes, and add it to Vrf trait

primitives/src/vrf/blsvrf.rs
tessico: @alxiong Indeed your proposed solution is faster. I've ran some benchmarks [in this branch](https://github.com/EspressoSystems/jellyfish/blob/vrf-bench/bench.md#vrf-benchmarks) and found about 13% speed difference for hashing (no difference for the other operations, as expected).

Taking things into perspective though, we should evaluate whether a 13% speedup on the hashing (which itself is about 1/1000th of the verification time) is worth the convenience that comes with a dynamic hasher and the control that is implied by the use of enums to instantiate only the allowed hashers.

primitives/src/vrf/mod.rs
tessico: @chancharles92 Done: https://github.com/EspressoSystems/jellyfish/commit/8513626094e72624ebab274dd416de57d4ad5756.

primitives/src/vrf/blsvrf.rs
alxiong: > a 13% speedup on the hashing (which itself is about 1/1000th of the verification time) is worth the convenience

good point, I'm on you side now.

primitives/src/vrf/mod.rs
alxiong: I'm a little confused, is that commit on this PR or on another branch @tessico ?

tessico:
I've updated the branches, and also references in the discussion threads here, such that that all links point to commits in this branch.
This branch `tessico/vrf-refactor` is in sync with `EspressoSys/vrf-refactor` branch for convenience of checking it out locally @alxiong.

primitives/src/vrf/blsvrf.rs
tessico: Done in https://github.com/EspressoSystems/jellyfish/pull/137/commits/b9f3c50802e78f674cef8de80722e871c23b4cff.

alxiong (approved):
LGTM! (feel free to add trait bounds of associate types on `trait Vrf` and `trait SignatureScheme` in a separate PR)

Issue #131, opened 2022-11-08
Gadget implementation for new Merkle tree


mrain:
- [ ] Gadgets for append-only Merkle tree instantiated by rate-3 rescue hash function.
- [ ] Gadgets for non-membership proof for sparse merkle tree.

Possible useful references: 
[Old merkle tree gadgets](https://github.com/EspressoSystems/jellyfish/blob/main/primitives/src/circuit/merkle_tree.rs),
[new Merkle tree refactoring](https://github.com/EspressoSystems/jellyfish/tree/mt-refactor)

mrain:
Closed by #144

PR #138, opened 2022-11-17
Implement gadgets for append-only MT with rate-3 Rescue

tessico:
## Description

ref: #131 

I've expanded some tests, only to realize that a more cohesive testing approach would help. I might create a separate issue for that later.

---

write a little note why.

- [x] Targeted PR against correct branch (mt-refactor)
- [x] Wrote unit tests

### DIFF
--- a/primitives/src/circuit/merkle_tree.rs
+++ b/primitives/src/circuit/merkle_tree.rs
@@ -25,6 +25,7 @@ type NodeVal<F> = <RescueMerkleTree<F> as MerkleTreeScheme>::NodeValue;
 type MerkleProof<F> = <RescueMerkleTree<F> as MerkleTreeScheme>::MembershipProof;
 type Index<F> = <RescueMerkleTree<F> as MerkleTreeScheme>::Index;
 type MerklePath<F> = Vec<MerkleNode<F, Index<F>, NodeVal<F>>>;
+use typenum::U3;
 
 #[derive(PartialEq, Eq, Clone, Debug)]
 struct MerkleNodeBooleanEncoding<F: PrimeField + RescueParameter> {
@@ -65,31 +66,20 @@ impl<F: PrimeField + RescueParameter> MerklePathBooleanEncoding<F> {
 
 impl<F: PrimeField + RescueParameter> From<&MerkleProof<F>> for MerklePathBooleanEncoding<F> {
     fn from(proof: &MerkleProof<F>) -> Self {
-        let path = proof.pos.to_traverse_path(proof.tree_height() - 1, 3);
+        let path =
+            <u64 as ToTraversalPath<U3>>::to_traversal_path(&proof.pos, proof.tree_height() - 1);
 
         let nodes: Vec<MerkleNodeBooleanEncoding<F>> = path
             .iter()
             .zip(proof.proof.iter().skip(1))
             .filter_map(|(branch, node)| match node {
-                MerkleNode::Branch { value: _, children } => {
-                    // println!("children: {:?}", children);
-                    Some(MerkleNodeBooleanEncoding::new(
-                        // <RescueMerkleTree<F> as MerkleTreeScheme>::Digest::digest_leaf(
-                        //     &0,
-                        //     &children[0].value(),
-                        // ),
-                        children[0].value(),
-                        // <RescueMerkleTree<F> as MerkleTreeScheme>::Digest::digest_leaf(
-                        //     &1,
-                        //     &children[1].value(),
-                        // ),
-                        children[1].value(),
-                        branch == &0,
-                        branch == &2,
-                    ))
-                },
+                MerkleNode::Branch { value: _, children } => Some(MerkleNodeBooleanEncoding::new(
+                    children[0].value(),
+                    children[1].value(),
+                    branch == &0,
+                    branch == &2,
+                )),
                 _ => None,
-                // _ => panic!("unexpected node type"),
             })
             .collect();
         // unimplemented!();
@@ -306,7 +296,8 @@ where
 
 #[cfg(test)]
 mod test {
-    use crate::merkle_tree::DigestAlgorithm;
+    use crate::merkle_tree::examples::RescueHash;
+    use crate::merkle_tree::{DigestAlgorithm, MerkleCommitment};
     use crate::{
         circuit::merkle_tree::{
             AccElemVars, MerkleNodeBooleanEncoding, MerklePathBooleanEncoding, MerkleProof,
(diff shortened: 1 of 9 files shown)


primitives/src/merkle_tree/examples.rs
```diff
@@ -31,7 +31,7 @@ impl<F: RescueParameter> DigestAlgorithm<F, u64, F> for RescueHash<F> {
     }
 
     fn digest_leaf(pos: &u64, elem: &F) -> F {
-        let data = [F::from(*pos), *elem, F::zero()];
```
tessico, line 34: This had to be changed to match the spec. 
Currently the MT-circuit is very specific for this MT data structure, we should either move it out of examples and re-export from crate root(?) as canonical MT or make it clear in the circuit implementation that it refers to append-only MT with 3-rate Rescue hasher.

chancharles92:
@tessico Let me know when the PR is ready, and you can add me and @mrain  as reviewers.

primitives/src/circuit/merkle_tree.rs
```diff
@@ -42,35 +48,37 @@ impl<F: PrimeField> MerkleNodeBooleanEncoding<F> {
...
 }
 
-struct MerklePathBooleanEncoding<F: PrimeField> {
+#[derive(PartialEq, Eq, Clone, Debug)]
+struct MerklePathBooleanEncoding<F: PrimeField + RescueParameter> {
     pub nodes: Vec<MerkleNodeBooleanEncoding<F>>,
 }
 
-impl<F: PrimeField> MerklePathBooleanEncoding<F> {
+impl<F: PrimeField + RescueParameter> MerklePathBooleanEncoding<F> {
```
mrain, line 56: Does it make sense to call it a `MerkleProofBooleanEncoding`?

mrain (changes requested):
There are some naming suggestions.

Good work! 👍

primitives/src/circuit/merkle_tree.rs
```diff
@@ -10,26 +10,32 @@
...
-#[derive(Clone)]
-struct MerkleNodeBooleanEncoding<F: PrimeField> {
-    sibling1: NodeValue<F>,
-    sibling2: NodeValue<F>,
+type NodeVal<F> = <RescueMerkleTree<F> as MerkleTreeScheme>::NodeValue;
+type MerkleProof<F> = <RescueMerkleTree<F> as MerkleTreeScheme>::MembershipProof;
+use typenum::U3;
+
+#[derive(PartialEq, Eq, Clone, Debug)]
+struct MerkleNodeBooleanEncoding<F: PrimeField + RescueParameter> {
```
chancharles92, line 28: Any reason that we need to add RescueParameter bound here?

primitives/src/circuit/merkle_tree.rs
```diff
@@ -10,26 +10,32 @@
...
-use ark_std::{vec, vec::Vec};
+use ark_std::vec::Vec;
 use jf_relation::{errors::CircuitError, BoolVar, Circuit, PlonkCircuit, Variable};
 
-#[derive(Clone)]
-struct MerkleNodeBooleanEncoding<F: PrimeField> {
-    sibling1: NodeValue<F>,
-    sibling2: NodeValue<F>,
+type NodeVal<F> = <RescueMerkleTree<F> as MerkleTreeScheme>::NodeValue;
+type MerkleProof<F> = <RescueMerkleTree<F> as MerkleTreeScheme>::MembershipProof;
```
chancharles92, line 24: Is it better to call it `MembershipProof<F>`? As we might support non-membership proofs in the future for sparse Merkle tree.

primitives/src/circuit/merkle_tree.rs
```diff
@@ -10,26 +10,32 @@
...
 use ark_ff::PrimeField;
-use ark_std::{vec, vec::Vec};
+use ark_std::vec::Vec;
 use jf_relation::{errors::CircuitError, BoolVar, Circuit, PlonkCircuit, Variable};
 
-#[derive(Clone)]
-struct MerkleNodeBooleanEncoding<F: PrimeField> {
-    sibling1: NodeValue<F>,
-    sibling2: NodeValue<F>,
+type NodeVal<F> = <RescueMerkleTree<F> as MerkleTreeScheme>::NodeValue;
```
chancharles92, line 23: Is it possible to make the circuit instantiation more modular so that it doesn't bind to RescueMerkleTree? E.g., suppose we want to support other hashes (e.g. Poseidon or sth else) in the future, it should be easy to do without changing much of the code here.

chancharles92 (commented):
Had the first round of review, but didn't check the tests yet.

primitives/src/circuit/merkle_tree.rs
tessico: Sure, good call.

primitives/src/circuit/merkle_tree.rs
tessico: As I mentioned in the comment above, the current circuit is very specific for the rate-3 append-only MT instantiated with the Rescue hash, as specified in the #131 issue:
"Gadgets for append-only Merkle tree instantiated by rate-3 rescue hash function."
I'm using a concrete type from the `examples` module, whose trait bounds require `RescueParameters` on `F`.

As also hinted at in the above comment, I'd be in favour of making it very explicit that the circuit is for a specific type only. We could even move it somewhere else, maybe a separate module where we define both `pub type RescueMerkleTree<F> = MerkleTree<F, RescueHash<F>, u64, U3, F>;` and circuit impl.

primitives/src/circuit/merkle_tree.rs
tessico: Sure, but maybe we could make this a separate task?
Previously the circuit was also [rescue specific](https://github.com/EspressoSystems/jellyfish/blob/main/primitives/src/circuit/merkle_tree.rs#L198) and I think it requires a bit more refactoring to make it more modular. Namely, the `MerkleTreeGadget` would need to be polymorphic over some kind of `HashGadget` (of which `RescueGadget` would be one instantiation). Unless you see an easier way?

primitives/src/merkle_tree/examples.rs
mrain: I just added a `prelude` module for mt_refactor. It now includes a canonical instantiation of `RescueMerkleTree`.

primitives/src/circuit/merkle_tree.rs
chancharles92: Yeah sounds good

primitives/src/circuit/merkle_tree.rs
tessico: @chancharles92 Also for `MerkleProofVar` -> `MembershipProofVar` and the like, right?

primitives/src/circuit/merkle_tree.rs
tessico: https://github.com/EspressoSystems/jellyfish/issues/142

primitives/src/merkle_tree/examples.rs
tessico: Great, I'll rebase later.

primitives/src/circuit/merkle_tree.rs
tessico: Done in https://github.com/EspressoSystems/jellyfish/pull/138/commits/4dcbd0b8a969fb3de588690c70408165ce435c50 and https://github.com/EspressoSystems/jellyfish/pull/138/commits/b61fa4e8213813f0876159bc075764a02ece1440.

primitives/src/merkle_tree/examples.rs
tessico: Done in https://github.com/EspressoSystems/jellyfish/pull/138/commits/b61fa4e8213813f0876159bc075764a02ece1440.

mrain:
Closed because `mt-refactor` is merged into `main` now.
We should have a new PR targeted at `main` branch.

Issue #15, opened 2022-01-27
Ark-sponge compatibilty


alxiong:
Currently we have rescue-based sponge construction whereas `ark-sponge` repo supports Poseidon. But it may be beneficial to be compliant to [their trait](https://github.com/arkworks-rs/sponge/blob/master/src/lib.rs#L108).

zhenfeizhang:
working on #46, might as well get this done :-)

zhenfeizhang:
while working on this, i realized that their interfaces is really not compatible with ours. They are designed to manipulate on bits as in the case of poseidon. we can fit there but we will be wasting a few (de)serializations.

I think we need to define a proper API ourselves, along the line of https://docs.rs/sha2/latest/sha2/trait.Digest.html

zhenfeizhang:
> while working on this, i realized that their interfaces is really not compatible with ours. They are designed to manipulate on bits as in the case of poseidon. we can fit there but we will be wasting a few (de)serializations.
> 
> I think we need to define a proper API ourselves, along the line of https://docs.rs/sha2/latest/sha2/trait.Digest.html

Take it back. We can support those APIs

PR #50, opened 2022-04-04
sync hash api with ark-sponge

zhenfeizhang:
## Description

closes: #15 

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [x] Targeted PR against correct branch (main)
- [x] Linked to GitHub issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [x] Wrote unit tests
- [x] Updated relevant documentation in the code
- [ ] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md`
- [x] Re-reviewed `Files changed` in the GitHub PR explorer

PR #146, opened 2022-11-24
Sponge API refactor

tessico:
## Description

This is a WIP. 

The refactor defines new explicit types for CRHF and PRF which are sponges that internally use the Rescue `Permutation`, and as such is more aligned with the CAPE spec.

By following the sponge API we are able to re-use code shared between the above two functionalities.

It should then be then easier to make the circuits more generic by also defining and implementing a trait like [CryptographicSpongeVar](https://github.com/arkworks-rs/sponge/blob/master/src/constraints/mod.rs#L101), though we'd probably need to define our own due to a dependency on a R1CS `ConstraintSystem`.

closes: https://github.com/EspressoSystems/jellyfish/issues/15

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/primitives/src/prf.rs
+++ b/primitives/src/prf.rs
@@ -7,9 +7,11 @@
 //! This module implements a pseudo random function that is derived from
 //! the rescue hash function.
 
+use core::marker::PhantomData;
+
 use crate::{
     errors::PrimitivesError,
-    rescue::{Permutation, RescueParameter, STATE_SIZE},
+    rescue::{sponge::RescueSpongePRF, RescueParameter, STATE_SIZE},
 };
 use ark_ff::PrimeField;
 use ark_serialize::*;
@@ -24,7 +26,7 @@ pub struct PRF<F: RescueParameter> {
     pub input_len: usize,
     /// Length of the output.
     pub output_len: usize,
-    permutation: Permutation<F>,
+    phantom_f: PhantomData<F>,
 }
 
 #[derive(
@@ -61,7 +63,7 @@ impl<F: RescueParameter> PRF<F> {
         PRF {
             input_len,
             output_len,
-            permutation: Default::default(),
+            phantom_f: PhantomData,
         }
     }
 
@@ -84,8 +86,7 @@ impl<F: RescueParameter> PRF<F> {
         // Ok to pad with 0's since input length is fixed for the PRF instance
         let mut padded = input.to_owned();
         pad_with_zeros(&mut padded, STATE_SIZE);
-        self.permutation
-            .full_state_keyed_sponge_no_padding(&key.0, &padded, self.output_len)
+        RescueSpongePRF::full_state_keyed_sponge_no_padding(&key.0, &padded, self.output_len)
             .map_err(|_| {
                 PrimitivesError::InternalError("Bug in PRF: bad padding for input".to_string())
             })
--- a/primitives/src/prf.rs
+++ b/primitives/src/prf.rs
@@ -11,7 +11,7 @@ use ark_std::marker::PhantomData;
 
 use crate::{
     errors::PrimitivesError,
-    rescue::{sponge::RescueSpongePRF, RescueParameter, STATE_SIZE},
+    rescue::{sponge::RescuePRF, RescueParameter, STATE_SIZE},
 };
 use ark_ff::PrimeField;
 use ark_serialize::*;
@@ -86,10 +86,9 @@ impl<F: RescueParameter> PRF<F> {
         // Ok to pad with 0's since input length is fixed for the PRF instance
         let mut padded = input.to_owned();
         pad_with_zeros(&mut padded, STATE_SIZE);
-        RescueSpongePRF::full_state_keyed_sponge_no_padding(&key.0, &padded, self.output_len)
-            .map_err(|_| {
-                PrimitivesError::InternalError("Bug in PRF: bad padding for input".to_string())
-            })
+        RescuePRF::full_state_keyed_sponge_no_padding(&key.0, &padded, self.output_len).map_err(
+            |_| PrimitivesError::InternalError("Bug in PRF: bad padding for input".to_string()),
+        )
     }
 }
 
--- a/primitives/src/prf.rs
+++ b/primitives/src/prf.rs
@@ -7,7 +7,7 @@
 //! This module implements a pseudo random function that is derived from
 //! the rescue hash function.
 
-use core::marker::PhantomData;
+use ark_std::marker::PhantomData;
 
 use crate::{
     errors::PrimitivesError,
--- a/primitives/src/rescue/mod.rs
+++ b/primitives/src/rescue/mod.rs
@@ -831,7 +831,6 @@ mod test_permutation {
 
     fn test_sponge_helper<F: RescueParameter>() {
         let rescue_prp = PRP::default();
-        // let rescue_permutation = Permutation::from(rescue_prp.clone());
         let mut prng = ark_std::test_rng();
         let e0 = F::rand(&mut prng);
         let e1 = F::rand(&mut prng);
--- a/primitives/src/rescue/mod.rs
+++ b/primitives/src/rescue/mod.rs
@@ -699,7 +699,7 @@ mod test_prp {
 #[cfg(test)]
 mod test_permutation {
     use crate::rescue::{
-        sponge::{RescueSpongeCRHF, RescueSpongePRF},
+        sponge::{RescueCRH, RescuePRF},
         Permutation, RescueParameter, RescueVector, PRP,
     };
     use ark_bls12_377::Fq as Fq377;
@@ -841,7 +841,7 @@ mod test_permutation {
 
         let input = [e0, e1, e2, e3, e4, e5];
 
-        let output = RescueSpongeCRHF::<F>::sponge_no_padding(&input, 1).unwrap()[0];
+        let output = RescueCRH::<F>::sponge_no_padding(&input, 1).unwrap()[0];
 
         let zero = RescueVector::zero();
         let mut state = RescueVector {
@@ -868,7 +868,7 @@ mod test_permutation {
             Fr254::from_le_bytes_mod_order(&OUTPUT254[1]),
             Fr254::from_le_bytes_mod_order(&OUTPUT254[2]),
         ];
-        let real_output = RescueSpongeCRHF::sponge_no_padding(&input, 3).unwrap();
+        let real_output = RescueCRH::sponge_no_padding(&input, 3).unwrap();
         assert_eq!(real_output, expected);
     }
 
@@ -879,7 +879,7 @@ mod test_permutation {
             Fr377::from_le_bytes_mod_order(&OUTPUT377[1]),
             Fr377::from_le_bytes_mod_order(&OUTPUT377[2]),
         ];
-        let real_output = RescueSpongeCRHF::sponge_no_padding(&input, 3).unwrap();
+        let real_output = RescueCRH::sponge_no_padding(&input, 3).unwrap();
         assert_eq!(real_output, expected);
     }
 
@@ -890,7 +890,7 @@ mod test_permutation {
             Fr381::from_le_bytes_mod_order(&OUTPUT381[1]),
             Fr381::from_le_bytes_mod_order(&OUTPUT381[2]),
         ];
-        let real_output = RescueSpongeCRHF::sponge_no_padding(&input, 3).unwrap();
+        let real_output = RescueCRH::sponge_no_padding(&input, 3).unwrap();
         assert_eq!(real_output, expected);
     }
 
@@ -901,7 +901,7 @@ mod test_permutation {
             Fq377::from_le_bytes_mod_order(&OUTPUT761[1]),
             Fq377::from_le_bytes_mod_order(&OUTPUT761[2]),
         ];
-        let real_output = RescueSpongeCRHF::sponge_no_padding(&input, 3).unwrap();
+        let real_output = RescueCRH::sponge_no_padding(&input, 3).unwrap();
         assert_eq!(real_output, expected);
     }
 
@@ -915,28 +915,18 @@ mod test_permutation {
     fn test_fsks_no_padding_errors_helper<F: RescueParameter>() {
         let key = F::rand(&mut ark_std::test_rng());
         let input = vec![F::from(9u64); 4];
-        assert!(
-            RescueSpongePRF::full_state_keyed_sponge_no_padding(&key, input.as_slice(), 1).is_ok()
-        );
+        assert!(RescuePRF::full_state_keyed_sponge_no_padding(&key, input.as_slice(), 1).is_ok());
         let input = vec![F::from(9u64); 12];
-        assert!(
-            RescueSpongePRF::full_state_keyed_sponge_no_padding(&key, input.as_slice(), 1).is_ok()
-        );
+        assert!(RescuePRF::full_state_keyed_sponge_no_padding(&key, input.as_slice(), 1).is_ok());
 
         // test should panic because number of inputs is not multiple of 3
         let input = vec![F::from(9u64); 10];
-        assert!(
-            RescueSpongePRF::full_state_keyed_sponge_no_padding(&key, input.as_slice(), 1).is_err()
-        );
+        assert!(RescuePRF::full_state_keyed_sponge_no_padding(&key, input.as_slice(), 1).is_err());
         let input = vec![F::from(9u64)];
-        assert!(
-            RescueSpongePRF::full_state_keyed_sponge_no_padding(&key, input.as_slice(), 1).is_err()
-        );
+        assert!(RescuePRF::full_state_keyed_sponge_no_padding(&key, input.as_slice(), 1).is_err());
 
         let input = vec![];
-        assert!(
-            RescueSpongePRF::full_state_keyed_sponge_no_padding(&key, input.as_slice(), 1).is_ok()
-        );
+        assert!(RescuePRF::full_state_keyed_sponge_no_padding(&key, input.as_slice(), 1).is_ok());
     }
 
     #[test]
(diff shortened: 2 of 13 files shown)


tessico:
Btw, it goes a little further than https://github.com/EspressoSystems/jellyfish/pull/50, in that it also uses the new `absorb/squeeze` methods in the encapsulating types' implementation of `full_state_keyed_sponge`,  `sponge_no_padding` etc.

tessico:
@EspressoSystems/jellyfish

primitives/src/rescue/mod.rs
```diff
@@ -161,8 +163,7 @@ impl<F: PrimeField> RescueVector<F> {
         aux
     }
 
-    fn add_assign_elems(&mut self, elems: &[F]) {
-        assert_eq!(elems.len(), STATE_SIZE);
+    fn add_assign_elems(&mut self, elems: &[F], _expected_size: usize) {
```
mrain, line 166: Any reason to add a new but unused parameter here?

primitives/src/rescue/mod.rs
```diff
@@ -881,7 +831,7 @@ mod test_permutation {
 
     fn test_sponge_helper<F: RescueParameter>() {
         let rescue_prp = PRP::default();
-        let rescue_permutation = Permutation::from(rescue_prp.clone());
+        // let rescue_permutation = Permutation::from(rescue_prp.clone());
```
mrain, line 834: Remove this comment.

primitives/src/rescue/sponge.rs
```diff
@@ -0,0 +1,420 @@
+//! This file contains the APIs wrappers for ark-sponge
```
mrain, line 1: Missing copyright information.

primitives/src/rescue/sponge.rs
```diff
@@ -0,0 +1,420 @@
...
+}
+
+/// CRHF
+pub type RescueSpongeCRHF<F> = RescueSponge<F, 3>;
+/// PRF
+pub type RescueSpongePRF<F> = RescueSponge<F, 4>;
+
+impl<F: RescueParameter, const CHUNK_SIZE: usize> RescueSponge<F, CHUNK_SIZE> {
+    /// Sponge hashing based on rescue permutation for Bls12_381 scalar field
+    /// for RATE 3 and CAPACITY 1. It allows unrestricted variable length
```
mrain, line 27: Should be `RATE CHUNK_SIZE`.

primitives/src/rescue/sponge.rs
```diff
@@ -0,0 +1,420 @@
...
+    /// input and number of output elements
+    pub fn sponge_with_padding(input: &[F], num_outputs: usize) -> Vec<F> {
+        // Pad input as follows: append a One, then pad with 0 until length is multiple
+        // of RATE
+        let mut padded = input.to_vec();
+        padded.push(F::one());
+        pad_with_zeros(&mut padded, RATE);
```
mrain, line 34: Should here be `CHUNK_SIZE` or `RATE`?

primitives/src/rescue/sponge.rs
```diff
@@ -0,0 +1,420 @@
...
+pub struct RescueSponge<F: RescueParameter, const ABSORB_RATE: usize> {
+    pub(crate) state: RescueVector<F>,
+}
+
+/// CRHF
+pub type RescueSpongeCRHF<F> = RescueSponge<F, 3>;
+/// PRF
+pub type RescueSpongePRF<F> = RescueSponge<F, 4>;
+
+impl<F: RescueParameter, const CHUNK_SIZE: usize> RescueSponge<F, CHUNK_SIZE> {
```
mrain, line 25: This `CHUNK_SIZE` seems not used in the following `impl`.
Also previously it's called `ABSORB_RATE` in the struct definition. Perhaps it's better to have a unified name? E.g. we could call it `RATE`, and rename `super::RATE` to `super::RESCUE_CRHF_RATE`.

primitives/src/rescue/sponge.rs
```diff
@@ -0,0 +1,420 @@
...
+        // ABSORB PHASE
+        let mut r = Self::from_state(RescueVector::zero(), &Permutation::default());
+        r.absorb(&input);
+
+        // SQUEEZE PHASE
+        Ok(r.squeeze_native_field_elements(num_output))
+    }
+}
+
+impl<F: RescueParameter, const CHUNK_SIZE: usize> RescueSponge<F, CHUNK_SIZE> {
```
mrain, line 58: I have the same question as above.

mrain (changes requested):
Good work 👍  Rescue interface looks cleaner now.
There are a few questions left.

primitives/src/rescue/sponge.rs
alxiong: we should automate this later in our precommit hook

primitives/src/rescue/sponge.rs
```diff
@@ -0,0 +1,420 @@
...
+        sizes: &[FieldElementSize],
+    ) -> Vec<F> {
+        if T::size_in_bits() == F::size_in_bits() {
+            RescueSponge::<T, CHUNK_SIZE>::squeeze_native_field_elements_with_sizes(self, sizes)
+                .iter()
+                .map(|x| field_switching(x))
+                .collect::<Vec<F>>()
+        } else {
+            // currently we do not support hashing into a non-native field
+            unimplemented!()
```
alxiong, line 215:
```suggestion
            unimplemented!("currently we do not support hashing into a non-native field");
```

primitives/src/prf.rs
```diff
@@ -7,9 +7,11 @@
 //! This module implements a pseudo random function that is derived from
 //! the rescue hash function.
 
+use core::marker::PhantomData;
```
alxiong, line 10: curious, should we use from `core` or `ark_std`, (I think the latter just re-export the `core` correct?) I wonder if there would be version incompatibility issue if we don't consistently using those reexported from `ark_std`

alxiong (commented):
I like the refactoring! being explicit about what we are using Rescue for is very good.

- IMO, `RescueSpongeCRHF` should be `RescueSpongeCRH` or even simpler `RescueCRH`

primitives/src/prf.rs
tessico: Good catch. Indeed `ark-std` is just re-exporting `core::marker`. Let me replace this usage to ark_std, and make a separate PR that replaces all remaining usages (since the repo rn is inconsistent).

primitives/src/rescue/sponge.rs
tessico: Done in https://github.com/EspressoSystems/jellyfish/pull/146/commits/abbdfadf85a826ce5ca406243f682ff6c4a050ea.

primitives/src/rescue/mod.rs
tessico: Done in https://github.com/EspressoSystems/jellyfish/pull/146/commits/b2bb4c7d2dfd6f1ed8dd8b3de884374b7e5ddbfd.

primitives/src/rescue/mod.rs
tessico: That was a leftover from experimentation. Previously we only allowed the number of elements to have a certain size. Now the `absorb` method will take any number of elements, and the length check is enforced by the `_with_padding` methods of the structs using `absorb`, if desired.
Done in https://github.com/EspressoSystems/jellyfish/pull/146/commits/c094a0d0acaed02e2b0ce602006f48ab02e0246d.

primitives/src/rescue/sponge.rs
tessico: I unified it, but to `CHUNK_SIZE`. This is an internal detail of what the absorb phase does, it splits things into chunks. But I understand the confusion and it's a very valid point.
This prompted me to go and enforce that `SpongeCRH` is **always** using a `CHUNK_SIZE = 3`, while the PRF is using `CHUNK_SIZE = 4`: see https://github.com/EspressoSystems/jellyfish/pull/146/commits/ef81c461e154bb3ba8041b7406d8a9d03ca1d954 & https://github.com/EspressoSystems/jellyfish/pull/146/commits/97e9c0899351196b7e8292bd10125bcf183373bb.
Let me know what you think.

primitives/src/rescue/sponge.rs
tessico: I'm now enforcing these to be the same in this implementation, see comment below.

primitives/src/rescue/sponge.rs
```diff
@@ -0,0 +1,441 @@
...
+/// A rescue hash function consists of a permutation function and
+/// an internal state.
+pub struct RescueSponge<F: RescueParameter, const CHUNK_SIZE: usize> {
+    pub(crate) state: RescueVector<F>,
+}
+
+/// CRHF
+pub type RescueCRH<F> = RescueSponge<F, RATE>;
+/// PRF
+pub type RescuePRF<F> = RescueSponge<F, STATE_SIZE>;
```
alxiong, line 29: Now I tend to think that we should use newtype or some wrapper struct here instead of type alias. 

`RescueCRH` should follow a normal CRH API (something [like this](https://github.com/arkworks-rs/crypto-primitives/blob/main/src/crh/mod.rs)) and internally it can use all the `RescueSponge`'s API. and lives in `primitives/src/rescue/crh.rs` probably.

same for `RescuePRF`.

Maybe we can do this in a separate PR.

primitives/src/rescue/sponge.rs
```diff
@@ -0,0 +1,441 @@
...
+            }
+        }
+
+        if all_full_sizes {
+            self.squeeze_native_field_elements(sizes.len())
+        } else {
+            // we do not currently want to output field elements other than T.
+            // This will be fixed once `squeeze_bytes` interfaces is fixed.
+            unimplemented!()
+        }
```
alxiong, line 304: again, fill unimplemented with reason string.

primitives/src/rescue/sponge.rs
alxiong: I also personally find `ABSORB_RATE` a clearer name.

either way, we need to add code documentation somewhere (non-public, just `//`) to explain the differences between these parameterizations.

primitives/src/rescue/sponge.rs
```diff
@@ -0,0 +1,441 @@
...
+}
+
+impl<T: RescueParameter + PrimeField, const CHUNK_SIZE: usize> CryptographicSponge
+    for RescueSponge<T, CHUNK_SIZE>
+{
+    /// Parameters used by the sponge.
+    type Parameters = Permutation<T>;
+
+    /// Initialize a new instance of the sponge.
+    fn new(_permutation: &Self::Parameters) -> Self {
```
chancharles92, line 137: How do we support a keyed sponge where the initial state is an input key rather than the default value?

chancharles92 (changes requested):

primitives/src/rescue/sponge.rs
```diff
@@ -0,0 +1,441 @@
...
+
+#[derive(Clone, Default)]
+/// A rescue hash function consists of a permutation function and
+/// an internal state.
+pub struct RescueSponge<F: RescueParameter, const CHUNK_SIZE: usize> {
+    pub(crate) state: RescueVector<F>,
+}
+
+/// CRHF
+pub type RescueCRH<F> = RescueSponge<F, RATE>;
```
mrain, line 27: Sorry to be picky. I felt like, `RescueCRHF` is a better name here.
WDYT @chancharles92 @alxiong

primitives/src/rescue/sponge.rs
tessico: Newtype probably makes sense.
I think in this PR I could still refactor it a bit more to separate out CRH and PRF (former having `sponge_with_padding` & `sponge_no_padding` methods, latter with `full_state_keyed_sponge_with_padding` & `full_state_keyed_sponge_no_padding`.

And potentially in a separate follow-up we could make our API follow the crypto-primitives trait from arkworks you 
linked.

WDYT? Also @EspressoSystems/jellyfish

primitives/src/rescue/sponge.rs
alxiong: hahah, sorry, I suggested the `CRH` name, probably because I'm influenced by the [arkwork's naming](https://github.com/arkworks-rs/crypto-primitives/tree/main/src/crh) convention and Aleo's snarkVM's naming. 

I'm okay with both, and my apology to @tessico first in case we change this back haha. 😜

primitives/src/rescue/sponge.rs
tessico: Do you mean actually passing a `Permutation` parameter? Is [this](https://github.com/EspressoSystems/jellyfish/pull/146/commits/de0ad5320f9cccdbd076ce290620acfd0487bb89) what you mean?

primitives/src/rescue/sponge.rs
```diff
@@ -0,0 +1,441 @@
...
+{
+    /// Squeeze `num_elements` field elements from the sponge.
+    fn squeeze_native_field_elements(&mut self, num_elements: usize) -> Vec<T> {
+        // SQUEEZE PHASE
+        let permutation = Permutation::default();
+        let mut result = vec![];
+        let mut remaining = num_elements;
+        // extract current rate before calling PRP again
+        loop {
+            let extract = remaining.min(RATE);
```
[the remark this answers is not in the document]
tessico: Yes, I think so. Good catch. https://github.com/EspressoSystems/jellyfish/pull/146/commits/3ec4b32118d5055f11ed9ebac731574ac14bee4a

primitives/src/rescue/sponge.rs
```diff
@@ -0,0 +1,441 @@
...
+        } else {
+            // we do not currently want to output field elements other than T.
+            // This will be fixed once `squeeze_bytes` interfaces is fixed.
+            unimplemented!()
+        }
+    }
+}
+
+#[inline]
+fn bools_to_u8(input: &[bool]) -> u8 {
```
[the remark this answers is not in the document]
tessico: N/A due to https://github.com/EspressoSystems/jellyfish/pull/146/commits/8ef1f140705bac8ce639b934cb14576e82a45086.

primitives/src/rescue/sponge.rs
```diff
@@ -0,0 +1,441 @@
...
+            .map(bools_to_u8)
+            .collect()
+    }
+
+    /// Squeeze `num_bits` bits from the sponge.
+    fn squeeze_bits(&mut self, num_bits: usize) -> Vec<bool> {
+        // we extract k number of field elements;
+        // each field elements will produce a maximum
+        // ```
+        // T::size_in_bits() - 129
```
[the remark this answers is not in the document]
tessico: N/A due to https://github.com/EspressoSystems/jellyfish/pull/146/commits/8ef1f140705bac8ce639b934cb14576e82a45086.

I will add the comments later as a warning. Also, in the `unimplemented()` we can add an extra message, I'll do that too.

primitives/src/rescue/sponge.rs
```diff
@@ -0,0 +1,441 @@
...
+            .into_iter()
+            .for_each(|chunk| {
+                self.state.add_assign_elems(chunk);
+                self.state = permutation.eval(&self.state)
+            });
+    }
+
+    /// Squeeze `num_bytes` bytes from the sponge.
+    fn squeeze_bytes(&mut self, num_bytes: usize) -> Vec<u8> {
+        self.squeeze_bits(num_bytes)
```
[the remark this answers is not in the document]
tessico: N/A due to https://github.com/EspressoSystems/jellyfish/pull/146/commits/8ef1f140705bac8ce639b934cb14576e82a45086.

primitives/src/rescue/sponge.rs
chancharles92: No, I mean to pass an initial `state`. Sometimes the initial state might be a secret rather than a default vector.

tessico: Ah, for that you can use the `from_state` associated function, on the `SpongeExt` trait.

primitives/src/rescue/sponge.rs
tessico: Done, unified all the unimplemented messages to: "Currently we only support squeezing native field elements!" in https://github.com/EspressoSystems/jellyfish/pull/146/commits/995ec471d37e5b3ef55cbdacee127cdd2c8afabd.

primitives/src/rescue/sponge.rs
tessico: I'm also neutral about this one. As seemingly there's preference for `CRHF`, I renamed it in the [latest commit](https://github.com/EspressoSystems/jellyfish/pull/146/commits/ae2bbe1b3a8de6ac7331e0bdc1c01dbe3e2aa3ff).

primitives/src/rescue/sponge.rs
tessico: @mrain After some thought, I went with your suggestion and renamed it to `RATE`.
@alxiong Since the sponge actually uses this for both absorb and squeeze step, I think `RATE` is more appropriate than `ABSORB_RATE` in the end. The reason it was `ABSORB_RATE` in the initial version was because of a mismatch between our usage (always squeeze 1 element) and the general squeeze API, which was pointed out by @chancharles92 [in this comment](https://github.com/EspressoSystems/jellyfish/pull/146#discussion_r1035996035).

mrain (approved):
LGTM!

primitives/src/rescue/sponge.rs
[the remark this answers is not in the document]
chancharles92: Hi @tessico , let me know when the warning comments are done. I'll then approve.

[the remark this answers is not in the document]
tessico: @chancharles92 I added the comments in https://github.com/EspressoSystems/jellyfish/pull/146/commits/2c817a4baf3b2fe9718ef03b3aaab18dbfaeca4b.
What I've also done in that commit was to completely "disable" the unintended usage of `CryptographicSponge`.
For our use case, we only want to allow users to squeeze **full native field elements**, meaning e.g. we shouldn't support truncated elements squeezing, as was previously done in e.g. `squeeze_native_field_elements_with_sizes`.

Under the hood, these methods would revert to `squeeze_native_field_elements` anyway if the `sizes` was full, and panic otherwise. The new changes reflect that under-the-hood behaviour, but are more explicit about what's supported and what's not, whereas previously a user might be tricked into using `squeeze_native_field_elements_with_sizes`, but then the method could panic if the input was not as expected.

chancharles92 (approved):

mrain (approved):

alxiong (approved):

Issue #127, opened 2022-10-24
Minimize unnecessary dependency


alxiong:
As discussed last week, we may want to spend time minimize unnecessary/unused dependencies for better build time and security reason.

consider
- add [cargo-udeps](https://github.com/est31/cargo-udeps) in CI to detect unused crates

alxiong:
we already have a cron job.

PR #147, opened 2022-11-25
Clean up deps

tessico:
## Description

Related to https://github.com/EspressoSystems/jellyfish/issues/127, but not closing it yet.
This is just an easy win since `cargo-udeps` might still leave some unused deps undetected. At some point I'll review them one-by-one.

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,6 +24,20 @@ jobs:
       - name: Checkout Repository
         uses: actions/checkout@v3
 
+      - name: Install nightly toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
+          override: true # just temporarily for cargo-udeps
+          default: false
+          components: rustfmt
+
+      - name: Run cargo-udeps
+        uses: aig787/cargo-udeps-action@v1
+        with:
+          args: '--all-targets'
+
       - name: Install stable toolchain
         uses: actions-rs/toolchain@v1
         with:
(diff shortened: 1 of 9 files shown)


alxiong:
is there a way to add this check in our CI? (even if there's false positive, it will be better than nothing)

tessico:
With a small hackaround I've added the [cargo-udeps-action](https://github.com/aig787/cargo-udeps-action).
The caveat there is that `cargo-udeps` requires `nightly` toolchain, but placing this action in an expected place (e.g. after the audit step) would default to `stable`, thus [failing](https://github.com/tessico/jellyfish/actions/runs/3620978377/jobs/6103852259) the job. 

I've opened an issue in the action's repo: https://github.com/aig787/cargo-udeps-action/issues/1, currently using the workaround described there.

alxiong:
I see, is it correct to conclude that just running `cargo +nightly udeps` in CI will require 2~3 min alone? (downloading nightly toolchain can be cached and not counted here), that's way slower than I expect. 

Thanks for the effort of different experimentations.

A possible alternative is to add this to our pre-commit hook instead. which I think is probably better, wdyt?

tessico:
@alxiong Well no, just running udeps by itself will be faster than that. My point was that most of the 2-3 minutes is spent on the download.

tessico:
Locally it takes ~18s, I expect it to stay under 30s on the GitHub ubuntu machine, discounting the download.
I honestly would prefer to not spend another 18s on waiting for a local commit.

tessico:
Maybe we can make it a separate cron job, say once a week on the jellyfish repo.

tessico:
I've moved the check to its own workflow. It is triggered weekly as a cron, or otherwise can be triggered manually, e.g. see [this run](https://github.com/tessico/jellyfish/actions/runs/3640096070/jobs/6144318827).

Currently jobs are failing due to an update in espresso-common re tags, see https://github.com/EspressoSystems/jellyfish/issues/155.

alxiong (approved):

Issue #142, opened 2022-11-18
Refactor `MerkleTreeGadget` to be generic over `HashGadget`


tessico:
The [WIP](https://github.com/EspressoSystems/jellyfish/pull/138) merkle tree circuit refactor is Rescue-hash specific.
We should have code that's polymorphic over different hash gadgets.

Tasks:
- [ ] Introduce a `HashGadget` and make `RescueGadget` implement it
- [ ] Make `MerkleTreeGadget` generic over `HashGadget`

See discussion here: https://github.com/EspressoSystems/jellyfish/pull/138#discussion_r1026763507

tessico:
Indirectly related to better sponge APIs: https://github.com/EspressoSystems/jellyfish/issues/15

PR #150, opened 2022-11-29
Unify sponge gadget

tessico:
## Description

Builds on top of https://github.com/EspressoSystems/jellyfish/pull/146. Still WIP, but I'd like to get some comments early on to avoid too much (or rather: even more...) refactoring later.

Refactor native & non-native sponge gadgets:
- [x]  add a `RescueStateVarGen` trait
  - [x] implement for both native and non-native RescueStateVar
  - [ ] discuss naming: this could be used outside of the "Rescue" context - it only specifies which fields are native vs. non-native, and how the underlying variables are represented. Perhaps `GenericStateVar` would be better? Other ideas?
- [x] pull out a unified `SpongeGadget` trait into `mod.rs`, generic over `RescueStateVarGen`
- [x] pull out a unified `PermutationGadget` trait into `mod.rs`, generic over `RescueStateVarGen`

Downside:
- since there are now two implementations of the methods in `RescueGadget`, the downstream users need to disambiguate the call, e.g. instead of:
```
self.rescue_sponge_no_padding(&input_labels, 1);
```
now you need to write
```
RescueGadget::<RescueStateVar, F, F>::rescue_sponge_no_padding(self, &input_labels, 1);
```

I feel that the ergonomics of the call is worse.
If you have any idea on how to keep a single `SpongeGadget` trait instead of `RescueGadget/RescueNonNativeGadget`, but without requiring disambiguation on every single call, please share :)

On the other hand, we should soon be able to benefit from being generic, for example: generic over a `DigestAlgorithmGadget` in the implementation of the `MerkleTreeGadget`, in line with https://github.com/EspressoSystems/jellyfish/issues/142:
```
pub trait DigestAlgorithmGadget<R, T, F>
where
    R: RescueStateVarGen<T, F>,
{
    fn digest(&mut self, data: &[Variable]) -> Variable;
    fn digest_leaf(pos: &usize, elem: &Variable) -> Variable;
}

impl<F: PrimeField + RescueParameter> DigestAlgorithmGadget<RescueStateVar, F, F> {...}

impl<F, D> MerkleTreeGadget<F, F, RescueStateVar, D> for PlonkCircuit<F>
where
    F: PrimeField + RescueParameter,
    D: DigestAlgorithmGadget<RescueStateVar, F, F>,
{
    fn compute_merkle_root(
        &mut self,
        elem: CommittedElemVar,
        path_vars: &MerklePathVar,
    ) -> Result<Variable, CircuitError> {
        let zero_var = self.zero();

        // leaf label = H(0, uid, arc)
        self.digest(&[zero_var, elem.uid, elem.elem]);
      ...
```

closes: #XXXX

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/primitives/src/circuit/rescue/native.rs
+++ b/primitives/src/circuit/rescue/native.rs
@@ -24,6 +24,7 @@ use super::{PermutationGadget, RescueGadget, RescueStateVarGen};
 /// Array of variables representing a Rescue state (4 field elements).
 pub struct RescueStateVar(pub(crate) [Variable; STATE_SIZE]);
 
+/// For the native field, there is only really one field `F`.
 impl<F> RescueStateVarGen<F, F> for RescueStateVar {
     type Native = F;
     type NonNative = F;
--- a/primitives/src/circuit/rescue/native.rs
+++ b/primitives/src/circuit/rescue/native.rs
@@ -652,7 +652,7 @@ mod tests {
 
     use super::{RescueGadget, RescueHelperGadget, RescueStateVar};
     use crate::rescue::{
-        sponge::{RescueCRH, RescuePRF},
+        sponge::{RescueCRHF, RescuePRF},
         Permutation, RescueMatrix, RescueParameter, RescueVector, PRP, RATE, STATE_SIZE,
     };
     use ark_ed_on_bls12_377::Fq as FqEd377;
@@ -972,7 +972,7 @@ mod tests {
             .map(|&x| circuit.create_variable(x).unwrap())
             .collect_vec();
 
-        let expected_sponge = RescueCRH::sponge_no_padding(&data, 1).unwrap()[0];
+        let expected_sponge = RescueCRHF::sponge_no_padding(&data, 1).unwrap()[0];
         let sponge_var = circuit
             .rescue_sponge_no_padding(data_vars.as_slice(), 1)
             .unwrap()[0];
@@ -1028,7 +1028,7 @@ mod tests {
             }
 
             // Check consistency between outputs
-            let expected_hash = RescueCRH::sponge_no_padding(&input_vec, output_len).unwrap();
+            let expected_hash = RescueCRHF::sponge_no_padding(&input_vec, output_len).unwrap();
 
             for (e, f) in out_var.iter().zip(expected_hash.iter()) {
                 assert_eq!(*f, circuit.witness(*e).unwrap());
@@ -1089,7 +1089,7 @@ mod tests {
                 }
 
                 // Check consistency between outputs
-                let expected_hash = RescueCRH::sponge_with_padding(&input_vec, output_len);
+                let expected_hash = RescueCRHF::sponge_with_padding(&input_vec, output_len);
 
                 for (&e, &f) in expected_hash.iter().zip(out_var.iter()) {
                     assert_eq!(e, circuit.witness(f).unwrap());
--- a/primitives/src/circuit/rescue/native.rs
+++ b/primitives/src/circuit/rescue/native.rs
@@ -653,7 +653,7 @@ mod tests {
     use super::{RescueGadget, RescueHelperGadget, RescueStateVar};
     use crate::rescue::{
         sponge::{RescueCRHF, RescuePRF},
-        Permutation, RescueMatrix, RescueParameter, RescueVector, PRP, RATE, STATE_SIZE,
+        Permutation, RescueMatrix, RescueParameter, RescueVector, CRHF_RATE, PRP, STATE_SIZE,
     };
     use ark_ed_on_bls12_377::Fq as FqEd377;
     use ark_ed_on_bls12_381::Fq as FqEd381;
@@ -966,7 +966,7 @@ mod tests {
         let mut circuit = PlonkCircuit::new_turbo_plonk();
 
         let mut prng = ark_std::test_rng();
-        let data = (0..2 * RATE).map(|_| F::rand(&mut prng)).collect_vec();
+        let data = (0..2 * CRHF_RATE).map(|_| F::rand(&mut prng)).collect_vec();
         let data_vars = data
             .iter()
             .map(|&x| circuit.create_variable(x).unwrap())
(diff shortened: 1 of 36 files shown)


primitives/src/circuit/rescue/native.rs
```diff
@@ -97,95 +106,7 @@ impl<F: PrimeField> Gate<F> for Power5NonLinearGate<F> {
...
-    fn prp_with_round_keys(
-        &mut self,
-        input_var: &RescueStateVar,
-        mds: &RescueMatrix<F>,
-        keys_vars: &[RescueStateVar],
-    ) -> Result<RescueStateVar, CircuitError>;
-}
-
-impl<F> RescueGadget<F> for PlonkCircuit<F>
+impl<F> RescueGadget<RescueStateVar, F, F> for PlonkCircuit<F>
```
chancharles92, line 109: Would it be helpful to have a type-alias or wrapper for the native/non-native gadget's API?

E.g.,  currently, we need to write: `RescueGadget::<RescueStateVar, F, F>::rescue_sponge_no_padding`.
It'll be great if the user only needs to write `RescueNativeGadget::<F>::rescue_sponge_no_padding`.

Similarly, for non-native setting, it'll be great if the user only needs to write `RescueNonNativeGadget<F, T>::rescue_sponge_no_padding`.

tessico: Yes, that would indeed improve the ergonomics of using the gadget. I'll have a look, thx for the suggestion!

mrain (approved):
Generally LGTM. Better if Binyi's comment is addressed.

tessico:
@chancharles92 I've introduced a type alias for the trait.
The downside is that we need to use a dynamic trait now, which as @alxiong has pointed out in some other PR for VRFs, can have a potential performance impact. I haven't ran the benches for this specific case yet though. WDYT?
Also @mrain.

alxiong:
> I've introduced a type alias for the trait. The downside is that we need to use a dynamic trait now,

So trait alias feature is available in rust nightly (see [unstable book](https://doc.rust-lang.org/unstable-book/language-features/trait-alias.html))
Personally I would stick with the original, more verbose and troublesome way of declaring `RescueGadget::<RescueStateVar, F, F>::rescue_sponge_no_padding` and add an issue tracking the progress on the stabilization of that rust RFC. 

Because after all it's an internal ergonomic issue, and we can tolerate this temporarily imo.

chancharles92:
> > I've introduced a type alias for the trait. The downside is that we need to use a dynamic trait now,
> 
> So trait alias feature is available in rust nightly (see [unstable book](https://doc.rust-lang.org/unstable-book/language-features/trait-alias.html)) Personally I would stick with the original, more verbose and troublesome way of declaring `RescueGadget::<RescueStateVar, F, F>::rescue_sponge_no_padding` and add an issue tracking the progress on the stabilization of that rust RFC.
> 
> Because after all it's an internal ergonomic issue, and we can tolerate this temporarily imo.

Can we create new traits `RescueNativeGadget` and `RescueNonNativeGadget` that inherit corresponding traits? Then we achieve the same as a type alias. (Check here: https://www.worthe-it.co.za/blog/2017-01-15-aliasing-traits-in-rust.html). 
We can create an issue and use type alias again after it's stable. The advantage over keeping `RescueGadget::<RescueStateVar, F, F>::rescue_sponge_no_padding` is that we don't need to change them again everywhere, which is a waste of effort.

tessico:
> can we create new traits RescueNativeGadget and RescueNonNativeGadget that inherit corresponding traits? 

No, unfortunately not cleanly, I tried that before as well. I think it's the same issue under the hood, we will run into needing to use the trait as dynamic.

For example, we could define:
```
// Instead of defining a type alias, define a new supertrait
pub trait RescueNativeGadget<F>: RescueGadget<RescueStateVar, F, F> {}
// blanket impl needed! Otherwise we can't use the methods defined on `PlonkCircuit`
impl<F: RescueParameter> RescueNativeGadget<F> for PlonkCircuit<F> {}
```

But then it requires you to use it as:
```
<dyn RescueNativeGadget::<F>>::rescue_permutation(self, state_var)?;
```

I think the current approach is better in that we explicitly define it as `dyn TraitName`, and then case use it as if it was just a normal trait.

If we decide to go with current design, we can definitely add a tracking issue in jellyfish for a trait alias.

alxiong:
Alright, good with me! I'll leave the issue creation to you Marti, thanks!

chancharles92:
> > can we create new traits RescueNativeGadget and RescueNonNativeGadget that inherit corresponding traits?
> 
> No, unfortunately not cleanly, I tried that before as well. I think it's the same issue under the hood, we will run into needing to use the trait as dynamic.
> 
> For example, we could define:
> 
> ```
> // Instead of defining a type alias, define a new supertrait
> pub trait RescueNativeGadget<F>: RescueGadget<RescueStateVar, F, F> {}
> // blanket impl needed! Otherwise we can't use the methods defined on `PlonkCircuit`
> impl<F: RescueParameter> RescueNativeGadget<F> for PlonkCircuit<F> {}
> ```
> 
> But then it requires you to use it as:
> 
> ```
> <dyn RescueNativeGadget::<F>>::rescue_permutation(self, state_var)?;
> ```
> 
> I think the current approach is better in that we explicitly define it as `dyn TraitName`, and then case use it as if it was just a normal trait.
> 
> If we decide to go with current design, we can definitely add a tracking issue in jellyfish for a trait alias. @chancharles92 @alxiong

I see, sounds good to me. Let's add an issue for that, then we can approve.

tessico:
Tracking issue here: https://github.com/EspressoSystems/jellyfish/issues/165.

chancharles92:
It looks like we are still using type alias. Are we going with the type alias approach or the original solution as @alxiong suggested? (@alxiong I'm fine with either way, pls take care of the approval.)

tessico:
> I think the current approach is better in that we explicitly define it as dyn TraitName, and then case use it as if it was just a normal trait.

Hmm I thought we were sticking with the type alias. That was the "current solution" at the time of writing :)

alxiong (approved):

chancharles92 (approved):

PR #156, opened 2022-12-07
Vrf Audit

alxiong:
## Description

This PR fix issues raised in our [self-audit](https://www.notion.so/espressosys/VRF-Code-Self-audit-165b6ff54444431f9179eb155a2e73c5) on BLS implementations. 

Major changes include:
- Add `Zeroizing::new(ikm)` internally
- Add warning and rustdoc examples to remind users to provide good RNG
- Add `key_gen_v5` to allow generation of multiple keys using the same `ikm`
- Update Ciphersuite ID to conform with standard
- Add options to deserialize compressed, unchecked (skipping subgroup check) for Signature and PublicKey.

---

write a little note why.

- [x] Wrote unit tests


primitives/src/signatures/bls.rs
```diff
@@ -1,24 +1,277 @@
...
+//!
+//! # Examples
+//!
+//! ```
+//! use rand_core::{RngCore, OsRng};
+//! use jf_primitives::signatures::{SignatureScheme, bls::BLSSignatureScheme};
+//!
+//! let pp = BLSSignatureScheme::param_gen::<OsRng>(None)?;
+//!
+//! // make sure the PRNG passed in is securely seeded, we RECOMMEND using `OsRng`
```
mrain, line 23: We should be careful about this recommendation. 

From the doc, it says
```
You may prefer to use [EntropyRng](https://docs.rs/rand/0.5.0/rand/rngs/struct.EntropyRng.html) instead of OsRng. It is unlikely, but not entirely theoretical, for OsRng to fail. In such cases [EntropyRng](https://docs.rs/rand/0.5.0/rand/rngs/struct.EntropyRng.html) falls back on a good alternative entropy source.
```

mrain: Oh nvm that EntropyRng is deprecated in the newest version.
My point is, we should only make concrete recommendation if we are sure about it.

alxiong: The main reasons that `OsRng` might fail are: 1) process gets blocked since it was just initialized and entropy pool is still warming up; 2) in an environment that doesn't give you such access, such as a restricted wasm. 

On one hand, I'm fairly confident that this is a secure recommendation. 
OTOH, I can tone down a bit and say "you could use something like `OsRng` or seed your `SeedableRng` with some os-level randomness".

do you have suggestions on how should I rephrase it?

mrain: >OTOH, I can tone down a bit and say "you could use something like OsRng or seed your SeedableRng with some os-level randomness".

I think this is a good one. Plus a warning that user should use a high entropy random seed in production.

primitives/src/constants.rs
```diff
@@ -9,12 +9,17 @@
 /// ciphersuite identifier for schnorr signature
 pub const CS_ID_SCHNORR: &str = "SCHNORR_WITH_RESCUE_HASH_v01";
 
-/// ciphersuite identifier for BLS signature
-pub const CS_ID_BLS_SIG_NAIVE: &str = "BLS_SIG_WITH_NAIVE_HtG_v01";
+/// ciphersuite identifier for BLS signature, see:
+/// <https://www.ietf.org/archive/id/draft-irtf-cfrg-bls-signature-05.html#name-ciphersuite-format>
+pub const CS_ID_BLS_MIN_SIG: &str = "BLS_SIG_BLS12381G1_XMD:SHA-256_SSWU_RO_NUL_";
```
tessico, line 14: We are definitely not using SSWU hash-to-curve algorithm, but instead a rejection sampling.
I don't know what we should call our current impl, since the identifiers for rejection sampling are not covered by the IRTF draft, but definitely not this.

tessico, line 14:
```suggestion
pub const CS_ID_BLS_MIN_SIG: &str = "BLS_SIG_BLS12381G1_XMD:SHA-256_REJECTION_SAMPLING";
```

primitives/src/signatures/bls.rs
```diff
@@ -1,157 +1,316 @@
...
 use tagged_base64::tagged;
+use zeroize::{Zeroize, Zeroizing};
 
-pub use blst::min_sig::{PublicKey, SecretKey, Signature};
-use zeroize::Zeroize;
-
-/// Newtype wrapper for a BLS Signing Key.
 #[tagged(tag::BLS_SIGNING_KEY)]
-#[derive(Clone, Debug, Zeroize)]
+#[derive(Clone, Debug, Default, Zeroize)]
```
tessico, line 95: I'm not sure whether `Default` is a good idea on secret key. It might lead to users abusing it and causing insecure instantiations of the protocol.

primitives/src/signatures/bls.rs
```diff
@@ -1,157 +1,316 @@
...
+        if len != BLS_SIG_SK_SIZE {
             return Err(SerializationError::InvalidData);
         }
 
-        let mut key = [0u8; BLS_SIG_KEY_SIZE];
-        reader.read_exact(&mut key)?;
-        SecretKey::deserialize(&key)
+        let mut sk_bytes = [0u8; BLS_SIG_SK_SIZE];
+        reader.read_exact(&mut sk_bytes)?;
+        SecretKey::from_bytes(&sk_bytes)
```
tessico, line 126: `from_bytes` is just a wrapper on `deserialize`

primitives/src/signatures/bls.rs
```diff
@@ -1,157 +1,316 @@
...
     fn eq(&self, other: &Self) -> bool {
-        self.0.serialize() == other.0.serialize()
+        // constant time comparison
+        let xor_res: Vec<u8> = self
+            .to_bytes()
+            .iter()
+            .zip(other.to_bytes().iter())
+            .map(|(a, b)| a ^ b)
+            .collect();
+        xor_res == [0u8; 32]
```
tessico, line 141: under the hood you're calling the same methods anyway, since `to_bytes == serialize`. 
On top of that you're iterating and XORing values, I think the previous method is more performant, and no less constant time.

primitives/src/signatures/bls.rs
```diff
@@ -219,12 +377,43 @@ impl SignatureScheme for BLSSignatureScheme {
...
+        salt: &[u8],
+        key_info: &[u8],
+    ) -> Result<
+        (
+            <Self as SignatureScheme>::SigningKey,
+            <Self as SignatureScheme>::VerificationKey,
+        ),
+        PrimitivesError,
+    > {
+        let sk = SecretKey::key_gen_v5(ikm, salt, key_info)?;
```
tessico, line 405: Great that this comes from blst already!

primitives/src/constants.rs
alxiong: why not? I think my previous notion writeup might confuse you. We are not using our own H2C implementation here, we are just calling `blst::SecretKey::sign()` which internally uses SSWU version of H2C.

primitives/src/signatures/bls.rs
alxiong: yeah, I hesitated also. Reason I added was that I saw the underlying type from `blst` have it.
I'll remove it.

primitives/src/signatures/bls.rs
alxiong: previous methods is doing comparison of two slices? `PartialEq for [u8]` under the hood use `.all()` on the iterator, which will early return, thus no longer constant time. 
(altho admittedly our library has never put constant-time as a high priority, kind of out-of-whim to do it here. 🤔 )

primitives/src/signatures/bls.rs
alxiong: yes, I know. Okay, I will change to your impl.

primitives/src/signatures/bls.rs
alxiong: I decide to switch back to your original impl. If we are concerned about constant-time, we should do a more thorough audting in the future.

primitives/src/signatures/bls.rs
```diff
@@ -1,157 +1,316 @@
...
+    fn deserialize_uncompressed<R: Read>(mut reader: R) -> Result<Self, SerializationError> {
+        let len = <usize as ark_serialize::CanonicalDeserialize>::deserialize(&mut reader)?;
+        if len != BLS_SIG_PK_SIZE {
+            return Err(SerializationError::InvalidData);
+        }
+
+        let mut pk_bytes = [0u8; BLS_SIG_PK_SIZE];
+        reader.read_exact(&mut pk_bytes)?;
+
+        let pk = PublicKey::deserialize(&pk_bytes).map_err(|_| SerializationError::InvalidData)?;
```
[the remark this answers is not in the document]
alxiong: most comments above addressed in [04e6f72](https://github.com/EspressoSystems/jellyfish/pull/156/commits/04e6f727bf001aa2c13acdce74e8f826f6dad425)

primitives/src/signatures/bls.rs
tessico: > `PartialEq for [u8]` under the hood use `.all()` on the iterator, which will early return

Ah, I wasn't aware of that. Thanks for pointing it out.

But yes in this case I think it's safer to put a simpler impl in place, than start worrying about const time algorithm for `PartialEq` :)

primitives/src/constants.rs
tessico: Ah right. I was confused indeed.

primitives/src/signatures/bls.rs
```diff
@@ -235,27 +398,36 @@ mod test {
...
+        let (sk, pk) = BLSSignatureScheme::key_gen(&pp, &mut rng).unwrap();
+        let msg = "The quick brown fox jumps over the lazy dog";
+        let sig = BLSSignatureScheme::sign(&pp, &sk, &msg, &mut rng).unwrap();
+
+        test_canonical_serde_helper(sk);
+        test_canonical_serde_helper(pk);
+        test_canonical_serde_helper(sig);
+    }
+
+    // TODO: (alex) update this after upgrading to arkwork 0.4.0
```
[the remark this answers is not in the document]
tessico: https://github.com/arkworks-rs/algebra/blob/master/CHANGELOG.md

tessico (dismissed):
The TODO doesn't hurt though, it was just FYI.

primitives/src/signatures/bls.rs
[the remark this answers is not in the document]
alxiong: oh yeah, that's exactly what I meant. `CanonicalDeserialize` have compete 4 combination of (un)compressed + (un)check. which makes more sense.

tessico (approved):

Issue #99, opened 2022-08-11
[Constraint] Interval Merkle Tree Gadget


alxiong:
This issue aims to implement gadget related to IMT, including non-membership proof and insertion. 

Blocked by #98

PR #158, opened 2022-12-13
Ergonomic `MerkleTreeGadget`

tessico:
## Description

closes: #https://github.com/EspressoSystems/jellyfish/issues/88
as well as some indirect progress towards https://github.com/EspressoSystems/jellyfish/issues/142.

Note that non membership proofs are out of scope of this PR, but we can still close the referenced issue #88, if approved, since we already have a separate issue for non-membership proof constraints: https://github.com/EspressoSystems/jellyfish/issues/99.

Regarding the design, the initial refactor I had started doing was based on the API proposed by @alxiong in #88 - thanks a lot for this! You will see that it has changed in that methods don't accept native field elements/structs, but rather variables representing these items in the circuit. This was done for two reasons:
- There would otherwise be no way to (efficiently) enforce multiple constraints with the same proof. E.g. given a Merkle path, it would be impossible (without allocating the path variables twice) to enforce that e.g. one element is a member and another is not.
- API consistency. Most of our other APIs follow the format of `create_<>_variable -> Result<SomeVariable, ()>` and then using `SomeVariable` in enforcing constraints. I feel that this is a clean abstraction, and we should probably stick to not enforcing constraints "directly" from native items.

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/primitives/src/circuit/merkle_tree/mod.rs
+++ b/primitives/src/circuit/merkle_tree/mod.rs
@@ -11,15 +11,6 @@ use jf_relation::{errors::CircuitError, BoolVar, Variable};
 
 mod rescue_merkle_tree;
 
-/// Circuit variable for a leaf element.
-#[derive(Debug, Clone)]
-pub struct LeafVar {
-    /// Position of the leaf element in the MT. Serves as UID.
-    pub uid: Variable,
-    /// The value of the leaf element.
-    pub elem: Variable,
-}
-
 /// Gadget for a Merkle tree
 ///
 /// # Examples
@@ -49,8 +40,11 @@ pub trait MerkleTreeGadget<M>
 where
     M: MerkleTreeScheme,
 {
-    /// Type to represent the merkle path of the concrete instantiation.
-    /// It is MT-specific, since arity will affect the exact definition of the
+    /// Type to represent the leaf element of the concrete MT instantiation.
+    type LeafVar;
+
+    /// Type to represent the merkle path of the concrete MT instantiation.
+    /// It is MT-specific, e.g arity will affect the exact definition of the
     /// Merkle path.
     type MerklePathVar;
 
@@ -59,7 +53,7 @@ where
         &mut self,
         pos: M::Index,
         elem: M::Element,
-    ) -> Result<LeafVar, CircuitError>;
+    ) -> Result<Self::LeafVar, CircuitError>;
 
     /// Allocate a variable for the membership proof.
     fn create_membership_proof_variable(
@@ -74,7 +68,7 @@ where
     /// return `BoolVar` indicating the correctness of its membership proof
     fn is_member(
         &mut self,
-        elem: LeafVar,
+        elem: Self::LeafVar,
         merkle_proof: Self::MerklePathVar,
         merkle_root: Variable,
     ) -> Result<BoolVar, CircuitError>;
@@ -83,7 +77,7 @@ where
     /// `expected_merkle_root`.
     fn enforce_merkle_proof(
         &mut self,
-        elem: LeafVar,
+        elem: Self::LeafVar,
         merkle_proof: Self::MerklePathVar,
         expected_merkle_root: Variable,
     ) -> Result<(), CircuitError>;
--- a/primitives/src/circuit/merkle_tree/mod.rs
+++ b/primitives/src/circuit/merkle_tree/mod.rs
@@ -1,0 +1,72 @@
+// Copyright (c) 2022 Espresso Systems (espressosys.com)
+// This file is part of the Jellyfish library.
+
+// You should have received a copy of the MIT License
+// along with the Jellyfish library. If not, see <https://mit-license.org/>.
+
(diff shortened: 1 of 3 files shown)


primitives/src/circuit/merkle_tree/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+/// let leaf_var = circuit.create_leaf_variable(Fq::from(2_u64), Fq::from(100_u64)).unwrap();
+/// let path_vars = circuit.create_membership_proof_variable(&proof).unwrap();
+/// let root_var = circuit.create_root_variable(expected_root).unwrap();
+/// circuit.enforce_merkle_proof(leaf_var, path_vars, root_var).unwrap();
+/// assert!(circuit.check_circuit_satisfiability(&[]).is_ok());
+/// ```
+pub trait MerkleTreeGadget<F, M>
+where
+    F: PrimeField,
+    M: MerkleTreeScheme,
```
alxiong, line 52: why does this trait have to be generic over both `F` and `M`?

primitives/src/circuit/merkle_tree/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+{
+    /// Type to represent the merkle path of the concrete instantiation.
+    /// It is MT-specific, since arity will affect the exact definition of the
+    /// Merkle path.
+    type MerklePathVar;
+
+    /// Allocate a variable for the leaf element.
+    fn create_leaf_variable(&mut self, pos: F, elem: M::Element) -> Result<LeafVar, CircuitError>;
```
alxiong, line 60:
```suggestion
    fn create_leaf_variable(&mut self, pos: M::Index, elem: M::Element) -> Result<LeafVar, CircuitError>;
```

primitives/src/circuit/merkle_tree/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+
+/// Circuit variable for a leaf element.
+#[derive(Debug, Clone)]
+pub struct LeafVar {
+    /// Position of the leaf element in the MT. Serves as UID.
+    pub uid: Variable,
+    /// The value of the leaf element.
+    pub elem: Variable,
+}
+
```
alxiong, line 23: Do you think we should make this an associated type of `trait MerkleTreeGadget` instead, just like `MerklePathVar`?
The main rationale being that some MT's leaf (e.g. SMT) doesn't have to care about or even have `uid`. So it may suggest leaving the concrete instantiation to specific gadgets.

alxiong: on this note, I noticed our current [`RescueSMT`'s `digest_leaf`](https://github.com/EspressoSystems/jellyfish/blob/main/primitives/src/merkle_tree/prelude.rs#L55) is actually hashing `[0, pos, elem]`, which I guess doesn't hurt, but when we treat it as a set (as we do when using SMT), we don't need that `pos`, no? @mrain

primitives/src/circuit/merkle_tree/mod.rs
tessico: Your suggestion below is exactly the reason why it needs F right now. `RescueMerkleTree` defines the Index as `u64`, and that prevents an implementation `MerkleTreeGadget` for this specific MT.

I originally didn't want to touch the definition of `RescueMerkleTree`, but indeed it will make the interface even cleaner. We anyway cast the u64 index to `F` when using it (e.g. for hashing a leaf).

primitives/src/circuit/merkle_tree/mod.rs
mrain: No. In case of set, `elem` is of type `()`. Information is stored in `pos` actually.

alxiong: oh, right!

my original comment on making it an associated type might still be worth considering.

primitives/src/circuit/merkle_tree/mod.rs
tessico: Done

tessico:
> I wonder if we should name rescue_merkle_tree.rs as append_only.rs for consistency with their non-circuit implementations?

Yes, but after https://github.com/EspressoSystems/jellyfish/issues/142 lands. Currently the Rescue hash is "hardcoded" in the gadget implementation, so I think for now we should keep the name as is.
Later we can have `append_only.rs` and then instantiate a concrete one in the prelude of the circuit, like the native version does.

primitives/src/circuit/merkle_tree/mod.rs
tessico: Yes, good suggestion. Done in https://github.com/EspressoSystems/jellyfish/pull/158/commits/7288912d5ed7b6bd92b4e4e4ac9a8537cdf12b3c.

alxiong (approved):
LGTM!

mrain (commented):
LGTM

mrain (approved):

PR #167, opened 2022-12-19
Digest-generic MerkleTreeGadget

tessico:
## Description

Making the circuit equivalent of MerkleTree generic over a digest algorithm is more tricky, because both the `MerkleTreeGadget` and `DigestAlgorithmGadget` need to be implemented for `PlonkCircuit`.

I'm proposing two designs here. One is to add it as a generic parameter to `MerkleTreeGadget`, see [the last commit](https://github.com/EspressoSystems/jellyfish/pull/167/commits/4e3a09c80fac905b26f2632f1d4e33ece9bd3398). The other option is to add `DigestAlgorithmGadget` as an associated type on the `MerkleTreeGadget` trait, see [here](https://github.com/EspressoSystems/jellyfish/pull/167/commits/371e87470b2261ee1ea459d9f7e87a5d634d0397). Both options work, it's just a matter of convenience and ergonomics. Please let me know your opinions.

(Note that initially I thought that the associated type would not work in the case where in the future we need to provide multiple implementations of the `MerkleTreeGadget` trait. But since the trait already is generic over `MerkleTreeScheme`, which, implicitly, will be using the counterpart of `DigestAlgorithmGadget`, it should still be possible to provide multiple impl even with `DigestAlgorithmGadget` being an associated type. See [the rust book](https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#specifying-placeholder-types-in-trait-definitions-with-associated-types) for more explanation on the difference between the two.)

closes: #142 

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/primitives/src/circuit/merkle_tree/mod.rs
+++ b/primitives/src/circuit/merkle_tree/mod.rs
@@ -60,10 +60,10 @@ use super::rescue::RescueNativeGadget;
 /// .unwrap();
 /// assert!(circuit.check_circuit_satisfiability(&[]).is_ok());
 /// ```
-pub trait MerkleTreeGadget<M, F>
+pub trait MerkleTreeGadget<M>
 where
     M: MerkleTreeScheme,
-    F: PrimeField,
+    M::NodeValue: PrimeField,
 {
     /// Type to represent the merkle proof of the concrete MT instantiation.
     /// It is MT-specific, e.g arity will affect the exact definition of the
@@ -71,7 +71,7 @@ where
     type MembershipProofVar;
 
     /// Gadget for the digest algorithm.
-    type DigestGadget: DigestAlgorithmGadget<F>;
+    type DigestGadget: DigestAlgorithmGadget<M::NodeValue>;
 
     /// Allocate a variable for the membership proof.
     fn create_membership_proof_variable(
@@ -153,10 +153,10 @@ where
 /// .unwrap();
 /// assert!(circuit.check_circuit_satisfiability(&[]).is_ok());
 /// ```
-pub trait UniversalMerkleTreeGadget<M, F>: MerkleTreeGadget<M, F>
+pub trait UniversalMerkleTreeGadget<M>: MerkleTreeGadget<M>
 where
     M: UniversalMerkleTreeScheme,
-    F: PrimeField,
+    M::NodeValue: PrimeField,
 {
     /// Type to represent the merkle non-membership proof of the concrete MT
     /// instantiation. It is MT-specific, e.g arity will affect the exact
--- a/primitives/src/circuit/merkle_tree/mod.rs
+++ b/primitives/src/circuit/merkle_tree/mod.rs
@@ -274,7 +274,7 @@ impl<F: RescueParameter> DigestAlgorithmGadget<F> for RescueDigestGadget {
 
     fn digest_leaf(
         circuit: &mut PlonkCircuit<F>,
-        pos: usize,
+        pos: Variable,
         elem: Variable,
     ) -> Result<Variable, CircuitError> {
         let zero = circuit.zero();
--- a/primitives/src/circuit/merkle_tree/mod.rs
+++ b/primitives/src/circuit/merkle_tree/mod.rs
@@ -10,11 +10,14 @@ use crate::{
     merkle_tree::{MerkleTreeScheme, UniversalMerkleTreeScheme},
     rescue::RescueParameter,
 };
+use ark_ff::PrimeField;
 use jf_relation::{errors::CircuitError, BoolVar, Circuit, PlonkCircuit, Variable};
 
 mod rescue_merkle_tree;
 mod sparse_merkle_tree;
 use ark_std::vec::Vec;
+
+use super::rescue::RescueNativeGadget;
 
 /// Gadget for a Merkle tree
 ///
@@ -57,14 +60,18 @@ use ark_std::vec::Vec;
 /// .unwrap();
 /// assert!(circuit.check_circuit_satisfiability(&[]).is_ok());
 /// ```
-pub trait MerkleTreeGadget<M>
+pub trait MerkleTreeGadget<M, F>
 where
     M: MerkleTreeScheme,
+    F: PrimeField,
 {
     /// Type to represent the merkle proof of the concrete MT instantiation.
     /// It is MT-specific, e.g arity will affect the exact definition of the
     /// underlying Merkle path.
     type MembershipProofVar;
+
+    /// Gadget for the digest algorithm.
+    type DigestGadget: DigestAlgorithmGadget<F>;
 
     /// Allocate a variable for the membership proof.
     fn create_membership_proof_variable(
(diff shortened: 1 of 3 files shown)


primitives/src/circuit/merkle_tree/mod.rs
```diff
@@ -36,9 +36,10 @@ mod rescue_merkle_tree;
 /// circuit.enforce_merkle_proof(leaf_var, path_vars, root_var).unwrap();
 /// assert!(circuit.check_circuit_satisfiability(&[]).is_ok());
 /// ```
-pub trait MerkleTreeGadget<M>
+pub trait MerkleTreeGadget<M, D>
 where
     M: MerkleTreeScheme,
+    D: DigestAlgorithmGadget<M> + ?Sized,
```
mrain, line 42: Why does it have to be `?Sized`?

mrain (approved):
Generally LGTM.

primitives/src/circuit/merkle_tree/mod.rs
alxiong: Usually it's because we use it in dynamic trait objects somewhere

primitives/src/circuit/merkle_tree/rescue_merkle_tree.rs
```diff
@@ -41,7 +41,19 @@ pub struct StandardLeafVar {
     pub elem: Variable,
 }
 
-impl<F> MerkleTreeGadget<RescueMerkleTree<F>> for PlonkCircuit<F>
+impl<F: RescueParameter> DigestAlgorithmGadget<RescueMerkleTree<F>> for PlonkCircuit<F> {
+    fn digest(&mut self, data: &[Variable]) -> Result<Variable, CircuitError> {
```
alxiong, line 45: I'm actually little unsure.
Ideally, shouldn't `DigestAlgorithmGadget` only implemented for `MerkleTreeGadget`? We don't need these `digest_leaf` gadget exposing to the general `PlonkCircuit` IMO.

this may solve our `dyn DigestAlgorithmGadget` issue at the same time.

tessico: @alxiong Since `MerkleTreeGadget` is a trait, that becomes a little problematic. You either need to use `dyn MerkleTreeGadget` here instead, or do something like:
```
impl<T> DigestAlgorithmGadget for T
where
    T: MerkleTreeGadget
```

which I'm not even sure is possible, since we need to pass in the parameter `F` somehow like `impl<T, F>` (because we want to restrict `T: MerkleTreeGadget<RescueMerkleTree<F>>`), but the trait `DigestAlgorithmGadget` doesn't itself restrict `F`, and we'd run into errors.
By all means please drop a few lines/PR if you have an alternative of how to implement `DigestAlgorithmGadget` only for `MerkleTreeGadget` and not entire `PlonkCircuit` in this design.

But before you do that, please first checkout https://github.com/EspressoSystems/jellyfish/pull/167/commits/371e87470b2261ee1ea459d9f7e87a5d634d0397, which has an alternative design, and actually avoids the issue you mention altogether. (note that the concrete location of the structs/impls there is up for discussion, I might move them later to where `RescueNativeGadget` is defined).

alxiong: I kind of like your alternative solution, particularly we don't have the `dyn` issue, also compliant with my intuition that `DigestGadget` should be associated/part of/used under the context of `MerkleTreeGadget`.

tessico:
As per [pdiscussion](https://github.com/EspressoSystems/jellyfish/commit/371e87470b2261ee1ea459d9f7e87a5d634d0397) on the first commit, I've gone back to the initial solution by reverting the second commit.

primitives/src/circuit/merkle_tree/rescue_merkle_tree.rs
```diff
@@ -166,7 +184,7 @@ pub struct Rescue3AryNodeVar {
     pub is_right_child: BoolVar,
 }
 
-trait MerkleTreeHelperGadget<F: RescueParameter> {
+trait MerkleTreeHelperGadget<F: RescueParameter>: MerkleTreeGadget<RescueMerkleTree<F>, F> {
```
alxiong, line 187: this is subtyping relationship is slightly strange. 

What do you think of `MerkleTreeHelperGadget<F: RescueParameter, D: DigestAlgorithmGadget<F>>`?

tessico: Upon review against [the other PR](https://github.com/EspressoSystems/jellyfish/pull/172/), I've decided to wait until that non-membership proof PR is merged, as the trait definitions will change there anyway. This PR is smaller and can be adapted more easily.

tessico: I've added a `blocked` label for clarification. Don't wanna duplicate the work to refactor traits here twice.

primitives/src/circuit/merkle_tree/mod.rs
```diff
@@ -240,3 +248,36 @@ pub struct Merkle3AryMembershipProofVar {
...
+pub struct RescueDigestGadget {}
+
+impl<F: RescueParameter> DigestAlgorithmGadget<F> for RescueDigestGadget {
+    fn digest(circuit: &mut PlonkCircuit<F>, data: &[Variable]) -> Result<Variable, CircuitError> {
+        Ok(RescueNativeGadget::<F>::rescue_sponge_no_padding(circuit, data, 1)?[0])
+    }
+
+    fn digest_leaf(
+        circuit: &mut PlonkCircuit<F>,
+        pos: usize,
```
mrain, line 277: Should this `pos` be of type `Variable`? Although that's `usize`.

primitives/src/circuit/merkle_tree/mod.rs
```diff
@@ -57,15 +60,19 @@ use ark_std::vec::Vec;
 /// .unwrap();
 /// assert!(circuit.check_circuit_satisfiability(&[]).is_ok());
 /// ```
-pub trait MerkleTreeGadget<M>
+pub trait MerkleTreeGadget<M, F>
```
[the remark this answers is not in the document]
mrain: I tried the following and it's working:

```
pub trait MerkleTreeGadget<M>
where
    M: MerkleTreeScheme,
    M::NodeValue: PrimeField,
{
    ...
    /// Gadget for the digest algorithm.
    type DigestGadget: DigestAlgorithmGadget<M::NodeValue>;
```

mrain (changes requested):
Excellent work! Will approve if comments are addressed.

primitives/src/circuit/merkle_tree/mod.rs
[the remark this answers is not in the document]
tessico: Great suggestion @mrain! Done in https://github.com/EspressoSystems/jellyfish/pull/167/commits/abc72a7730547d893c7071678b67f581a286dd05

primitives/src/circuit/merkle_tree/mod.rs
tessico: Good catch. Done in https://github.com/EspressoSystems/jellyfish/pull/167/commits/d3560a6b9cf692989587fc43f39f2dbe4241ddf1

mrain (approved):
LGTM

Issue #168, opened 2022-12-19
Trait definition for CRHF and PRF


alxiong:
As part of the effort for stablizing APIs for 0.2.0 release, we plan to define traits for `CRHF` and `PRF` and conform our implementations under them.

chancharles92:
What about commitments?

Also, another part is to improve/complete the refactoring of SNARK/PCS/VRF/Signature traits if incomplete.

Issue #159, opened 2022-12-13
Unify transcript implementation


mrain:
Right now we have duplicated implementation of transcript module.

One lies in jf-plonk as some of its components depend on plonk implementation.
One lies in jf-primitives as pcs relies on it.

We should refactor the code to unify them.

PR #169, opened 2022-12-20
Created CRHF trait and conform RescueCRH impl

alxiong:
## Description

Part of: #168 

Changes include:
- add `trait CRHF`
- Provide `struct FixedLengthRescueCRHF` and `struct VariableLengthRescueCRHF` conforming to the trait.
- add `trait CommitmentScheme`
- Provide `struct FixedLengthRescueCRHF`

NOTE:
to hack around some nightly-only feature on using const generic in an expression, we have a temporarily excessive generic parameter on `FixedLengthRescueCRHF<F, INPUT_LEN, INPUT_LEN_PLUS_ONE>` which will be fixed in the future.

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/primitives/src/circuit/merkle_tree/mod.rs
+++ b/primitives/src/circuit/merkle_tree/mod.rs
@@ -30,7 +30,7 @@ mod rescue_merkle_tree;
 /// let (_, proof) = mt.lookup(2).expect_ok().unwrap();
 ///
 /// // Circuit computation with a MT
-/// let leaf_var = circuit.create_leaf_variable(Fq::from(2_u64), Fq::from(100_u64)).unwrap();
+/// let leaf_var = circuit.create_leaf_variable(2_u64, Fq::from(100_u64)).unwrap();
 /// let path_vars = circuit.create_membership_proof_variable(&proof).unwrap();
 /// let root_var = circuit.create_root_variable(expected_root).unwrap();
 /// circuit.enforce_merkle_proof(leaf_var, path_vars, root_var).unwrap();
--- a/primitives/src/crhf.rs
+++ b/primitives/src/crhf.rs
@@ -29,6 +29,12 @@ pub trait CRHF {
 
     /// evaluate inputs and return hash output
     fn evaluate<T: Borrow<Self::Input>>(input: T) -> Result<Self::Output, PrimitivesError>;
+
+    /// same as [`evaluate`] with caller assurance that `input` won't cause any
+    /// error, mostly for internal use, invoke with caution!
+    fn evaluate_safe<T: Borrow<Self::Input>>(input: T) -> Self::Output {
+        Self::evaluate(input).expect("Should NOT panic")
+    }
 }
 
 #[derive(Debug, Clone)]
@@ -45,12 +51,17 @@ impl<F: RescueParameter, const INPUT_LEN: usize, const OUTPUT_LEN: usize> CRHF
 {
     type Input = [F; INPUT_LEN];
     type Output = [F; OUTPUT_LEN];
+
+    /// ## Padding
+    /// if `input` length is not a multiple of `CRHF_RATE`, then it will be
+    /// padded. By default, we use "zero padding"-style where as many "0" as
+    /// required are added.
     fn evaluate<T: Borrow<Self::Input>>(input: T) -> Result<Self::Output, PrimitivesError> {
         let mut output = [F::zero(); OUTPUT_LEN];
 
         let res = match INPUT_LEN % CRHF_RATE {
             0 => RescueCRHF::<F>::sponge_no_padding(input.borrow(), OUTPUT_LEN)?,
-            _ => RescueCRHF::<F>::sponge_with_padding(input.borrow(), OUTPUT_LEN),
+            _ => RescueCRHF::<F>::sponge_with_zero_padding(input.borrow(), OUTPUT_LEN),
         };
         if res.len() != OUTPUT_LEN {
             return Err(PrimitivesError::InternalError(
@@ -71,9 +82,14 @@ impl<F: RescueParameter, const OUTPUT_LEN: usize> CRHF for VariableLengthRescueC
     type Input = Vec<F>;
     type Output = [F; OUTPUT_LEN];
 
+    /// ## Padding
+    /// if `input` length is not a multiple of `CRHF_RATE`, then it will be
+    /// padded. By default, we use "bit padding"-style where "1" is always
+    /// appended, then as many "0" as required are added for the overall
+    /// length to be a multiple of `CRHF_RATE`.
     fn evaluate<T: Borrow<Self::Input>>(input: T) -> Result<Self::Output, PrimitivesError> {
         let mut output = [F::zero(); OUTPUT_LEN];
-        let res = RescueCRHF::<F>::sponge_with_padding(input.borrow(), OUTPUT_LEN);
+        let res = RescueCRHF::<F>::sponge_with_bit_padding(input.borrow(), OUTPUT_LEN);
         if res.len() != OUTPUT_LEN {
             return Err(PrimitivesError::InternalError(
                 "Unexpected rescue sponge return length".to_string(),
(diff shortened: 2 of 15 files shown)


primitives/src/crhf.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    /// evaluate inputs and return hash output
+    fn evaluate<T: Borrow<Self::Input>>(
+        param: &Self::Parameters,
+        input: T,
+    ) -> Result<Self::Output, PrimitivesError>;
+}
+
+#[derive(Debug, Clone)]
+/// A rescue-sponge-based CRHF with fixed-input size (if not multiple of 3 will
```
tessico, line 50:
```suggestion
/// A rescue-sponge-based CRHF with fixed-input size (if not multiple of 3 will get
```

primitives/src/crhf.rs
```diff
@@ -0,0 +1,117 @@
...
+    // FIXME: (alex) wait until arkwork 0.4.0 to add the following:
+    // + Default + CanonicalSerialize + CanonicalDeserialize;
+    // right now, const-generic are not supported yet.
+    type Output: Clone + PartialEq + Eq + Hash + Debug;
+
+    /// Public parameters of the CRHF (if any)
+    type Parameters: Clone + Debug + Sync + Send;
+
+    /// picking the concrete CRHF from the parameterized hash family
+    fn setup<R: RngCore + CryptoRng>(rng: &mut R) -> Result<Self::Parameters, PrimitivesError>;
```
tessico, line 40: I wouldn't define it if it's not used at all in the current impl.
If we were to take arkworks traits and implement these, then sure, we need to have the setup, but since (for now at least) we're defining our own trait, I think it's ok to leave it out, especially that (potentially) adding it back once we move to arkwork's traits is going to be trivial

tessico (commented):
What about replacing all the current calls to `RescueCRHF` with `(Fixed/Variable)LengthRescueCRHF`?

primitives/src/crhf.rs
alxiong: good point! will remove for now.

alxiong:
> What about replacing all the current calls to RescueCRHF with (Fixed/Variable)LengthRescueCRHF?

Yes I have considered this, the problem is, currently `RescueCRHF::sponge_with_padding() -> Vec<F>` always succeed whereas `sponge_no_padding() -> Result<>` could fail.
Some functions that internally call padded version API also have a function signature that won't fail. If we switch to `VariableLengthRescueCRHF::evaluate() -> Result<Self::Output>`, since we need to conform with `trait CRH` then we will have to return a `Result` and do `unwrap()`.

A possible middle ground is we declare another API: `VariableLengthRescueCRHF::evaluate_safe() -> Self::Output` which internal call `evaluate()` and `unwrap()` once cuz caller know it won't panic.

primitives/src/rescue/mod.rs
```diff
@@ -30,7 +30,7 @@ use ark_std::{vec, vec::Vec};
 /// The state size of rescue hash.
 pub const STATE_SIZE: usize = 4;
 /// The rate of the sponge used in RescueCRHF.
-pub const CRHF_RATE: usize = 3;
+pub(crate) const CRHF_RATE: usize = 3;
```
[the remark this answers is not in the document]
alxiong: currently still usage in `transcript` in jf-plonk that utilize this constant [here](https://github.com/EspressoSystems/jellyfish/blob/crhf-trait/plonk/src/transcript/rescue.rs#L180), I think we might be able make it consistent after #159 ?

I will think more meanwhile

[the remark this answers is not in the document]
mrain: Suddenly I feel like we should export these 2 constants. Perhaps we could make them trait/struct dependent?

alxiong:
Update: during refactoring of @tessico's suggestion 

>    What about replacing all the current calls to RescueCRHF with (Fixed/Variable)LengthRescueCRHF?

I need to use `FixedLengthRescueCRHF::<F, INPUT_LEN, 1>::evaluate()` inside our `Commitment`, but currently the length is just a field (variable) instead of a constant value (meaning not knowable at compile time), thus compiler complains.

Therefore this forces me try the following: 
```rust
pub struct FixedLengthRescueCommitment<F: RescueParameter, const INPUT_LEN: usize>(PhantomData<F>);
```
meanwhile, I might as well just declare a `trait CommitmentScheme` in this PR.

primitives/src/rescue/mod.rs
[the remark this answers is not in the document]
alxiong: I tend to disagree. I think it's intuitive that `CRHF_RATE` is a internal detail of CRHF, thus try to hide from public is sensible. whereas `STATE_SIZE` is a parameter of our Rescue permutation instance, which like `A` and `MDS` should be public. 

I think the current visibility makes sense to me.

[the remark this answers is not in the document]
mrain: But what if the user need to know the "rate of compression"?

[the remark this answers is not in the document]
alxiong: why would the user want to know that? I couldn't think of a use case for now.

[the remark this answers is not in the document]
alxiong: changed back in [7b77e79](https://github.com/EspressoSystems/jellyfish/pull/169/commits/7b77e79ce23f4d5439a5815573ff8646de4b8d69)

primitives/src/circuit/merkle_tree/mod.rs
```diff
@@ -30,7 +30,7 @@ mod rescue_merkle_tree;
 /// let (_, proof) = mt.lookup(2).expect_ok().unwrap();
 ///
 /// // Circuit computation with a MT
-/// let leaf_var = circuit.create_leaf_variable(Fq::from(2_u64), Fq::from(100_u64)).unwrap();
+/// let leaf_var = circuit.create_leaf_variable(2_u64, Fq::from(100_u64)).unwrap();
```
tessico, line 33: Thanks for the fix!

primitives/src/rescue/sponge.rs
```diff
@@ -39,20 +40,36 @@ impl<F: RescueParameter> RescueCRHF<F> {
     /// Sponge hashing based on rescue permutation for RATE 3. It allows
     /// unrestricted variable length input and returns a vector of
     /// `num_outputs` elements.
-    pub fn sponge_with_padding(input: &[F], num_outputs: usize) -> Vec<F> {
-        // Pad input as follows: append a One, then pad with 0 until length is multiple
-        // of RATE
+    ///
+    /// we use ["bit padding"-style][padding] where "1" is always appended, then
```
tessico, line 44: Great!

primitives/src/signatures/schnorr.rs
```diff
@@ -299,7 +300,7 @@ where
         let mut msg_input = vec![instance_description, fr_to_fq::<F, P>(&self.sk.0)];
         msg_input.extend(msg.iter());
 
-        let r = fq_to_fr::<F, P>(&RescueCRHF::sponge_with_padding(&msg_input, 1)[0]);
+        let r = fq_to_fr::<F, P>(&VariableLengthRescueCRHF::<F, 1>::evaluate_safe(&msg_input)[0]);
```
tessico, line 303: Now that I see this coded up, I prefer removing `evaluate_safe` function, and instead simply have the caller `unwrap` . The behaviour will be exactly the same - but the interface will be cleaner. What do others think?

tessico (commented):
Looks good, I left some final minor comments.

primitives/src/signatures/schnorr.rs
alxiong: I can see where you are coming from. it's more explicit about the function contract (what it promises) instead of some hidden, less obvious choice of potential panicking.

I will change to `.unwarp()` as suggested

tessico (approved):

PR #172, opened 2022-12-21
Non-membership proof for SMT

### LINKED ISSUE #161: [Constraint] SMT-based Non-membership Proof
Sparse MT-based non-membership in the circuit.

tessico:
## Description

closes #161 

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/primitives/src/circuit/merkle_tree/mod.rs
+++ b/primitives/src/circuit/merkle_tree/mod.rs
@@ -10,6 +10,7 @@ use crate::merkle_tree::MerkleTreeScheme;
 use jf_relation::{errors::CircuitError, BoolVar, Variable};
 
 mod rescue_merkle_tree;
+mod sparse_merkle_tree;
 
 /// Gadget for a Merkle tree
 ///
--- a/primitives/src/circuit/merkle_tree/mod.rs
+++ b/primitives/src/circuit/merkle_tree/mod.rs
@@ -32,13 +32,7 @@ use ark_std::vec::Vec;
 /// let (_, proof) = mt.lookup(2).expect_ok().unwrap();
 ///
 /// // Circuit computation with a MT
-/// let leaf_var =
-///     MerkleTreeGadget::<RescueMerkleTree<Fq>>::create_leaf_variable(
-///         &mut circuit,
-///         2_u64,
-///         Fq::from(100_u64)
-///     )
-///     .unwrap();
+/// let leaf_idx = circuit.create_variable(2_u64.into()).unwrap();
 /// let proof_var =
 ///     MerkleTreeGadget::<RescueMerkleTree<Fq>>::create_membership_proof_variable(
 ///         &mut circuit,
@@ -53,7 +47,7 @@ use ark_std::vec::Vec;
 ///     .unwrap();
 /// MerkleTreeGadget::<RescueMerkleTree<Fq>>::enforce_membership_proof(
 ///     &mut circuit,
-///     leaf_var,
+///     leaf_idx,
 ///     proof_var,
 ///     root_var
 /// )
@@ -64,20 +58,10 @@ pub trait MerkleTreeGadget<M>
 where
     M: MerkleTreeScheme,
 {
-    /// Type to represent the leaf element of the concrete MT instantiation.
-    type LeafVar;
-
     /// Type to represent the merkle proof of the concrete MT instantiation.
     /// It is MT-specific, e.g arity will affect the exact definition of the
     /// underlying Merkle path.
     type MembershipProofVar;
-
-    /// Allocate a variable for the leaf element.
-    fn create_leaf_variable(
-        &mut self,
-        pos: M::Index,
-        elem: M::Element,
-    ) -> Result<Self::LeafVar, CircuitError>;
 
     /// Allocate a variable for the membership proof.
     fn create_membership_proof_variable(
@@ -92,7 +76,7 @@ where
     /// return `BoolVar` indicating the correctness of its membership proof
     fn is_member(
         &mut self,
-        elem_var: Self::LeafVar,
+        elem_var: Variable,
         proof_var: Self::MembershipProofVar,
         root_var: Variable,
     ) -> Result<BoolVar, CircuitError>;
@@ -101,7 +85,7 @@ where
     /// `expected_root_var`.
     fn enforce_membership_proof(
         &mut self,
-        elem_var: Self::LeafVar,
+        elem_var: Variable,
         proof_var: Self::MembershipProofVar,
         expected_root_var: Variable,
     ) -> Result<(), CircuitError>;
@@ -170,10 +154,16 @@ where
     /// definition of the underlying Merkle path.
     type NonMembershipProofVar;
 
+    /// Allocate a variable for the membership proof.
+    fn create_non_membership_proof_variable(
+        &mut self,
+        membership_proof: &M::MembershipProof,
+    ) -> Result<Self::NonMembershipProofVar, CircuitError>;
+
     /// checking non-membership proof
     fn is_non_member(
         &mut self,
-        elem_var: Self::LeafVar,
+        elem_var: Variable,
         proof_var: Self::NonMembershipProofVar,
         root_var: Variable,
     ) -> Result<BoolVar, CircuitError>;
@@ -182,7 +172,7 @@ where
     /// `expected_root_var`.
     fn enforce_non_membership_proof(
         &mut self,
-        empty_elem_var: Self::LeafVar,
+        empty_elem_var: Variable,
         proof_var: Self::NonMembershipProofVar,
         expected_root_var: Variable,
     ) -> Result<(), CircuitError>;
@@ -231,7 +221,7 @@ where
     ///   tree.
     fn compute_merkle_root(
         &mut self,
-        elem_var: LeafVar,
+        elem_var: Variable,
         proof_var: &Self::MembershipProofVar,
     ) -> Result<Variable, CircuitError>;
 }
@@ -274,4 +264,5 @@ pub struct Merkle3AryNonMembershipProofVar {
 #[derive(Debug, Clone)]
 pub struct Merkle3AryMembershipProofVar {
     nodes: Vec<Merkle3AryNodeVar>,
+    leaf: Variable,
 }
--- a/primitives/src/circuit/merkle_tree/mod.rs
+++ b/primitives/src/circuit/merkle_tree/mod.rs
@@ -6,7 +6,7 @@
 
 //! Trait definitions for a Merkle tree gadget.
 
-use crate::merkle_tree::MerkleTreeScheme;
+use crate::merkle_tree::{MerkleTreeScheme, UniversalMerkleTreeScheme};
 use jf_relation::{errors::CircuitError, BoolVar, Variable};
 
 mod rescue_merkle_tree;
@@ -145,7 +145,7 @@ where
 /// ```
 pub trait UniversalMerkleTreeGadget<M>: MerkleTreeGadget<M>
 where
-    M: MerkleTreeScheme,
+    M: UniversalMerkleTreeScheme,
 {
     /// Type to represent the merkle non-membership proof of the concrete MT
     /// instantiation. It is MT-specific, e.g arity will affect the exact
@@ -155,7 +155,7 @@ where
     /// Allocate a variable for the membership proof.
     fn create_non_membership_proof_variable(
         &mut self,
-        membership_proof: &M::MembershipProof,
+        membership_proof: &M::NonMembershipProof,
     ) -> Result<Self::NonMembershipProofVar, CircuitError>;
 
     /// checking non-membership proof
--- a/primitives/src/circuit/merkle_tree/mod.rs
+++ b/primitives/src/circuit/merkle_tree/mod.rs
@@ -92,18 +92,18 @@ where
     /// return `BoolVar` indicating the correctness of its membership proof
     fn is_member(
         &mut self,
-        elem: Self::LeafVar,
-        merkle_proof: Self::MembershipProofVar,
-        merkle_root: Variable,
+        elem_var: Self::LeafVar,
+        proof_var: Self::MembershipProofVar,
+        root_var: Variable,
     ) -> Result<BoolVar, CircuitError>;
 
-    /// Enforce correct `merkle_proof` for the `elem` against
-    /// `expected_merkle_root`.
+    /// Enforce correct `proof_var` for the `elem` against
+    /// `expected_root_var`.
     fn enforce_membership_proof(
         &mut self,
-        elem: Self::LeafVar,
-        merkle_proof: Self::MembershipProofVar,
-        expected_merkle_root: Variable,
+        elem_var: Self::LeafVar,
+        proof_var: Self::MembershipProofVar,
+        expected_root_var: Variable,
     ) -> Result<(), CircuitError>;
 }
 
@@ -165,21 +165,26 @@ pub trait UniversalMerkleTreeGadget<M>: MerkleTreeGadget<M>
 where
     M: MerkleTreeScheme,
 {
+    /// Type to represent the merkle non-membership proof of the concrete MT
+    /// instantiation. It is MT-specific, e.g arity will affect the exact
+    /// definition of the underlying Merkle path.
+    type NonMembershipProofVar;
+
     /// checking non-membership proof
     fn is_non_member(
         &mut self,
-        elem: Self::LeafVar,
-        merkle_proof: Self::MembershipProofVar,
-        merkle_root: Variable,
+        elem_var: Self::LeafVar,
+        proof_var: Self::NonMembershipProofVar,
+        root_var: Variable,
     ) -> Result<BoolVar, CircuitError>;
 
-    /// Enforce correct `merkle_proof` for the empty leaf `elem` against
-    /// `expected_merkle_root`.
+    /// Enforce correct `proof_var` for the empty leaf `empty_elem_var` against
+    /// `expected_root_var`.
     fn enforce_non_membership_proof(
         &mut self,
-        empty_elem: Self::LeafVar,
-        merkle_proof: Self::MembershipProofVar,
-        expected_merkle_root: Variable,
+        empty_elem_var: Self::LeafVar,
+        proof_var: Self::NonMembershipProofVar,
+        expected_root_var: Variable,
     ) -> Result<(), CircuitError>;
 }
 
@@ -221,15 +226,15 @@ where
 
     /// Computes the merkle root based on some element placed at a leaf and a
     /// merkle path.
-    /// * `elem` - variables corresponding to the uid and the element value
+    /// * `elem_var` - variables corresponding to the uid and the element value
     ///   (e.g.: record commitment).
-    /// * `path_vars` - variables corresponding to the Merkle path.
+    /// * `proof_var` - variable corresponding to the Merkle proof.
     /// * `return` - variable corresponding to the root value of the Merkle
     ///   tree.
     fn compute_merkle_root(
         &mut self,
-        elem: LeafVar,
-        path_vars: &Self::MembershipProofVar,
+        elem_var: LeafVar,
+        proof_var: &Self::MembershipProofVar,
     ) -> Result<Variable, CircuitError>;
 }
 
@@ -292,12 +297,20 @@ pub(crate) struct MerklePathRepr<M: MerkleTreeScheme> {
     pub(crate) nodes: Vec<Merkle3AryNodeRepr<M>>,
 }
 
-/// Circuit variable for a Merkle proof of a 3-ary Merkle tree.
+/// Circuit variable for a Merkle non-membership proof of a 3-ary Merkle tree.
 /// Constains:
 /// * a list of node variables in the path,
 /// * a variable correseponsing to the position of the leaf element.
 #[derive(Debug, Clone)]
-pub struct Merkle3AryMembershipProofVar {
+pub struct Merkle3AryNonMembershipProofVar {
     nodes: Vec<Merkle3AryNodeVar>,
     pos: Variable,
+}
+
+/// Circuit variable for a Merkle proof of a 3-ary Merkle tree.
+/// Constains:
+/// * a list of node variables in the path,
+#[derive(Debug, Clone)]
+pub struct Merkle3AryMembershipProofVar {
+    nodes: Vec<Merkle3AryNodeVar>,
 }
--- a/primitives/src/circuit/merkle_tree/mod.rs
+++ b/primitives/src/circuit/merkle_tree/mod.rs
@@ -189,16 +189,17 @@ where
 {
     type MerklePathEncoding;
     type MembershipProofVar;
-    /// Produces an ordered list of variables based on the relative position of
-    /// a node and its siblings.
-    /// * `node` - node to be inserted in the final list.
+    /// Produces a list of circuit variables representing the ordered nodes,
+    /// based on the location of a `node` among its siblings, and otherwise
+    /// preserving the relative location of the siblings.
+    /// * `node` - node to be placed in the correct position
     /// * `sibling1` - first sibling
     /// * `sibling2` - second sibling
     /// * `node_is_left` - variable that is true if node is the leftmost one.
     /// * `node_is_right` -  variable that is true if node is the rightmost one.
     /// * `returns` - list of variables corresponding to the node and its
     ///   siblings in the correct order.
-    fn permute(
+    fn constrain_sibling_order(
         &mut self,
         node: Variable,
         sib1: Variable,
--- a/primitives/src/circuit/merkle_tree/mod.rs
+++ b/primitives/src/circuit/merkle_tree/mod.rs
@@ -32,10 +32,32 @@ use ark_std::vec::Vec;
 /// let (_, proof) = mt.lookup(2).expect_ok().unwrap();
 ///
 /// // Circuit computation with a MT
-/// let leaf_var = circuit.create_leaf_variable(2_u64, Fq::from(100_u64)).unwrap();
-/// let path_vars = circuit.create_membership_proof_variable(&proof).unwrap();
-/// let root_var = circuit.create_root_variable(expected_root).unwrap();
-/// circuit.enforce_merkle_proof(leaf_var, path_vars, root_var).unwrap();
+/// let leaf_var =
+///     <PlonkCircuit<Fq> as MerkleTreeGadget<RescueMerkleTree<Fq>>>::create_leaf_variable(
+///         &mut circuit,
+///         2_u64,
+///         Fq::from(100_u64)
+///     )
+///     .unwrap();
+/// let path_vars =
+///     <PlonkCircuit<Fq> as MerkleTreeGadget<RescueMerkleTree<Fq>>>::create_membership_proof_variable(
+///         &mut circuit,
+///         &proof
+///     )
+///     .unwrap();
+/// let root_var =
+///     <PlonkCircuit<Fq> as MerkleTreeGadget<RescueMerkleTree<Fq>>>::create_root_variable(
+///         &mut circuit,
+///         expected_root
+///     )
+///     .unwrap();
+/// <PlonkCircuit<Fq> as MerkleTreeGadget<RescueMerkleTree<Fq>>>::enforce_membership_proof(
+///     &mut circuit,
+///     leaf_var,
+///     path_vars,
+///     root_var
+/// )
+/// .unwrap();
 /// assert!(circuit.check_circuit_satisfiability(&[]).is_ok());
 /// ```
 pub trait MerkleTreeGadget<M>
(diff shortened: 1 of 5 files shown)


primitives/src/circuit/merkle_tree/mod.rs
```diff
@@ -75,10 +77,133 @@ where
...
+    fn compute_merkle_root(
+        &mut self,
+        elem: StandardLeafVar,
+        path_vars: &Self::MembershipProofVar,
+    ) -> Result<Variable, CircuitError>;
+}
+
+#[derive(Debug, Clone)]
+/// Circuit variable for a Merkle node.
+pub struct Rescue3AryNodeVar {
```
chancharles92, line 162: Is it better to call it `Merkle3AryNodeVar`?

primitives/src/circuit/merkle_tree/mod.rs
```diff
@@ -75,10 +77,133 @@ where
...
+    pub sibling2: Variable,
+    /// Boolean variable indicating whether the node is a left child.
+    pub is_left_child: BoolVar,
+    /// Boolean variable indicating whether the node is a right child.
+    pub is_right_child: BoolVar,
+}
+
+/// Circuit variable for a leaf element.
+#[derive(Debug, Clone)]
+pub struct StandardLeafVar {
```
chancharles92, line 175: Why not just call it `LeafVar`? Are there other types of LeafVar?

primitives/src/circuit/merkle_tree/sparse_merkle_tree.rs
```diff
@@ -0,0 +1,688 @@
...
+
+use super::{
+    MembershipProofBooleanEncoding, MerkleNodeBooleanEncoding, MerkleTreeGadget,
+    MerkleTreeHelperGadget, Rescue3AryNodeVar, SparseMerkleTreeGadget, StandardLeafVar,
+};
+
+#[derive(Debug, Clone)]
+/// Circuit variable for a Merkle authentication path for a Rescue-based, 3-ary
+/// Merkle tree.
+pub struct Rescue3AryMembershipProofVar {
```
chancharles92, line 37: This struct can be more general as it doesn't use any Rescue-related info inside. Is it better to call it `Merkle3AryMembershipProofVar`?

primitives/src/circuit/merkle_tree/sparse_merkle_tree.rs
```diff
@@ -0,0 +1,688 @@
...
+    fn constrain_membership_proof(
+        &mut self,
+        merkle_path: &MembershipProofBooleanEncoding<SparseMerkleTree<F>>,
+        pos: Index<F>,
+    ) -> Result<Rescue3AryMembershipProofVar, CircuitError> {
+        // Setup node variables
+        let nodes = merkle_path
+            .nodes
+            .clone()
+            .into_iter()
```
chancharles92, line 245: Is it different from `Merkle_path.nodes.iter().map..`? Do we have to move `node` here?

primitives/src/circuit/merkle_tree/sparse_merkle_tree.rs
```diff
@@ -0,0 +1,688 @@
...
+    }
+
+    fn compute_merkle_root(
+        &mut self,
+        elem: StandardLeafVar,
+        path_vars: &Rescue3AryMembershipProofVar,
+    ) -> Result<Variable, CircuitError> {
+        let zero_var = self.zero();
+
+        // leaf label = H(0, uid, arc)
```
chancharles92, line 278: H(0, uid , arc) -> H(0, uid, elem)

primitives/src/circuit/merkle_tree/mod.rs
```diff
@@ -75,10 +77,133 @@ where
...
         &mut self,
         elem: Self::LeafVar,
         merkle_proof: Self::MerklePathVar,
         expected_merkle_root: Variable,
     ) -> Result<(), CircuitError>;
 }
+
+/// Gadget for the sparse merkle tree
+pub trait SparseMerkleTreeGadget<M>: MerkleTreeGadget<M>
+where
```
alxiong, line 90: please consider adding some example (testable) rustdoc, similar to doc for `MerkleTreeGadget`.

primitives/src/circuit/merkle_tree/mod.rs
```diff
@@ -75,10 +77,133 @@ where
...
+    /// Produces an ordered list of variables based on the relative position of
+    /// a node and its siblings.
+    /// * `node` - node to be inserted in the final list.
+    /// * `sibling1` - first sibling
+    /// * `sibling2` - second sibling
+    /// * `node_is_left` - variable that is true if node is the leftmost one.
+    /// * `node_is_right` -  variable that is true if node is the rightmost one.
+    /// * `returns` - list of variables corresponding to the node and its
+    ///   siblings in the correct order.
+    fn permute(
```
alxiong, line 126: this sounds more like `sort/order/sequence`?
without reading the doc, I would think `permute(vec) -> permuted_vec`

primitives/src/circuit/merkle_tree/mod.rs
```diff
@@ -75,10 +77,133 @@ where
...
+        expected_merkle_root: Variable,
+    ) -> Result<(), CircuitError>;
+}
+
+pub(crate) trait MerkleTreeHelperGadget<M>
+where
+    M: MerkleTreeScheme,
+{
+    type MembershipProofBooleanEncoding;
+    type MembershipProofVar;
```
alxiong, line 116: add some documentation on these types? I have some trouble understanding (the semantic meaning of) `BooleanEncoding`, also the mixed usage of `merkle_path`, `merkle_proof` and `memebership_proof`.

primitives/src/circuit/merkle_tree/mod.rs
```diff
@@ -75,10 +77,133 @@ where
...
+    /// Ensure that the position of each node of the path is correctly encoded
+    /// Used for testing purposes.
+    /// * `merkle_path` - list of node of an authentication path
+    /// * `pos` - position of the missing leaf
+    /// * `returns` - list of variables corresponding to the authentication path
+    fn constrain_membership_proof(
+        &mut self,
+        merkle_path: &Self::MembershipProofBooleanEncoding,
+        pos: M::Index,
+    ) -> Result<Self::MembershipProofVar, CircuitError>;
```
alxiong, line 144: curious why isn't this called `create_membership_proof_var`? (during creation, we check some structural validity, which is common). 

the first interface comes to mind when I saw `constrain_membership_proof` is something like `verify(proof_var)-> Result<(), E>`

primitives/src/circuit/merkle_tree/rescue_merkle_tree.rs
```diff
@@ -80,66 +86,40 @@ where
...
-        MembershipProofBooleanEncoding {
-            nodes: nodes.to_vec(),
-        }
-    }
-}
-
-impl<F: RescueParameter> From<&MembershipProof<F>> for MembershipProofBooleanEncoding<F> {
+impl<F: RescueParameter> From<&MembershipProof<F>>
+    for MembershipProofBooleanEncoding<RescueMerkleTree<F>>
+{
```
alxiong, line 117: Maybe consider `impl<T: Borrow<MembershipProof>> From<T>` to accept conversion from both ownered and borrowed?

primitives/src/circuit/merkle_tree/rescue_merkle_tree.rs
```diff
@@ -80,66 +86,40 @@ where
...
-        let bool_val = self
-            .is_member(elem, merkle_proof, expected_merkle_root)
-            .unwrap();
+        let bool_val = <PlonkCircuit<F> as MerkleTreeGadget<RescueMerkleTree<F>>>::is_member(
+            self,
+            elem,
+            merkle_proof,
+            expected_merkle_root,
+        )
+        .unwrap();
```
alxiong, line 110: I don't see why we are using so many `.unwrap()` in non-test code. @tessico do you mind helping us remove all of them and propagate `CircuitError` properly?

primitives/src/merkle_tree/prelude.rs
```diff
@@ -64,4 +64,4 @@ impl<F: RescueParameter> DigestAlgorithm<F, F, F> for RescueHash<F> {
 }
 
 /// Example instantiation of a SparseMerkleTree indexed by I
-pub type RescueSparseMerkleTree<I, E, F> = UniversalMerkleTree<E, RescueHash<F>, I, U3, F>;
+pub type RescueSparseMerkleTree<I, F> = UniversalMerkleTree<F, RescueHash<F>, I, U3, F>;
```
mrain, line 67: Any reason for enforcing the element type to be a field element?

primitives/src/circuit/merkle_tree/mod.rs
```diff
@@ -75,10 +77,133 @@ where
...
+    fn is_non_member(
+        &mut self,
+        elem: Self::LeafVar,
+        merkle_proof: Self::MerklePathVar,
+        merkle_root: Variable,
+    ) -> Result<BoolVar, CircuitError>;
+
+    /// Enforce correct `merkle_proof` for the empty leaf `elem` against
+    /// `expected_merkle_root`.
+    fn enforce_non_membership_proof(
```
mrain, line 103: I think it might be better to have a consistent naming, by calling this `enforce_non_member`.

primitives/src/circuit/merkle_tree/mod.rs
mrain: Or `RescueMT3AryNodeVar`?

primitives/src/circuit/merkle_tree/mod.rs
alxiong: I like `enforce_non_membership_proof` better, it's clear what it's enforcing.

primitives/src/circuit/merkle_tree/mod.rs
```diff
@@ -75,10 +77,133 @@ where
...
+    fn enforce_membership_proof(
         &mut self,
         elem: Self::LeafVar,
         merkle_proof: Self::MerklePathVar,
         expected_merkle_root: Variable,
     ) -> Result<(), CircuitError>;
 }
+
+/// Gadget for the sparse merkle tree
+pub trait SparseMerkleTreeGadget<M>: MerkleTreeGadget<M>
```
mrain, line 89: Is it better to call it `UniversalMerkleTreeGadget`?
There was some discussion before, where we concluded that non_membership_proof is a feature of Universal Merkle tree.

primitives/src/circuit/merkle_tree/mod.rs
mrain: True. I agree that this is more clear.

primitives/src/circuit/merkle_tree/mod.rs
tessico: Ah sure, I think I either missed the discussion or it happened before my time :)

primitives/src/merkle_tree/prelude.rs
tessico: Actually no. Since we weren't using other types than field element in our code, it just made the initial version of this PR easier.

But I suppose one could ask the same question about [RescueMerkleTree](https://github.com/EspressoSystems/jellyfish/blob/4e017caf3091cee90019911ef330422f808e3f4c/primitives/src/merkle_tree/prelude.rs#L42).

In any case we should probable be consistent.

primitives/src/circuit/merkle_tree/mod.rs
tessico: Hmm good question!
I suppose the reason is that when [making the interface more ergonomic](https://github.com/EspressoSystems/jellyfish/pull/158/), I mostly focused on the public API, whereas the MTGadgetHelper was an internal detail, really, its naming left untouched.
The name `create_membership_proof_variable` is already taken for the public API.
I'm open to suggestions for further naming options, though in this case I find `constrain_membership_proof` quite accurate.

primitives/src/circuit/merkle_tree/mod.rs
tessico: Right, the path/proof terminology should definitely be reworked.

primitives/src/circuit/merkle_tree/mod.rs
tessico: I'd go with @chancharles92's suggestion here, since it's actually not Rescue specific. Thx!

primitives/src/circuit/merkle_tree/mod.rs
tessico: Done: https://github.com/EspressoSystems/jellyfish/pull/172/commits/4abcab9afcd6e88beae03449ff85b02f86820bf4

primitives/src/circuit/merkle_tree/mod.rs
tessico: Done: https://github.com/EspressoSystems/jellyfish/pull/172/commits/86b0eb179b563747ad6099049e580056f02f5c61

primitives/src/circuit/merkle_tree/sparse_merkle_tree.rs
tessico: Right, I've renamed as suggested, and also unified across the two MT circuit impls: https://github.com/EspressoSystems/jellyfish/pull/172/commits/981a20e40dfb0dcd4ef1d6f4deefd726d5114c9a.

primitives/src/circuit/merkle_tree/sparse_merkle_tree.rs
tessico: Good catch, fixed here: https://github.com/EspressoSystems/jellyfish/pull/172/commits/82ca71a9da9de693812963a3892d81613eef25d8

primitives/src/circuit/merkle_tree/sparse_merkle_tree.rs
tessico: https://github.com/EspressoSystems/jellyfish/pull/172/commits/07e0aea9e80484ba89e201db34ed676ac2a2af8c

primitives/src/circuit/merkle_tree/rescue_merkle_tree.rs
tessico: Nice catch, fixed in https://github.com/EspressoSystems/jellyfish/pull/172/commits/82ca71a9da9de693812963a3892d81613eef25d8.

primitives/src/circuit/merkle_tree/rescue_merkle_tree.rs
tessico: Good suggestion, though I run into issues where the generic type `T` wouldn't know what is its `pos`, which is needed for a concrete impl.

I realized it's the same blocker as implementing a generic `From` for this type, which I have tried earlier to reduce code duplication. Notice how we have both:
```
impl<F: RescueParameter> From<&MembershipProof<F>>
    for MembershipProofBooleanEncoding<SparseMerkleTree<F>>
```

and 

```
impl<F: RescueParameter> From<&MembershipProof<F>>
    for MembershipProofBooleanEncoding<RescueMerkleTree<F>>
```

though of course `MembershipProof` is a different type, since both impls define their own type alias.
Maybe I'll consider adding a macro for generating this, as well as the whole impl of `MerkleTreeHelperGadget` - but we can probably leave it for another issue if I don't in this PR, it's just a nice to have.

primitives/src/circuit/merkle_tree/mod.rs
tessico: Done: https://github.com/EspressoSystems/jellyfish/pull/172/commits/c6fae19db16356b33ac0c8ab09ecc72fb61adc5a

primitives/src/circuit/merkle_tree/mod.rs
tessico: Yes we should make sure to add doctests in our new PR. Done [here](https://github.com/EspressoSystems/jellyfish/pull/172/commits/1c64b57625d1db40ad59b2fe5d1d6e2554f8528b)

primitives/src/circuit/merkle_tree/mod.rs
tessico: Here's my attempt at separating the path/proof, as well as better name for encoding. https://github.com/EspressoSystems/jellyfish/pull/172/commits/767d8d9c0fd9743add2dbc9c3d4a8cc277bf93ac
@EspressoSystems/jellyfish Would be good if we all reach consensus on the naming.

primitives/src/circuit/merkle_tree/mod.rs
tessico: Yes good point. I've changed it to be more explicit, let me know if you agree with the new name and docs:
https://github.com/EspressoSystems/jellyfish/pull/172/commits/3cc4ab218c53a218d797a7a59719a34efd87137d

primitives/src/circuit/merkle_tree/mod.rs
```diff
@@ -69,16 +93,211 @@ where
...
+
+/// Circuit variable for a Merkle proof of a 3-ary Merkle tree.
+/// Constains:
+/// * a list of node variables in the path,
+/// * a variable correseponsing to the position of the leaf element.
+#[derive(Debug, Clone)]
+pub struct Merkle3AryMembershipProofVar {
+    nodes: Vec<Merkle3AryNodeVar>,
+    pos: Variable,
+}
```
alxiong, line 303: I'd like to raise the consistency issue.

- `nodes` -> `proof`
- consider removing `Merkle3AryNodeRepr`, directly use `pub(crate) fn create_merkle_node(node: MerkleNode) -> Result<Merkle3AryNodeVar, E>`

primitives/src/circuit/merkle_tree/mod.rs
```diff
@@ -69,16 +93,211 @@ where
...
+        }
+    }
+}
+
+/// Intermediate encoding for Merkle path, used for constructing
+/// `Merkle3AryMembershipProofVar`. Constains a list of nodes in its path.
+#[derive(PartialEq, Eq, Clone, Debug)]
+pub(crate) struct MerklePathRepr<M: MerkleTreeScheme> {
+    pub(crate) nodes: Vec<Merkle3AryNodeRepr<M>>,
+}
```
alxiong, line 293: not very sure why we need this "seemingly intermediate" struct. why can't we only have `Merkle3AryMemebershipProofVar` ?

primitives/src/circuit/merkle_tree/mod.rs
```diff
@@ -30,10 +32,32 @@ mod rescue_merkle_tree;
 /// let (_, proof) = mt.lookup(2).expect_ok().unwrap();
 ///
 /// // Circuit computation with a MT
-/// let leaf_var = circuit.create_leaf_variable(2_u64, Fq::from(100_u64)).unwrap();
-/// let path_vars = circuit.create_membership_proof_variable(&proof).unwrap();
-/// let root_var = circuit.create_root_variable(expected_root).unwrap();
-/// circuit.enforce_merkle_proof(leaf_var, path_vars, root_var).unwrap();
+/// let leaf_var =
+///     <PlonkCircuit<Fq> as MerkleTreeGadget<RescueMerkleTree<Fq>>>::create_leaf_variable(
```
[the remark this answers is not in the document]
tessico: Good catch, thx, fixed!

primitives/src/circuit/merkle_tree/mod.rs
```diff
@@ -46,7 +70,7 @@ where
     /// Type to represent the merkle path of the concrete MT instantiation.
     /// It is MT-specific, e.g arity will affect the exact definition of the
     /// Merkle path.
-    type MerklePathVar;
+    type MerkleProofVar;
```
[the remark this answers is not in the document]
tessico: > * the code comment is not updated (still use "merkle path")
> * can we have names that's more aligned with their non-circuit counterpart.
>   
>   * `type LeafVar` -> `type ElemVar`

I agree we should aim for consistency with native impl.
Maybe we should change the native impl naming though? I feel like Leaf is much more descriptive than Element.
A worry I have is that for example `create_leaf_variable` -> `create_element_variable` becomes pretty confusing, at least to me.

primitives/src/circuit/merkle_tree/mod.rs
```diff
@@ -69,16 +93,211 @@ where
...
+/// ```
+pub trait UniversalMerkleTreeGadget<M>: MerkleTreeGadget<M>
+where
+    M: MerkleTreeScheme,
+{
+    /// checking non-membership proof
+    fn is_non_member(
+        &mut self,
+        elem: Self::LeafVar,
+        merkle_proof: Self::MerkleProofVar,
```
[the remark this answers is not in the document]
tessico: Done in https://github.com/EspressoSystems/jellyfish/pull/172/commits/94db35fab3e50473d06c8015f42e3b6b3f9b9106

primitives/src/circuit/merkle_tree/mod.rs
```diff
@@ -30,10 +32,32 @@ mod rescue_merkle_tree;
...
-/// let root_var = circuit.create_root_variable(expected_root).unwrap();
-/// circuit.enforce_merkle_proof(leaf_var, path_vars, root_var).unwrap();
+/// let leaf_var =
+///     <PlonkCircuit<Fq> as MerkleTreeGadget<RescueMerkleTree<Fq>>>::create_leaf_variable(
+///         &mut circuit,
+///         2_u64,
+///         Fq::from(100_u64)
+///     )
+///     .unwrap();
+/// let path_vars =
```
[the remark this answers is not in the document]
tessico: Done in https://github.com/EspressoSystems/jellyfish/pull/172/commits/ac696412234c0a1babe6b85e2dc9218c3b3201df

primitives/src/circuit/merkle_tree/mod.rs
```diff
@@ -69,16 +93,211 @@ where
     fn is_member(
         &mut self,
         elem: Self::LeafVar,
-        merkle_proof: Self::MerklePathVar,
+        merkle_proof: Self::MerkleProofVar,
         merkle_root: Variable,
```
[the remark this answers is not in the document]
tessico: Agree, modified in https://github.com/EspressoSystems/jellyfish/pull/172/commits/94db35fab3e50473d06c8015f42e3b6b3f9b9106

primitives/src/circuit/merkle_tree/mod.rs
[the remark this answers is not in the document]
alxiong: interesting proposal, so you are saying we introduce `Leaf` concept in native impl? let's discuss this over a call today @mrain @tessico.

primitives/src/circuit/merkle_tree/mod.rs
tessico: Yeah okey, I've refactored the code slightly to not use `Repr` structs anymore, in 721f9fd129cc054e9c66cb0c145060432a1b55aa

primitives/src/circuit/merkle_tree/mod.rs
```diff
@@ -30,10 +32,32 @@ mod rescue_merkle_tree;
...
+///         &proof
+///     )
+///     .unwrap();
+/// let root_var =
+///     <PlonkCircuit<Fq> as MerkleTreeGadget<RescueMerkleTree<Fq>>>::create_root_variable(
+///         &mut circuit,
+///         expected_root
+///     )
+///     .unwrap();
+/// <PlonkCircuit<Fq> as MerkleTreeGadget<RescueMerkleTree<Fq>>>::enforce_membership_proof(
```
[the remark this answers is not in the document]
tessico: I don't think it's necessary, at least not for the trait. The example shows how to import the trait and init a concrete instantiation of it. I wouldn't necessarily showcase each and every method of it.
If we want to do that, we could add rustdocs to the individual methods. We could consider that, but so far I think it's only been traits that we are documenting with examples. WDYT? Also @mrain @chancharles92

primitives/src/circuit/merkle_tree/mod.rs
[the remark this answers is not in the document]
tessico: > okay I notice a mismatch of concepts.
> 
> in mt impl, `verify(&self, pos, proof)` the `elem` is assumed to be part of `proof`; whereas our mt_circuit impl, `is_member(leaf, proof, root)` where `leaf = uid + elem` and `proof` doesn't have `elem`?
> 
> we should make them consistent
> 
> * i think `is_member(root_var, uid_var, proof_var)` may seem longer, but clearer and less error prone IMHO. and matches with non-circuit impl, where `&self` corresponds to `root_var`.
> * making `MerkleProofVar` inclusive of `elem_var` makes it more consistent

Addressed these in https://github.com/EspressoSystems/jellyfish/pull/172/commits/59e321c6b87dab4e65c647afd995eea344e76310. Good suggestions, thx @alxiong!

For now I've taken the liberty to also rename the fields in circuit impl from `elem` -> `leaf`. Since the usage was already mixed (leaf & elem) inside the circuit impl, at least here we can have consistency. TBD is whether the native impl should switch to `leaf` (or otherwise leaf here back to `elem`), but I think this can be a separate change, separate PR?

primitives/src/circuit/merkle_tree/mod.rs
tessico: Also addressed naming in https://github.com/EspressoSystems/jellyfish/pull/172/commits/79c13166aff5a59aebe1be478f85127d9bdd2432 and removal of `Merkle3AryNodeRepr` in https://github.com/EspressoSystems/jellyfish/pull/172/commits/721f9fd129cc054e9c66cb0c145060432a1b55aa

primitives/src/circuit/merkle_tree/mod.rs
alxiong: yes, I like the new name, 
- meanwhile given `constrain_sibling_order()-> Result<[Variable; 3], E>`, should we rename `MerkleTreeHelperGadget` -> `3AryMerkleTreeHelperGadget`. 
- more importantly, after giving some thoughts, I question if we really need `pub(crate) trait MerkleTreeHelperGadget` 🤔 

originally, we have the helper gadget because we want to implement non-private methods on a foreign struct (i.e. `struct PlonkCircuit`). The two ways we can do it is: 1) our crate-internal helper trait; 2) just make it a function instead of a method of that struct.
After reading the usage, `constrain_membership_proof()` and `compute_merkle_root()` can be inlined as they are really called once (meaning they don't even need to be private functions at all!)
we did invoke `constrain_silbing_order()` couple of times, therefore may I suggest we have the following private stand-alone function:
`fn constrain_silbing_order<F>(circuit: &mut PlonkCircuit<F>, ...) -> ...`

I like removing unnecessary mental load on "why we need such helper gadget again" when we look at the code in the future.

alxiong: frankly the triggering cringy point is when I saw the redundant associates type `MerkleTreeHelperGadget::MembershipProofVar`, that's an indication that we might be designing our API wrong-ish.

primitives/src/circuit/merkle_tree/mod.rs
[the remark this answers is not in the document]
alxiong: sure, this is minor issue.

primitives/src/circuit/merkle_tree/mod.rs
```diff
@@ -30,55 +32,226 @@ mod rescue_merkle_tree;
...
+{
+    /// Type to represent the merkle non-membership proof of the concrete MT
+    /// instantiation. It is MT-specific, e.g arity will affect the exact
+    /// definition of the underlying Merkle path.
+    type NonMembershipProofVar;
+
+    /// Allocate a variable for the membership proof.
+    fn create_non_membership_proof_variable(
+        &mut self,
+        membership_proof: &M::MembershipProof,
```
alxiong, line 158: this makes more sense?
```rust 
pub trait UniversalMerkleTreeGadget<M>: MerkleTreeGadget<M>
where
    M: UniversalMerkleTreeScheme,

fn create_non_membership_proof_variable(
        &mut self,
        proof: &M::NonMembershipProof,
)
```

primitives/src/circuit/merkle_tree/sparse_merkle_tree.rs
```diff
@@ -0,0 +1,581 @@
...
+            self,
+            leaf_idx_var,
+            proof_var,
+            expected_root_var,
+        )?;
+        self.enforce_true(bool_val.into())
+    }
+}
+
+trait SparseMerkleTreeHelperGadget<F: RescueParameter> {
```
alxiong, line 155: similar to my comment here: https://github.com/EspressoSystems/jellyfish/pull/172#discussion_r1069233944, I tend to prefer removing this trait and using a private stand-alone function instead of methods of `PlonkCircuit` instead.

alxiong (commented):
I'd like to toast to @tessico 🥂 here, I believe this PR has improved our code quality quite a bit, by removing redundant code and unnecessary intermediate abstractions. 

Just a few more small comments.

primitives/src/circuit/merkle_tree/sparse_merkle_tree.rs
```diff
@@ -0,0 +1,581 @@
...
+            merkle_proof.tree_height() - 1,
+        );
+
+        let leaf_elem = match merkle_proof.elem() {
+            Some(elem) => elem,
+            None => {
+                return Err(CircuitError::InternalError(
+                    "The proof doesn't contain a leaf element".to_string(),
+                ))
+            },
```
alxiong, line 229: should this be `ParameterError` instead? (conventionally, we use `InternalError` to represents something wrong with our code whereas this is a user-provided input which might be malformed)

primitives/src/circuit/merkle_tree/mod.rs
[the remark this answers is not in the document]
alxiong: I generally like your change (esp removing `LeafVar`)
but my confusion is: 

```rust
pub struct Merkle3AryMembershipProofVar {
    node_vars: Vec<Merkle3AryNodeVar>, // <-- this should be called path_var instead?
    leaf_var: Variable, // <-- why is this called leaf_var instead of elem_var
}
```

and since we already removed concept of "leaf" for user-facing interfaces, (I'm fine with internal variable named `leaf_xxx` to refer to the union of `uid` and `elem`) , why we still use `leaf_idx_var` instead of `elem_idx_var`?

primitives/src/circuit/merkle_tree/mod.rs
tessico: yes inlining `compute_merkle_root` crossed my mind too, TBH, though I was ok with leaving it in, as it was an internal impl detail. But yes, if you also see that it's a bit of a complication, I think it's best to remove the Helper gadget completely.

primitives/src/circuit/merkle_tree/sparse_merkle_tree.rs
tessico: You're absolutely right.

primitives/src/circuit/merkle_tree/mod.rs
tessico: Yes, good catch!

tessico: Done in https://github.com/EspressoSystems/jellyfish/pull/172/commits/f2d2e8e0aa915c3025ac117c040a3b3ff93e2c36

primitives/src/circuit/merkle_tree/sparse_merkle_tree.rs
tessico: Done in https://github.com/EspressoSystems/jellyfish/pull/172/files/f2d2e8e0aa915c3025ac117c040a3b3ff93e2c36..1a266b75b17276b05eca60092d78ec07b02901b5

primitives/src/circuit/merkle_tree/mod.rs
tessico: Done in https://github.com/EspressoSystems/jellyfish/pull/172/files/f2d2e8e0aa915c3025ac117c040a3b3ff93e2c36..1a266b75b17276b05eca60092d78ec07b02901b5

primitives/src/circuit/merkle_tree/sparse_merkle_tree.rs
tessico: Done in c0acab04c0d94c9df4c38a04927bcefa0f2f6f2d

alxiong (approved):
Hooray! 🥳 LFG!

Issue #141, opened 2022-11-18
Light-weight merkle tree implementation


mrain:
A light weight merkle tree is an append-only merkle tree that only keeps it's frontier.

PR #176, opened 2023-01-03
Light weight merkle tree implementation.

mrain:
## Description

closes: #141 

Gadget implementation not required.

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/primitives/src/merkle_tree/internal.rs
+++ b/primitives/src/merkle_tree/internal.rs
@@ -66,6 +66,11 @@ where
             Self::ForgettenSubtree { value } => *value,
         }
     }
+
+    #[inline]
+    pub(crate) fn is_forgotten(&self) -> bool {
+        matches!(self, Self::ForgettenSubtree { .. })
+    }
 }
 
 /// A merkle commitment consists a root hash value, a tree height and number of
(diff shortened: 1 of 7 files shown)


primitives/src/merkle_tree/internal.rs
```diff
@@ -226,6 +231,90 @@ where
...
+pub(crate) fn build_light_weight_tree_internal<E, H, I, Arity, T>(
+    height: usize,
+    elems: impl IntoIterator<Item = impl Borrow<E>>,
+) -> Result<(Box<MerkleNode<E, I, T>>, u64), PrimitivesError>
+where
+    E: Element,
+    H: DigestAlgorithm<E, I, T>,
+    I: Index + From<u64>,
+    Arity: Unsigned,
+    T: NodeValue,
```
chancharles92, line 244: What's the difference from `NodeValue` and `Element`? Is that `Element` has a uid?

primitives/src/merkle_tree/internal.rs
```diff
@@ -226,6 +231,90 @@ where
...
+{
+    let leaves: Vec<_> = elems.into_iter().collect();
+    let num_leaves = leaves.len() as u64;
+    let capacity = BigUint::from(Arity::to_u64()).pow(height as u32);
```
chancharles92, line 248: How large the tree can be? Is it necessary to use `BigUint`?

primitives/src/merkle_tree/internal.rs
```diff
@@ -576,7 +665,107 @@ where
...
+            MerkleNode::Leaf { .. } => Err(PrimitivesError::ParameterError(
+                "Incompatible merkle tree: index already occupied".to_string(),
+            )),
+            MerkleNode::ForgettenSubtree { .. } => Err(PrimitivesError::ParameterError(
+                "Given part of merkle tree is not in memory".to_string(),
+            )),
+        }
+    }
+
+    pub(crate) fn extend_and_forget_internal<H, Arity>(
```
chancharles92, line 691: Maybe add some comments explaining the API's functionality and the usage of the API arguments.

chancharles92 (commented):
👍

primitives/src/merkle_tree/internal.rs
mrain: `NodeValue` is the internal Merkle node value. `Element` is the type of the stored element.

primitives/src/merkle_tree/internal.rs
mrain: Good point. For an append-only light-weight Merkle tree `u64` or even `u32` suffices. Let me change it, with an additional range check.

chancharles92: Thanks, let's still use u64 for safety and future-compatibility.

primitives/src/merkle_tree/light_weight.rs
```diff
@@ -0,0 +1,195 @@
...
+    #[test]
+    fn test_light_mt_insertion() {
+        test_light_mt_insertion_helper::<Fq254>();
+        test_light_mt_insertion_helper::<Fq377>();
+        test_light_mt_insertion_helper::<Fq381>();
+    }
+
+    fn test_light_mt_insertion_helper<F: RescueParameter>() {
+        let mut mt = RescueLightWeightMerkleTree::<F>::from_elems(2, &[]).unwrap();
+        assert_eq!(mt.capacity(), BigUint::from(9u64));
```
chancharles92, line 99: Is `capacity()` a BigUint?

primitives/src/merkle_tree/light_weight.rs
```diff
@@ -0,0 +1,195 @@
...
+
+    #[test]
+    fn test_light_mt_builder() {
+        test_light_mt_builder_helper::<Fq254>();
+        test_light_mt_builder_helper::<Fq377>();
+        test_light_mt_builder_helper::<Fq381>();
+    }
+
+    fn test_light_mt_builder_helper<F: RescueParameter>() {
+        assert!(RescueLightWeightMerkleTree::<F>::from_elems(1, &[F::from(0u64); 3]).is_ok());
```
chancharles92, line 86: Maybe make `3` a parameter (e.g. `let arity = RescueLightWeightMerkleTree::ARITY`)? Then replace the numbers below to be expressions related to `arity`.

The advantage is that when we change the setup of `RescueLightWeightMerkleTree`, the tests won't be broken.

primitives/src/merkle_tree/internal.rs
```diff
@@ -226,6 +231,92 @@ where
...
+                    })
+                })
+                .collect_vec();
+            for i in 1..cur_nodes.len() - 1 {
+                cur_nodes[i] = Box::new(MerkleNode::ForgettenSubtree {
+                    value: cur_nodes[i].value(),
+                })
+            }
+        }
+        Ok((cur_nodes[0].clone(), num_leaves))
```
chancharles92, line 314: What's the difference between `cur_nodes[0]` and `cur_nodes[0].clone()` here?

primitives/src/merkle_tree/internal.rs
```diff
@@ -576,7 +673,109 @@ where
...
+                "Incompatible merkle tree: index already occupied".to_string(),
+            )),
+            MerkleNode::ForgettenSubtree { .. } => Err(PrimitivesError::ParameterError(
+                "Given part of merkle tree is not in memory".to_string(),
+            )),
+        }
+    }
+
+    /// Similar to [`extend_internal`], but this function will automatically
+    /// forget every leave except for the Merkle tree frontier.
```
chancharles92, line 700: every leave -> every leaf

primitives/src/merkle_tree/light_weight.rs
mrain: Yes. It's from the trait definition of `MerkleTree`.

primitives/src/merkle_tree/internal.rs
mrain: `cur_nodes` is a local variable and it cannot be moved outside.

primitives/src/merkle_tree/light_weight.rs
mrain: Good point, will do.

primitives/src/merkle_tree/internal.rs
chancharles92: Ah right, it's a local vector; we can't just move the first element of it outside.

chancharles92 (approved):

PR #180, opened 2023-01-12
Add ForgetableUniversalMerkleTreeScheme

jbearer:
_Needed this for a personal project, not work-related. Review at your leisure. I believe it does fix a bug, though._

This trait extends the existing `UniversalMerkleTreeScheme` with two new functions:
* `universal_forget` is like `forget` from `ForgetableMerkleTreeScheme`, but returns a non-membership proof if the result is `NotFound`
* `non_membership_remember` re-inserts an empty sub-tree given a non-membership proof

Also fixes a bug in `forget_internal` where `height` was not decremented in recursive calls.

### DIFF
--- a/primitives/src/merkle_tree/internal.rs
+++ b/primitives/src/merkle_tree/internal.rs
@@ -73,10 +73,21 @@ where
     }
 }
 
+type MerklePath<E, I, T> = Vec<MerkleNode<E, I, T>>;
+
 /// A merkle commitment consists a root hash value, a tree height and number of
 /// leaves
 #[derive(
-    Eq, PartialEq, Clone, Copy, Ord, PartialOrd, Hash, CanonicalSerialize, CanonicalDeserialize,
+    Eq,
+    PartialEq,
+    Clone,
+    Copy,
+    Debug,
+    Ord,
+    PartialOrd,
+    Hash,
+    CanonicalSerialize,
+    CanonicalDeserialize,
 )]
 #[tagged("MERKLE_COMM")]
 pub struct MerkleTreeCommitment<T: NodeValue> {
@@ -126,7 +137,7 @@ where
     #[serde(with = "canonical")]
     pub pos: I,
     /// Nodes of proof path, from root to leaf
-    pub proof: Vec<MerkleNode<E, I, T>>,
+    pub proof: MerklePath<E, I, T>,
 
     /// Place holder for Arity
     _phantom_arity: PhantomData<Arity>,
@@ -143,7 +154,7 @@ where
         self.proof.len()
     }
 
-    pub fn new(pos: I, proof: Vec<MerkleNode<E, I, T>>) -> Self {
+    pub fn new(pos: I, proof: MerklePath<E, I, T>) -> Self {
         MerkleProof {
             pos,
             proof,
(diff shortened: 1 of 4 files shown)


primitives/src/merkle_tree/internal.rs
```diff
@@ -341,11 +352,13 @@ where
...
         traversal_path: &[usize],
-    ) -> LookupResult<E, Vec<MerkleNode<E, I, T>>, ()> {
+    ) -> LookupResult<E, MerklePath<E, I, T>, MerklePath<E, I, T>> {
         match self {
-            MerkleNode::Empty => LookupResult::NotFound(()),
+            MerkleNode::Empty => LookupResult::NotFound(vec![MerkleNode::Empty; height + 1]),
             MerkleNode::Branch { value, children } => {
-                match children[traversal_path[height - 1]].forget_internal(height, traversal_path) {
+                match children[traversal_path[height - 1]]
+                    .forget_internal(height - 1, traversal_path)
```
mrain, line 360: nice catch :)

primitives/src/merkle_tree/internal.rs
```diff
@@ -410,67 +440,39 @@ where
...
+                // Recurse into the appropriate sub-tree to remember the rest of the path.
+                let mut children = children.clone();
+                children[traversal_path[height - 1]].remember_internal::<H, Arity>(
+                    height - 1,
+                    traversal_path,
+                    path_values,
+                    proof,
+                )?;
+                // Compute new value and remember `*self`.
+                *self = Self::Branch {
```
mrain, line 455: I'm thinking about the necessity of recomputing the internal node value. `Remember` function shouldn't change the internal value, right?

jbearer: Oh of course. I was thinking I can't use the value from the proof because they're all zeroed, but forgot I could just get it from `self`. Good catch, will fix

mrain (approved):
LGTM. Thanks Jeb!

primitives/src/merkle_tree/mod.rs
```diff
@@ -341,3 +341,41 @@ pub trait ForgetableMerkleTreeScheme: MerkleTreeScheme {
...
+    /// function is called for a `pos` which is in memory but not in the
+    /// tree, it will return a non-membership proof.
+    ///
+    /// The return value is the same as if
+    /// [universal_lookup](UniversalMerkleTreeScheme::universal_lookup) were
+    /// called before this call.
+    fn universal_forget(
+        &mut self,
+        pos: Self::Index,
+    ) -> LookupResult<Self::Element, Self::MembershipProof, Self::NonMembershipProof>;
```
alxiong, line 370: thanks for the detailed code doc, question: is this more like a convenient API that combines the forget-then-universal-lookup two steps into one? or is this functionally different?

primitives/src/merkle_tree/mod.rs
```diff
@@ -341,3 +341,41 @@ pub trait ForgetableMerkleTreeScheme: MerkleTreeScheme {
...
+
+    /// "Re-insert" an empty leaf into the tree using its proof.
+    ///
+    /// Returns `Ok(())` if insertion is successful, or `Err(err)` if the proof
+    /// disagrees with the merkle tree
+    fn non_membership_remember(
+        &mut self,
+        pos: Self::Index,
+        proof: impl Borrow<Self::NonMembershipProof>,
+    ) -> Result<(), PrimitivesError>;
```
alxiong, line 380: just checking my understanding. since there are `NotInMemory` and `NotFound` states, sometimes, we may forgot a non-member in a SMT, turning a non-member from `Ok` to `NotInMemory`

and this `non_member_remeber`  turn these non-members from `NotInMemory` back to `Ok`, cuz they are never `NotFound` in a SMT. so that we can produce non-membership proof quickly in the near future, correct?

Issue #178, opened 2023-01-09
Migrate the multilinear PCS batch opening API


chancharles92:
Migrate from https://github.com/EspressoSystems/hyperplonk/blob/main/subroutines/src/pcs/multilinear_kzg/batching.rs, which has better complexity

PR #182, opened 2023-01-16
Update naming for IOPTranscript

mrain:
## Description

closes: #159 
After investigation, the transcript modules in `primitives/pcs` and `plonk` are not overlapping. Former one is an IOPTranscript that help build batch commitment/opening for PCS and the later one is specific for plonk.
It's okay to keep both modules.
I do rename all the method `multi_*` to `batch_*`, because I think the later one is more clear. The former one may confuse users with the other term "multi-linear extension".

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/primitives/src/pcs/mod.rs
+++ b/primitives/src/pcs/mod.rs
@@ -92,13 +92,14 @@ pub trait PolynomialCommitmentScheme<E: PairingEngine> {
     /// &Self::ProverParam, ..)` or `commit(prover_param:
     /// Box<Self::ProverParam>, ..)` or `commit(prover_param:
     /// Arc<Self::ProverParam>, ..)` etc.
+    /// Also, the commitment is not necessarily hiding.
     fn commit(
         prover_param: impl Borrow<Self::ProverParam>,
         poly: &Self::Polynomial,
     ) -> Result<Self::Commitment, PCSError>;
 
-    /// Generate a commitment for a list of polynomials
-    fn multi_commit(
+    /// Batch commit a list of polynomials
+    fn batch_commit(
         prover_param: impl Borrow<Self::ProverParam>,
         polys: &[Self::Polynomial],
     ) -> Result<Self::BatchCommitment, PCSError>;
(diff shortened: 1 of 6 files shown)


primitives/src/pcs/mod.rs
```diff
@@ -92,13 +92,14 @@ pub trait PolynomialCommitmentScheme<E: PairingEngine> {
     /// &Self::ProverParam, ..)` or `commit(prover_param:
     /// Box<Self::ProverParam>, ..)` or `commit(prover_param:
     /// Arc<Self::ProverParam>, ..)` etc.
+    /// Also, the commitment is not necessarily hiding.
```
tessico, line 95: Is is sometimes hiding and sometimes not, depending on the implementation? I suppose it's actually never hiding, in which case we should say it's explicitly **non hiding**.

chancharles92 (commented):
LGTM, do you think it's possible also to finish issue #178 and migrate the file https://github.com/EspressoSystems/hyperplonk/blob/main/subroutines/src/pcs/multilinear_kzg/batching.rs to Jellyfish? @mrain (It's also fine to create another PR for #178 )

mrain:
> LGTM, do you think it's possible also to finish issue #178 and migrate the file https://github.com/EspressoSystems/hyperplonk/blob/main/subroutines/src/pcs/multilinear_kzg/batching.rs to Jellyfish? @mrain (It's also fine to create another PR for #178 )

This looks like a good starting issue for @ggutoski . WDYT?

primitives/src/pcs/mod.rs
mrain: Hiding is usually required for many commitment scheme. However, in our use case, Plonk doesn't need PCS to be hiding. Also our KZG implementation is not hiding. I'm just putting this as some kind of warning.

chancharles92:
> > LGTM, do you think it's possible also to finish issue #178 and migrate the file https://github.com/EspressoSystems/hyperplonk/blob/main/subroutines/src/pcs/multilinear_kzg/batching.rs to Jellyfish? @mrain (It's also fine to create another PR for #178 )
> 
> This looks like a good starting issue for @ggutoski . WDYT?

Discussed with @mrain , @mrain and I will work on it. @ggutoski you can focus on Hotshot first.

primitives/src/pcs/mod.rs
tessico: That's what I mean. Usually (always?) hiding would require an additional randomness input, and since in our API we don't even have it, it cannot be hiding, so we should just say it's **not hiding**.

mrain:
> > > LGTM, do you think it's possible also to finish issue #178 and migrate the file https://github.com/EspressoSystems/hyperplonk/blob/main/subroutines/src/pcs/multilinear_kzg/batching.rs to Jellyfish? @mrain (It's also fine to create another PR for #178 )
> > 
> > 
> > This looks like a good starting issue for @ggutoski . WDYT?
> 
> Discussed with @mrain , @mrain and I will work on it. @ggutoski you can focus on Hotshot first.

We decided not to do so in the release. Because we then need to migrate the whole poly-iop module.

primitives/src/pcs/mod.rs
mrain: Even if there's no randomness input. The PCS could bake some rng inside to make it somewhat hiding. Isn't it?

mrain: Second thought, I'll take your suggestion and propose a low-priority issue to add hiding property to the PCS trait.

chancharles92 (approved):
LGTM

chancharles92:
LGTM, but there is a format check warning.

primitives/src/pcs/mod.rs
tessico: ok fair

PR #193, opened 2023-01-19
Make scripts error as expected

Ancient123:


alxiong (approved):

tessico:
I don't see these used anywhere (except two scripts mentioned in the readme).
Also, it seems that the no_std script actually errors out (I acknowledge the tool is not supposed to be perfect yet), but it suggests:

```
jf-relation: ❌
  - Did not find a #![no_std] attribute or a simple conditional attribute like #[cfg_attr(not(feature = "std"), no_std)] in the crate source. Crate most likely doesn't support no_std without changes.
 ```
which makes me wonder whether we should add it to `jf-relation` (it's the only crate in jellyfish that doesn't have that attribute)?

PR #200, opened 2023-02-06
Fix clippy error in CI

alxiong:
## Description

Quick fix CI error [here](https://github.com/EspressoSystems/jellyfish/runs/11123870191).

Changes include:
- Inlined all format args, see clippy [hints here](https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args): e.g. `format!("{:?}, error")` -> `format!("{error:?}")` for better readability
  - reason why we didn't catch previously in CI or locally is because our CI is not running inside nix-shell, thus two CI runs might end up using different clippy and rust (based on upstream release), and there has been a new release `1.67.0` which add this clippy rule. No big deal.
- temporarily use `version = {workspace = true}` instead of `version.workspace = true` syntax, due to https://github.com/DevinR528/cargo-sort/issues/47 (so that our precommit hook won't fail) 

## Post PR TODO:
- [ ] re-tag `0.2.0`

### DIFF
--- a/plonk/Cargo.toml
+++ b/plonk/Cargo.toml
@@ -1,11 +1,13 @@
 [package]
 name = "jf-plonk"
 description = "UltraPlonk implementation"
-version.workspace = true
-authors.workspace = true
-edition.workspace = true
-license.workspace = true
-rust-version.workspace = true
+# can change back to version.workspace = true after the following issue is fixed:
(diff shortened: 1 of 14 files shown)


chancharles92:
Thanks @alxiong ! There are still 2 clippy warnings in unchanged files. Could you help fix it? Thanks

chancharles92 (approved):
LGTM

Issue #218, opened 2023-03-28
Support BN curves in the BLS aggregate signature


chancharles92:
BLS signature on ethereum (uses BN256 curve)
- Not implemented in [[blst](https://docs.rs/blst/0.3.10/blst/index.html)](https://docs.rs/blst/0.3.10/blst/index.html) rust library we are using now
- Problem: hash to curve algorithm not specified in [https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve) for the BN256 curve, even though some generic approach is described at [https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-09#section-3](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-09#section-3)
- Solutions
   - Use [https://github.com/arkworks-rs/algebra/blob/1f7b3c6b215e98fa3130b39d2967f6b43df41e04/ec/src/hashing/map_to_curve_hasher.rs#L49](https://github.com/arkworks-rs/algebra/blob/1f7b3c6b215e98fa3130b39d2967f6b43df41e04/ec/src/hashing/map_to_curve_hasher.rs#L49)
   
cc @philippecamacho @mrain

philippecamacho:
# Update regarding hashing to curve.

## Context

We need to implement a hash to curve algorithm for the pairing friendly curve currently supported by Ethereum which is [Bn254](https://hackmd.io/@jpw/bn254). We can either implement the naive method sometimes called "hash and pray" that consists of taking a field element $x$ , plugging it into the elliptic curve equation and hoping it is a quadratic residue. If not we try with $x+1$, if this fails again we try with $x+2$ and so on and so forth. This method takes only a few iterations as the probability of failure for k attempts heuristically is $2^{-k}$. A problem of this approach is that the time to compute the hash value is not constant which may leak information about the input or allow for some DoS attack if the hash function is computed inside a smart contract and an attacker is able to find an input $x$ so that $k$ is large.

## Constant time hashing functions

Other hashing methods that are constant time exist though.

[...]

PR #223, opened 2023-03-30
Feat/t218 bls sig arkworks

philippecamacho:
Closes: #218

### DIFF
--- a/primitives/src/signatures/bls_over_bn254.rs
+++ b/primitives/src/signatures/bls_over_bn254.rs
@@ -214,13 +214,12 @@ impl PartialEq for Signature {
 pub fn hash_to_curve<H: Default + DynDigest + Clone>(msg: &[u8]) -> G1Projective {
     let hasher_init = &[1u8];
     let hasher = <DefaultFieldHasher<H> as HashToField<BaseField>>::new(hasher_init);
-    let field_elems: Vec<BaseField> = hasher.hash_to_field(msg, 1);
 
     // General equation of the curve: y^2 = x^3 + ax + b
     // For BN254 we have a=0 and b=3 so we only use b
     let coeff_b: BaseField = MontFp!("3");
 
-    let mut x = field_elems[0];
+    let mut x: BaseField = hasher.hash_to_field(msg, 1)[0];
     let mut Y: BaseField = x * x * x + coeff_b;
 
     // Loop until we find a quadratic residue
--- a/primitives/src/signatures/bls_over_bn254.rs
+++ b/primitives/src/signatures/bls_over_bn254.rs
@@ -4,8 +4,36 @@
 // You should have received a copy of the MIT License
 // along with the Jellyfish library. If not, see <https://mit-license.org/>.
 
-//! This module implements the BLS signature over the BN254 curve.ç
-// TODO more comments
+//! This module implements the BLS signature over the BN254 curve.
+//! The [BLS signature scheme][bls] relies on a bilinear map `e:G1 x G2 -> GT`
+//! and a hash function `H` from a message input space (e.g. bytes) to a group
+//! element `G1`. `H` must be such that for all `m`, the discrete logarithm
+//! `H(m)` w.r.t to the generator of `G1` is unknown.
+//!
+//! The scheme works as follows:
+//! Let `g1` and `g2` be generators of `G1` and `G2`.
+//!
+//! **KeyGen()**
+//!    * sample a random `s` in the scalar field `Fr` and return the key pair
+//!      `(sk,pk):=(s,g2^s)`
+//!
+//! **Sign(sk,m)**
+//!    * return `sigma=H(m)^{sk}`
+//!
+//! **Verify(pk,m,sigma)**
+//!    * Check that `e(sigma,g_2)=e(H(m),pk)`
+//!
+//! In this module:
+//! * `e` is the pairing over the curve [BN254][bn254] supported by the EVM
+//!   [EIP-196][eip196], [EIP197][eip197]
+//! * `H` is implemented using the "hash-and-pray" approach. See function
+//!   [`hash_to_curve`]
+//!
+//! [bls]: https://hovav.net/ucsd/dist/sigs.pdf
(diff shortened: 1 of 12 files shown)


primitives/src/signatures/bls_over_bn254.rs
```diff
@@ -0,0 +1,399 @@
...
+    }
+
+    /// Get the internal of the signing key, namely a P::ScalarField element
+    pub fn sign_key_internal(&self) -> &ScalarField {
+        &self.sk.0
+    }
+
+    /// Signature function
+    pub fn sign<B: AsRef<[u8]>>(&self, msg: &[u8], csid: B) -> Signature {
+        let msg_input = [msg, csid.as_ref()].concat();
```
chancharles92, line 272: Do we need to hash the message into a fixed-length digest first?

chancharles92 (commented):
Thanks @philippecamacho ! Left a few comments.

primitives/src/signatures/bls_over_bn254.rs
```diff
@@ -0,0 +1,399 @@
...
+        Ok(())
+    }
+
+    /// Sample a pair of keys.
+    fn key_gen<R: CryptoRng + RngCore>(
+        _pp: &Self::PublicParameter,
+        prng: &mut R,
+    ) -> Result<(Self::SigningKey, Self::VerificationKey), PrimitivesError> {
+        let kp = KeyPair::generate(prng);
+        Ok((kp.sk.clone(), kp.vk.clone()))
```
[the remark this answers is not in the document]
philippecamacho: Yes, it is possible to remove this call to `clone`. See 18f5243b.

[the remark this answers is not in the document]
philippecamacho: Yes, it is possible to remove this call to `clone`. See 18f5243b.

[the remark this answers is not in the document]
chancharles92: Ah I was wondering if it's OK to move `kp.vk` without cloning/copying. Anyway, it's a minor comment.

primitives/src/signatures/bls_over_bn254.rs
```diff
@@ -0,0 +1,399 @@
...
+/// y^2=x^3+3)   3. Check if `Y` is a quadratic residue (QR), in which case
+/// return `y=sqrt(Y)` otherwise try with `x+1, x+2` etc... until `Y` is a QR.
+///   4. Return `P=(x,y)`
+///
+///  In the future we may switch to a constant time algorithm such as Fouque-Tibouchi <https://www.di.ens.fr/~fouque/pub/latincrypt12.pdf>
+/// * `H` - parameterizable hash function (e.g. SHA256, Keccak)
+/// * `msg` - input message
+/// * `returns` - A group element in G1
+#[allow(non_snake_case)]
+pub fn hash_to_curve<H: Default + DynDigest + Clone>(msg: &[u8]) -> G1Projective {
```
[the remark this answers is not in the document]
philippecamacho: The `hash_to_field` function calls [self.expander.expand](https://github.com/arkworks-rs/algebra/blob/bc991d44c5e579025b7ed56df3d30267a7b9acac/ff/src/fields/field_hashers/expander/mod.rs#L68) where most of the logic is placed. In particular the message is first hashed with a bytes to bytes hash function such as SHA256 of Keccak and then some field(s) elements are derived. Does this answer your question?

primitives/src/signatures/bls_over_bn254.rs
philippecamacho: As discussed [above](https://github.com/EspressoSystems/jellyfish/pull/223#discussion_r1158635887), I don' t think this is necessary.

primitives/src/signatures/bls_over_bn254.rs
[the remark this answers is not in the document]
chancharles92: OK, so essentially the `hash_to_field` already somewhat executes the hash-to-digest step internally. Make sense, thanks!

primitives/src/signatures/bls_over_bn254.rs
```diff
@@ -0,0 +1,398 @@
...
+                let pk = key_pair.ver_key_ref();
+                assert!(pk.verify(&msg, &sig, CS_ID_BLS_BN254).is_ok());
+                // wrong public key
+                assert!(pk_bad.verify(&msg, &sig, CS_ID_BLS_BN254).is_err());
+                // wrong message
+                msg.push(i as u8);
+                assert!(pk.verify(&msg, &sig, CS_ID_BLS_BN254).is_err());
+            }
+        }
+    }
```
chancharles92, line 355: Shall we additionally test 2 messages m1, m2=(m1||x) whose lengths are larger than a base field length? E.g., |m1| = 35B, |x| = 1B

philippecamacho: Yes it is a good idea, but why do we need two messages? Is it not enough to test with m1?

philippecamacho: E.g:

```rust
// Test for long messages
let size=35; // Bigger than 32 which is the number of bytes needed to encode a field element
let key_pair = KeyPair::generate(&mut rng);
let mut msg = [33u8;35];
let sig = key_pair.sign(&msg, CS_ID_BLS_BN254);
let pk = key_pair.ver_key_ref();
assert!(pk.verify(&msg, &sig, CS_ID_BLS_BN254).is_ok());
```

chancharles92: To check that a signature for m1 won't verify for m2, or vice versa.

chancharles92: The above code looks good. Maybe setup a `msg2 = [33u8; 36]` and add one more line at the bottom:
`assert!(pk.verify(&msg2, &sig, CS_ID_BLS_BN254).is_err());`

philippecamacho: Ok, done in a07c2293.

chancharles92 (approved):

Issue #215, opened 2023-03-21
Erasure code, RS code


ggutoski:
Use something that already exists. eg. https://github.com/rust-rse/reed-solomon-erasure

PR #225, opened 2023-03-31
Reed solomon erasure code

mrain:
## Description

closes: #215 

A very naive implementation of Reed Solomon erasure code.
 * Encoding is a very naive polynomial evaluation. It could be accelerated by NTT/FFT which may be already in arkworks repo (PrimeField is also a FftField in ark_ff).
 * Decoding is also a very naive Lagrange interpolation. I don't find an easy way to optimize it yet.

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/primitives/src/erasure_code/mod.rs
+++ b/primitives/src/erasure_code/mod.rs
@@ -1,0 +1,41 @@
+// Copyright (c) 2022 Espresso Systems (espressosys.com)
+// This file is part of the Jellyfish library.
+
+// You should have received a copy of the MIT License
+// along with the Jellyfish library. If not, see <https://mit-license.org/>.
+
+//! Module for erasure code
+
+use crate::errors::PrimitivesError;
+use ark_ff::Field;
+use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
+use ark_std::{fmt::Debug, vec::Vec};
+
+pub mod reed_solomon_erasure;
+
+/// Erasure code trait
+pub trait ErasureCode: Sized {
+    /// Associated field of the erasure code
+    type Field: Field;
+    /// Type for each data shards
+    type Shards: Debug
+        + Clone
+        + Eq
+        + PartialEq
+        + Sync
+        + Send
+        + CanonicalSerialize
+        + CanonicalDeserialize;
+
+    /// Create a new instance
+    ///  * data_len: the message length, and the minimum number of shards
+    ///    required for reconstruction
+    ///  * num_shards: the block (codeword) length
+    fn new(data_len: usize, num_shards: usize) -> Result<Self, PrimitivesError>;
+
+    /// Encoding
+    fn encode(&self, data: &[Self::Field]) -> Result<Vec<Self::Shards>, PrimitivesError>;
+
+    /// Decoding
+    fn decode(&self, shards: &[Self::Shards]) -> Result<Vec<Self::Field>, PrimitivesError>;
+}
--- a/primitives/src/erasure_code/reed_solomon_erasure.rs
+++ b/primitives/src/erasure_code/reed_solomon_erasure.rs
@@ -1,0 +1,189 @@
+// Copyright (c) 2022 Espresso Systems (espressosys.com)
+// This file is part of the Jellyfish library.
+
+// You should have received a copy of the MIT License
+// along with the Jellyfish library. If not, see <https://mit-license.org/>.
+
+//! Module for Reed Solomon Erasure Code
+
+use crate::errors::PrimitivesError;
+use ark_ff::PrimeField;
+use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
+use ark_std::{string::ToString, vec, vec::Vec};
+use core::marker::PhantomData;
+
+use super::ErasureCode;
+
+/// Very naive implementation of Reed Solomon erasure code.
+///  * data_len: the message length, and the minimum number of shards required
+///    for reconstruction
+///  * num_shards: the block (codeword) length
+/// The encoding of a message is the evaluation on (1..num_shards) of the
+/// polynomial whose coefficients are the message entries. Decoding is a naive
+/// Lagrange interpolation.
+pub struct ReedSolomonErasureCode<F: PrimeField> {
+    data_len: usize,
+    num_shards: usize,
+    phantom_f: PhantomData<F>,
+}
+
+/// Shards for Reed Solomon erasure code
+#[derive(Clone, Eq, Debug, CanonicalSerialize, CanonicalDeserialize, Derivative)]
+#[derivative(PartialEq, Hash(bound = "F: PrimeField"))]
+pub struct ReedSolomonErasureCodeShard<F: PrimeField> {
+    /// Index of shard shard
+    pub index: usize,
+    /// Value of this shard
+    pub value: F,
+    phantom_f: PhantomData<F>,
+}
+
+impl<F: PrimeField> ReedSolomonErasureCodeShard<F> {
+    /// Create a new shard
+    pub fn new(index: usize, value: F) -> Self {
+        Self {
+            index,
+            value,
+            phantom_f: PhantomData,
+        }
+    }
+}
+
+impl<F: PrimeField> ErasureCode for ReedSolomonErasureCode<F> {
+    type Field = F;
+    type Shards = ReedSolomonErasureCodeShard<F>;
+
+    fn new(data_len: usize, num_shards: usize) -> Result<Self, PrimitivesError> {
+        if data_len > num_shards {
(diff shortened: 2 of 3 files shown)


primitives/src/erasure_code/reed_solomon_erasure.rs
```diff
@@ -0,0 +1,218 @@
...
+        Ok(result)
+    }
+
+    /// Lagrange interpolation
+    /// Given a list of points (x_1, y_1) ... (x_n, y_n)
+    ///  1. Define l(x) = \prod (x - x_i)
+    ///  2. Calculate the barycentric weight w_i = \prod_{j \neq i} 1 / (x_i -
+    /// x_j)  
+    ///  3. Calculate l_i(x) = w_i * l(x) / (x - x_i)
+    ///  4. Return f(x) = \sum_i y_i * l_i(x)
```
ggutoski, line 101: I'm confused. `encode` takes an arbitrary size `data` and partitions it into multiple polynomials of degree `reconstruction_size`. Yet this comment suggests that `decode` returns only a _single_ polynomial `f` of degree `reconstruction_size`. What happened to all the other polynomials?

primitives/src/erasure_code/reed_solomon_erasure.rs
```diff
@@ -0,0 +1,218 @@
...
+                    vals.push(val);
+                }
+                ReedSolomonErasureCodeShard::new(i, vals)
+            })
+            .collect::<Vec<_>>();
+        Ok(result)
+    }
+
+    /// Lagrange interpolation
+    /// Given a list of points (x_1, y_1) ... (x_n, y_n)
```
ggutoski, line 96: The input is `&[Self::Shard]` which is a 2D array, yet this comment suggests that the input defines only a _single_ polynomial.

primitives/src/erasure_code/reed_solomon_erasure.rs
```diff
@@ -0,0 +1,218 @@
...
+
+        // Test for extra long input
+        let data = vec![F::from(1u64); 10];
+        let code = rs.encode(&data).unwrap();
+        assert_eq!(code[0].values.len(), 5);
+        let decode = rs.decode(&code).unwrap();
+        assert_eq!(data, decode);
+
+        // Extra long input test 2
+        let mut data = vec![F::from(1u64); 11];
```
ggutoski, line 204: should have a test with coeffs different from 1

ggutoski (dismissed):
API doc could be improved but otherwise ok for now.

primitives/src/erasure_code/reed_solomon_erasure.rs
```diff
@@ -0,0 +1,218 @@
...
+                phantom_f: PhantomData,
+            })
+        }
+    }
+
+    /// The encoding will split the data into chunks of length
+    /// `reconstruction_size`. And represent each chunk as a polynomial. The
+    /// codeword composes of evaluations of those polynomials on
+    /// (1..num_shards).
+    fn encode(&self, data: &[Self::Field]) -> Result<Vec<Self::Shard>, PrimitivesError> {
```
ggutoski, line 69: Should we use a `Polynomial` trait here instead of just slice of `Field`? [polynomial trait](https://github.com/EspressoSystems/jellyfish/blob/31ca8d9b585d9de55c0bda7c3bc1795543a49c84/primitives/src/pcs/mod.rs#L38) ?

ggutoski (commented):
a couple more questions

primitives/src/erasure_code/reed_solomon_erasure.rs
```diff
@@ -0,0 +1,218 @@
...
+pub struct ReedSolomonErasureCode<F: PrimeField> {
+    reconstruction_size: usize,
+    num_shards: usize,
+    phantom_f: PhantomData<F>,
+}
+
+/// Shards for Reed Solomon erasure code
+#[derive(Clone, Eq, Debug, CanonicalSerialize, CanonicalDeserialize, Derivative)]
+#[derivative(PartialEq, Hash(bound = "F: PrimeField"))]
+pub struct ReedSolomonErasureCodeShard<F: PrimeField> {
```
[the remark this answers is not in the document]
mrain: You are right. It only requires `Field` which provides multiplicative inverse.
Meanwhile, Feel free to do whatever you want here (code modification, etc) to make a functioning VID implementation. I can check into their implementation when I'm back.

ggutoski:
Now that I've had a chance to think about this PR wrt the VID use case: The level of abstraction here is too high. Encoding/decoding should not automatically split/recombine data into chunks of length `reconstruction_size`. (I need to do this again anyway in VID.)

I have a good idea of the API I want. Let's chat next week.

ggutoski:
Ready for review @mrain @chancharles92 .

TODO/Questions:
1. Proposal: change `new()` args to be `data_size` and `parity_size` so that `num_shards = data_size + parity_size` and it's impossible to give invalid args to `new()`. Any thoughts?
2. Shall we encode in coordinate form or coefficient form as per discussion?

primitives/src/erasure_code/mod.rs
```diff
@@ -7,34 +7,19 @@
...
-        + CanonicalSerialize
-        + CanonicalDeserialize;
-
-    /// Create a new instance
-    ///  * `reconstruction_size`: and the minimum number of shards required for
-    ///    reconstruction
-    ///  * `num_shards`: the block (codeword) length
-    /// TODO should we allow constructors at the trait level?
-    fn new(reconstruction_size: usize, num_shards: usize) -> Result<Self, PrimitivesError>;
+    type Shard;
```
mrain, line 18: I prefer to keep these trait bounds.

primitives/src/erasure_code/reed_solomon_erasure.rs
```diff
@@ -48,6 +48,8 @@ where
     ///  * `reconstruction_size`: and the minimum number of shards required for
     ///    reconstruction
     ///  * `num_shards`: the block (codeword) length
+    /// TODO instead use `data_size`, `parity_size` so that num_shards =
```
mrain, line 51: Good suggestion. We should take it.

ggutoski:
Current erasure code API looks like this:
```rust
fn encode(&self, data: &[F])
fn decode(&self, shards: &[Self::Shard])
```
where data size and codeword size are implicit and set by the constructor `new()`.

Problem: This API is stateful. Also, there's nothing to stop a user from taking shards from one erasure code and passing them to another (with different data and codeword sizes), so we don't get much argument safety anyway.

Proposal: make `encode` and `decode` associated functions (ie. don't take `&self`):
```rust
fn encode(parity_size: usize, data: &[F]) // data size is data.len()
fn decode(data_size: usize, shards: &[Self::Shard]) // parity size is shards.len()
```
@mrain @chancharles92 any opinions?

primitives/src/erasure_code/mod.rs
ggutoski: As discussed offline, we'll keep the trait bounds due to issues discussed here:
- https://github.com/rust-lang/rust/issues/20671
- https://justin.restivo.me/posts/2022-10-16-dealing-with-unwieldy-generics.html

primitives/src/erasure_code/reed_solomon_erasure.rs
```diff
@@ -0,0 +1,218 @@
...
+use core::marker::PhantomData;
+
+use super::ErasureCode;
+
+/// Very naive implementation of Reed Solomon erasure code.
+///  * `reconstruction_size`: and the minimum number of shards required for
+///    reconstruction
+///  * `num_shards`: the block (codeword) length
+/// The encoding of a message is the evaluation on (1..num_shards) of the
+/// polynomial whose coefficients are the message entries. Decoding is a naive
```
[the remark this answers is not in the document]
ggutoski: As discussed offline, we'll keep data in coefficient form because:
1. It's easier
2. We need the data in coefficient form anyway for use in VID. Thus, evaluation form would require 2 FFTs in `encode`, which is unacceptable in a performance-critical section

primitives/src/erasure_code/mod.rs
```diff
@@ -0,0 +1,35 @@
...
+        + Sync
+        + Send
+        + CanonicalSerialize
+        + CanonicalDeserialize;
+
+    /// Encoding
+    fn encode(data: &[T], parity_size: usize) -> Result<Vec<Self::Share>, PrimitivesError>;
+
+    /// Decoding
+    fn decode(shares: &[Self::Share]) -> Result<Vec<T>, PrimitivesError>;
```
mrain, line 34: I still prefer to include `data_size` as a parameter here. As it reminds users to keep track of it and prevents users from making mistakes.

ggutoski (approved):

Issue #235, opened 2023-04-15
ErasureCode encode/decode should accept IntoIterator args


ggutoski:
Currently `ErasureCode::decode` is declared to accept a slice:
https://github.com/EspressoSystems/jellyfish/blob/c730140b7abac367090bcad80ae4d34e0917243f/primitives/src/erasure_code/mod.rs#L35
Suppose I want to call `decode(shares)` but for `shares` I have only a `&[T]` where `T` is some super-struct that has a member `s: Self::Share`. In order to produce a `&[Self::Share]` suitable for use in `decode` I need to allocate a new slice and clone each `s` into it like so
```rust
let shares: Vec<_> = shares.iter().map(|sh| sh.s.clone()).collect();
```
See for example:
https://github.com/EspressoSystems/hotshot-primitives/blob/fd9a66a3bd3bafacf8534f4a51d49342ee8fe895/src/vid/advz.rs#L208-L212
Instead `decode` could be abstracted to this:
https://github.com/EspressoSystems/jellyfish/blob/9655ad2af4d83660a4b5bb9759a2e72ef92f1dd9/primitives/src/erasure_code/mod.rs#L41-L45
Now instead of passing a slice I can pass an iterator
```rust
shares.iter().map(|sh| &sh.s)
```
This saves a copy of the entire slice!

Why the trait bound `ExactSizeIterator + Clone`? Because `ExactSizeIterator` has a `len()` method, just like a slice. `Clone` is needed if `decode` needs to make more than one pass through the iterator (which it does).

All the above applies to `ErasureCode::encode`, too.

PR #236, opened 2023-04-15
Accept IntoIterator args to save memory

ggutoski:
## Description

closes: #235 

This is not a priority but it's Friday night and I'm having fun so I did it anyway. 😛 

## Tests

No change in functionality. Successful use of new API here https://github.com/EspressoSystems/hotshot-primitives/blob/97e2fe3c955b233e9c622c258f2570ab220b3790/src/vid/advz.rs#L207-L210

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/primitives/src/erasure_code/mod.rs
+++ b/primitives/src/erasure_code/mod.rs
@@ -6,6 +6,8 @@
 
 //! Module for erasure code
 
+use core::borrow::Borrow;
+
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 use ark_std::{fmt::Debug, vec::Vec};
 
@@ -28,9 +30,17 @@ pub trait ErasureCode<T> {
         + CanonicalDeserialize;
 
     /// Encode `data` into `data.len() + parity_size` shares.
-    fn encode(data: &[T], parity_size: usize) -> Result<Vec<Self::Share>, PrimitivesError>;
+    fn encode<D>(data: D, parity_size: usize) -> Result<Vec<Self::Share>, PrimitivesError>
+    where
+        D: IntoIterator,
+        D::Item: Borrow<T>,
+        D::IntoIter: ExactSizeIterator + Clone;
 
     /// Decode `shares` into `data_size` data elements.
     /// Return `Result::Err` if `shares.len() < data_size`.
-    fn decode(shares: &[Self::Share], data_size: usize) -> Result<Vec<T>, PrimitivesError>;
+    fn decode<S>(shares: S, data_size: usize) -> Result<Vec<T>, PrimitivesError>
+    where
+        S: IntoIterator,
+        S::Item: Borrow<Self::Share>,
+        S::IntoIter: ExactSizeIterator + Clone;
 }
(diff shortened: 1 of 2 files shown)


mrain (approved):
LGTM

primitives/src/erasure_code/mod.rs
```diff
@@ -28,9 +30,17 @@ pub trait ErasureCode<T> {
         + CanonicalDeserialize;
 
     /// Encode `data` into `data.len() + parity_size` shares.
-    fn encode(data: &[T], parity_size: usize) -> Result<Vec<Self::Share>, PrimitivesError>;
+    fn encode<D>(data: D, parity_size: usize) -> Result<Vec<Self::Share>, PrimitivesError>
+    where
+        D: IntoIterator,
+        D::Item: Borrow<T>,
+        D::IntoIter: ExactSizeIterator + Clone;
```
alxiong, line 38: May I ask if we can simply accept an `Iterator` instead?

My thinking is as follows:
either we expect the caller of this `encode` to create an iterator and pass in that iter. (this is the example @ggutoski gave in the issue https://github.com/EspressoSystems/jellyfish/issues/235)
or we let the caller to pass in that super-struct `T` directly and expect `T: IntoIter`. 

```rust
// upstream
struct MyCollection { shares: T }
impl<T> IntoIter<Item: Borrow<T>> for MyCollection {}

decode(my_collection); 
// downstream
fn encode(data: D) where D: IntoIterator, D::Item: Borrow<T>
```

or 

```rust
// upstream
struct MyStruct { shares: T }
let my_struct_vec: Vec<MyStruct> = ...;

decode(my_struct_vec.iter().map(|d| &d.s)); 

// downstream
fn encode(data: I) where Iterator<Item: Borrow<T>>
```

ggutoski: > May I ask if we can simply accept an `Iterator` instead?

From [stackoverflow](https://stackoverflow.com/questions/57543399/how-to-properly-pass-iterators-to-a-function-in-rust/57545430#57545430):

> Write functions that accept `impl IntoIterator<...>`. Because all Iterators implement `IntoIterator`, this is strictly more general than a function that accepts only `impl Iterator<...>`.

If we accept only `Iterator` the the caller is forced to give us an iterator. So, if the caller already has a collection `my_collection` of type `&[T]` (or any other collection type) then he can't just pass `my_collection` but must instead call `my_collection.iter()` first.

By accepting `IntoIterator` the function works regardless of whether we're given a collection or a custom iterator. We allow the use case described in #235 without changing the syntax for callers who just expect to give us a slice.

Not sure I understand your code snippets, but both use cases should work fine when `decode` accepts `IntoIterator`.

alxiong: yeah, agree that `IntoIterator` is a more relaxed/general API, which should be preferred.

ggutoski: Note that tests that call `decode` did not even need to change to accommodate this new API.

alxiong (approved):

PR #241, opened 2023-04-19
fix: bytes_from_field_elements remove lower bound check on result_len

ggutoski:
## Description

Currently `bytes_from_field_elements` enforces that the result length be within the final field element.

However, some use cases (such as polynomial interpolation) might pad the list of field elements with extra zeros. In this case result length will be below the final field element, so `bytes_from_field_elements` will throw an error.

We could rely on the caller to remember how many field elements it started with and trim appropriately, but that cuts against the purpose of encoding the result length in the field elements. It makes more sense to simply allow the user to pad with extra zeros and then truncate down to the encoded result length at the end.

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/utilities/src/conversion.rs
+++ b/utilities/src/conversion.rs
@@ -178,14 +178,15 @@ where
     let field_chunk_len = primefield_chunk_len * extension_degree;
     let result_capacity = elems.len() * field_chunk_len;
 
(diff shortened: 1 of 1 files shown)


mrain (approved):
LGTM

PR #242, opened 2023-04-19
Minor changes to support BLS signature verification inside a smart contract

philippecamacho:
This PR is related to https://github.com/EspressoSystems/espresso-sequencer/pull/292. The changes are:
* Add a public function to have access to the attribute of the `Signature` struct.
* Change the logic of hash_to_curve  about computing the square root of a field element so that it is consistent with https://github.com/EspressoSystems/cape/blob/77f343849db3d9e721c6e3d0f7108155c178dbee/contracts/contracts/libraries/BN254.sol#L338

### DIFF
--- a/primitives/src/signatures/bls_over_bn254.rs
+++ b/primitives/src/signatures/bls_over_bn254.rs
@@ -272,7 +272,8 @@ pub struct KeyPair {
 #[derive(CanonicalSerialize, CanonicalDeserialize, Eq, Clone, Debug)]
 #[allow(non_snake_case)]
 pub struct Signature {
-    pub(crate) sigma: G1Projective,
+    /// The signature is a G1 group element.
+    pub sigma: G1Projective,
 }
 
 impl Hash for Signature {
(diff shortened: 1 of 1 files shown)


chancharles92 (approved):
LGTM

alxiong (commented):
May I ask:
- why only use `BigIntger`, not `MontFp`, any incompatibility there?
- maybe in the future, just mark `pub sigma` instead of having this getter with a mediocre fn name.

@philippecamacho

philippecamacho:
> May I ask:
> 
> * why only use `BigIntger`, not `MontFp`, any incompatibility there?

The tests pass both with `MontFp!("3")` and `BaseField::from(3)`, but I think `MontFp!` is more low level as it is specific to how the field operations are implemented. Indeed I went into issues when working on https://github.com/EspressoSystems/espresso-sequencer/issues/205 as I was comparing the canonical (solidity) representation of a field element with the Montgomery one which is different.

> * maybe in the future, just mark `pub sigma` instead of having this getter with a mediocre fn name.
> 

Well, I first made the `sigma` attribute public and then changed my mind... Made a small [PR](https://github.com/EspressoSystems/jellyfish/pull/245) to fix this.

PR #250, opened 2023-04-25
make `bytes_from_field_elements` infallible with paranoid checks for overflow

ggutoski:
## Description

- `bytes_from_field_elements` error type is `String`, which does not implement the `Error` trait. Instead, make it infallible. Because it's infallible we need paranoid checks for overflow.
- Improve tests for `bytes_[to|from]_field_elements`.
- Minor quality-of-life improvement for downstream error handling: `PCSError` does not implement the `Error` trait -> implement it
- Appease clippy way over in `glv.rs`.

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/relation/src/gadgets/ecc/glv.rs
+++ b/relation/src/gadgets/ecc/glv.rs
@@ -94,7 +94,8 @@ where
         self.check_var_bound(scalar)?;
         self.check_point_var_bound(base)?;
 
-        let (s1_var, s2_var, s2_sign_var) = scalar_decomposition_gate::<_, P, _>(self, &scalar)?;
+        let (s1_var, s2_var, s2_sign_var) =
+            scalar_decomposition_gate::<P::BaseField, P::ScalarField>(self, &scalar)?;
 
         let endo_base_var = endomorphism_circuit::<_, P>(self, base)?;
         multi_scalar_mul_circuit::<_, P>(self, base, s1_var, &endo_base_var, s2_var, s2_sign_var)
@@ -264,13 +265,13 @@ macro_rules! int_to_fq {
 // Return the variables for k1 and k2
 // and sign bit for k2.
 #[allow(clippy::type_complexity)]
-fn scalar_decomposition_gate<F, P, S>(
+fn scalar_decomposition_gate<F, S>(
     circuit: &mut PlonkCircuit<F>,
     s_var: &Variable,
 ) -> Result<(Variable, Variable, BoolVar), CircuitError>
 where
     F: PrimeField,
-    P: TECurveConfig<BaseField = F, ScalarField = S>,
+    // P: TECurveConfig<BaseField = F, ScalarField = S>,
     S: PrimeField,
 {
     // the order of scalar field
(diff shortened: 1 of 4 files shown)


utilities/Cargo.toml
```diff
@@ -8,6 +8,7 @@ license = { workspace = true }
 rust-version = { workspace = true }
 
 [dependencies]
+anyhow = "1.0"
```
mrain, line 11: consider adding `default-features = false` here for no_std support.

mrain (commented):
Generally LGTM. Minor comments above and may also consider fixing the clippy error in untouched file.

utilities/Cargo.toml
ggutoski: I made `bytes_from_field_elements` infallible instead. Don't need `anyhow` dependency anymore so I removed it.

relation/src/gadgets/ecc/glv.rs
```diff
@@ -264,13 +265,13 @@ macro_rules! int_to_fq {
...
 #[allow(clippy::type_complexity)]
-fn scalar_decomposition_gate<F, P, S>(
+fn scalar_decomposition_gate<F, S>(
     circuit: &mut PlonkCircuit<F>,
     s_var: &Variable,
 ) -> Result<(Variable, Variable, BoolVar), CircuitError>
 where
     F: PrimeField,
-    P: TECurveConfig<BaseField = F, ScalarField = S>,
+    // P: TECurveConfig<BaseField = F, ScalarField = S>,
```
mrain, line 274: Remove this commented line.

mrain (commented):
Looks **paranoidly** good to me with a minor comment. 😆

mrain (approved):

utilities/src/conversion.rs
```diff
@@ -95,59 +93,94 @@ where
...
+
+    if bytes.as_ref().is_empty() {
+        return Vec::new();
+    }
+
+    // various quantities
+    let primefield_bytes_len = usize::try_from((F::BasePrimeField::MODULUS_BIT_SIZE - 1) / 8)
+        .expect("prime field modulus byte len should fit into usize");
+    let extension_degree =
+        usize::try_from(F::extension_degree()).expect("extension degree should fit into usize");
```
alxiong, line 121: these checks seem a little redundant.
I guess "infallible" really means a success-or-nothing guarantee, this code is panic-able anyway, `usize.try_from().expect()` is too verbose than just `xx as usize` which is the same effect, correct?

utilities/src/conversion.rs
```diff
@@ -95,59 +93,94 @@ where
...
+            primefield_elems.len(),
+            extension_degree,
+            "prime field elems len {} differs from extension degree {}",
+            primefield_elems.len(),
+            extension_degree
+        );
+        result.push(
+            F::from_base_prime_field_elems(&primefield_elems)
+                .expect("field elem construction should succeed"),
+        );
```
alxiong, line 166: here's a more elegant way of writing this (better double-check):

```rust
    for field_elem_bytes in bytes.as_ref().chunks(field_byte_len) {
        let primefield_elems = field_elem_bytes
            .chunks(primefield_byte_len)
            .map(F::BasePrimeField::from_le_bytes_mod_order)
            .chain(std::iter::repeat(F::BasePrimeField::ZERO).take(extension_degree - field_elem_bytes.len() / primefield_byte_len))
            .collect::<Vec<_>>();
        result.push(
            F::from_base_prime_field_elems(&primefield_elems)
                .expect("failed to construct field element"),
        );
    }
```

utilities/src/conversion.rs
ggutoski: `xx as usize` is not the same effect.
[Operator expressions - The Rust Reference](https://doc.rust-lang.org/reference/expressions/operator-expr.html#semantics):
> Casting from a larger integer to a smaller integer (e.g. u32 -> u8) will truncate

We want it to panic in the event of overflow.

Issue #252, opened 2023-05-10
cross compilation to wasm32 for blst


alxiong:
In task of #248, I encountered difficulties in cross-compiling `blst` crate into the `wasm32-unknown-unknown` target.

As time of writing, its `master` branch (let alone `v0.3.10` which is what we originally depends on) is not `no_std` compliant. Thanks to the work https://github.com/supranational/blst/pull/150 and alternative approach by blst's maintainer: https://github.com/supranational/blst/pull/153, it's getting closer to no_std. 
Unfortunately, as seen in the discussion in the latter PR, there's disagreement on how to achieve `no_std` and the feature flag designs, and personally I favor the former whose APIs and Rust patterns are more idiomatic, thus I created our own fork directly porting their work (full credits to Nazar and Eric) here: https://github.com/EspressoSystems/blst/tree/no-std 

### Native host ✅ 

I'm running a MacOS with m1 chip, so basically `aarch64-darwin` and `clang` from default homebrew's llvm toolchains.

Now, outside nix environment, I'm able to compile that fork of blst using: `AR=$(which llvm-ar) CC=$(which clang) RUSTFLAGS="-C target-cpu=generic" cargo build --target wasm32-unknown-unknown --no-default-features`
- I need explicit env var to ensure the correct choice of clang (this is most likely due to the build script is looking for `$CC` which I didn't set by default in my `.profile`)
- Since I'm running m1, the default `target-cpu=apple-m13` which will cause emission of "'apple-m1' is not a recognized processor for this target (ignoring processor)" warning and some other errors. Thus, setting it to `generic` should be a good idea.

### Inside nix-shell ❌

I got lots of help from the team, but still, it seems a bit of a hustle to switch to different clang and llvm tools and cxx headers easily. (or maybe we are not doing it correctly)

We have tried:
- https://nixos.wiki/wiki/Using_Clang_instead_of_GCC 
- upgrading clang (from 11.10 to 15) inside `flake.nix` using this syntax [here](https://discourse.nixos.org/t/how-to-override-stdenv-for-all-packages-in-mkshell/10368/16)

Several problems remains:
- still need to add `export CC=clang_15` in `shellHook` to point to the right clang (the default `clang` still points to the 11.1.0 version)
- with `flake.nix` like:
  ```nix
          devShell = mkShell {
            buildInputs = [
              llvmPackages_15.clang
              llvmPackages_15.libstdcxxClang
              llvm_15
            ];
           shellHook = ''
[...]

Issue #248, opened 2023-04-25
Make jellyfish wasm compatibility


alxiong:
This issue aims to achieve wasm compatibility:

- [ ] finish #111 
- [ ] document slowdown (if any) and limitations in the wasm version. 
- [ ] add wasm compilation check in CI

alxiong:
While doing #318, on my local M1, 

native benchmark:
```
$ cargo flamegraph --root --bench=plonk-benches --features test-srs

proving time for Bls12_381, PlonkType::TurboPlonk: 52120 ns/gate
proving time for Bls12_377, PlonkType::TurboPlonk: 53111 ns/gate
proving time for Bn254, PlonkType::TurboPlonk: 44123 ns/gate
proving time for BW6_761, PlonkType::TurboPlonk: 215765 ns/gate
proving time for Bls12_381, PlonkType::UltraPlonk: 79688 ns/gate
proving time for Bls12_377, PlonkType::UltraPlonk: 78137 ns/gate
proving time for Bn254, PlonkType::UltraPlonk: 69546 ns/gate
proving time for BW6_761, PlonkType::UltraPlonk: 341935 ns/gate
verifying time for Bls12_381, PlonkType::TurboPlonk: 2504441 ns
verifying time for Bls12_377, PlonkType::TurboPlonk: 2597979 ns
verifying time for Bn254, PlonkType::TurboPlonk: 3948429 ns
verifying time for BW6_761, PlonkType::TurboPlonk: 8686679 ns
verifying time for Bls12_381, PlonkType::UltraPlonk: 2658983 ns
verifying time for Bls12_377, PlonkType::UltraPlonk: 2570262 ns
verifying time for Bn254, PlonkType::UltraPlonk: 1644525 ns
verifying time for BW6_761, PlonkType::UltraPlonk: 9560779 ns
batch verifying time for Bls12_381, PlonkType::TurboPlonk, 1000 proofs: 15505 ns/proof
batch verifying time for Bls12_377, PlonkType::TurboPlonk, 1000 proofs: 16257 ns/proof
batch verifying time for Bn254, PlonkType::TurboPlonk, 1000 proofs: 16841 ns/proof
batch verifying time for BW6_761, PlonkType::TurboPlonk, 1000 proofs: 31831 ns/proof
batch verifying time for Bls12_381, PlonkType::UltraPlonk, 1000 proofs: 17502 ns/proof
[...]

PR #254, opened 2023-05-11
Initial WASM Support

alxiong:
## Description

Closes: #252 
Partially address: #248

---

write a little note why.

- [x] ~Wrote unit tests~

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -69,7 +69,9 @@ jobs:
       - name: Check no_std support and WASM compilation
         env:
           RUSTFLAGS: -C target-cpu=generic
-        run: cargo build --target wasm32-unknown-unknown --no-default-features
+        run: |
+          cargo check --no-default-features
+          cargo build --target wasm32-unknown-unknown --no-default-features
 
       - name: Test
         run: bash ./scripts/run_tests.sh
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,7 +32,7 @@ jobs:
           override: true
           default: true
           components: rustfmt, clippy
-          targets: |
+          target: |
             wasm32-unknown-unknown
 
       - name: Install nightly toolchain
@@ -66,8 +66,10 @@ jobs:
       - name: Check all tests and binaries compilation
         run: cargo check --workspace --tests --lib --bins
 
-      - name: Check no_std compilation
-        run: bash ./scripts/build_wasm.sh
+      - name: Check no_std support and WASM compilation
+        env:
+          RUSTFLAGS: -C target-cpu=generic
+        run: cargo build --target wasm32-unknown-unknown --no-default-features
 
       - name: Test
         run: bash ./scripts/run_tests.sh
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,6 +32,8 @@ jobs:
           override: true
           default: true
           components: rustfmt, clippy
+          targets: |
+            wasm32-unknown-unknown
 
       - name: Install nightly toolchain
         uses: actions-rs/toolchain@v1
@@ -65,7 +67,7 @@ jobs:
         run: cargo check --workspace --tests --lib --bins
 
       - name: Check no_std compilation
-        run: cargo check --workspace --lib --no-default-features
+        run: bash ./scripts/build_wasm.sh
 
       - name: Test
         run: bash ./scripts/run_tests.sh
--- a/primitives/Cargo.toml
+++ b/primitives/Cargo.toml
@@ -12,7 +12,7 @@ ark-bls12-377 = "0.4.0"
 ark-bls12-381 = "0.4.0"
 ark-bn254 = "0.4.0"
 ark-bw6-761 = "0.4.0"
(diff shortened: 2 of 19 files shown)


.github/workflows/build.yml
```diff
@@ -64,8 +66,10 @@ jobs:
       - name: Check all tests and binaries compilation
         run: cargo check --workspace --tests --lib --bins
 
-      - name: Check no_std compilation
-        run: cargo check --workspace --lib --no-default-features
+      - name: Check no_std support and WASM compilation
+        env:
+          RUSTFLAGS: -C target-cpu=generic
+        run: cargo build --target wasm32-unknown-unknown --no-default-features
```
nyospe, line 72: Do you think it would be beneficial to still keep a second check for non-wasm `no-std` support, just in case?

primitives/benches/pcs.rs
```diff
@@ -36,7 +36,7 @@ mod bench {
                 10
             };
 
-            let poly = Arc::new(DenseMultilinearExtension::rand(nv, &mut rng));
+            let poly = Rc::new(DenseMultilinearExtension::rand(nv, &mut rng));
```
nyospe, line 39: single threaded benchmarks only?

primitives/src/pcs/multilinear_kzg/mod.rs
```diff
@@ -28,9 +28,9 @@ use ark_std::{
     end_timer, format,
     marker::PhantomData,
     rand::{CryptoRng, RngCore},
+    rc::Rc,
```
nyospe, line 31: All of this is guaranteed to only be accessed within a single thread?

nyospe: Okay, never mind, I get it. This stems from `merge_polynomials` taking `&[Self::Polynomial]`, and Rust not really having a good way to gracefully handle `&[&T]` for reentrant operations.

Support for variadic generics would be very nice here. 😒 

It would seem `Rc` is, in fact, correct.

primitives/Cargo.toml
```diff
@@ -12,7 +12,7 @@ ark-bls12-377 = "0.4.0"
 ark-bls12-381 = "0.4.0"
 ark-bn254 = "0.4.0"
 ark-bw6-761 = "0.4.0"
-ark-crypto-primitives = { version = "0.4.0", features = ["sponge"] }
+ark-crypto-primitives = { version = "0.4.0", default-features = false, features = ["sponge"] }
```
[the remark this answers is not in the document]
alxiong: good point, I will add them

.github/workflows/build.yml
alxiong: yes, good idea. will add.

primitives/src/pcs/multilinear_kzg/mod.rs
alxiong: I have to switch to this to compile due to missing `CanonicalSerde` mentioned here: https://github.com/arkworks-rs/algebra/issues/649.

It's a temporary (and wrong) decision to use `Rc`, `Arc` doesn't have CanonicalSerde in `no_std` mode at all, so I will have to remove the trait bounds in `pcs/mod.rs::trait PolynomicalCommitmentScheme::Polynomial`; but for `Rc` only the CanonicalDeserialize is missing, but at least it's available in `no_std` mode, thus I can still compile it.

For WASM env, the "multi-threading" seems to only support message-passing concurrency, but not shared reference concurrency, I need to change this code, not even `Rc` would work.

alxiong: > This stems from merge_polynomials taking &[Self::Polynomial], and Rust not really having a good way to gracefully handle &[&T] for reentrant operations.

the actual reason was simpler than this, your point is something I didn't consider btw.
Originally we use `Arc` and it was fine (on non-wasm platforms). I'm not sure about the "reentrant op" you mentioned. 

Now, I'm leaning towards, in `no_std` I just pass in owned value itself, whereas otherwise, I'll just use `Arc`

alxiong: see if you like [d1e8c3b](https://github.com/EspressoSystems/jellyfish/pull/254/commits/d1e8c3b1667052923932283cb68a3788c7252c42) @nyospe

primitives/benches/pcs.rs
alxiong: changed back in [d1e8c3b](https://github.com/EspressoSystems/jellyfish/pull/254/commits/d1e8c3b1667052923932283cb68a3788c7252c42)

primitives/src/pcs/multilinear_kzg/mod.rs
nyospe: That'll work. Somewhere in the back of my brain, I think I have the glimmering of a better solution, but it's being stubborn, and this works for now, at the cost of some potentially avoidable copies.

The _correct_ solution is a variadic function, but in Rust, that requires a macro, and that's a PitA, and borderline impossible in a trait method, so let's go with this.

nyospe (approved):
Yeah, LGTM now.

PR #257, opened 2023-05-15
fix cargo udeps failure in ci

alxiong:
## Description

Fixing periodic check on unused dep in CI: https://github.com/EspressoSystems/jellyfish/actions/runs/4969545780

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,7 +77,7 @@ jobs:
         run: bash ./scripts/run_tests.sh
 
       - name: Example
-        run: cargo run --release --example proof_of_exp --features test-srs
+        run: cargo run --release --example proof-of-exp --features test-srs
 
       - name: Generate Documentation
         run: |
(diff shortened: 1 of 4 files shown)


mrain (commented):
Is it better to add a required feature in the example?
Like [here](https://github.com/EspressoSystems/jellyfish/pull/256/files#diff-4336bdd5f194d1847c08e50879d78c7fb123197d41577baa9206771cc1b4ed0bR56).

Ancient123 (approved):

Ancient123:
Seems fine to me. But @mrain has a good idea as well.

alxiong:
> Is it better to add a required feature in the example?

done in [944b771](https://github.com/EspressoSystems/jellyfish/pull/257/commits/944b77112b6a61767f6f4086ffe2e0f6e928f97d)
(thanks for the suggestion, I didn't know this feature before)

mrain (approved):
LGTM

PR #259, opened 2023-05-16
Nix fixups

sveitser:
- Remove overlay, it seems it's not required with the current configuration.
- Set env vars from nix variables, this should be a bit more robust.
- Use `libclang.version` in include path.

## Description

closes: #XXXX

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -23,22 +23,6 @@
       let
         overlays = [ 
           (import rust-overlay)
-          (self: super: {
-            rustc = (super.rustc.override {
-                stdenv = self.stdenv.override {
-                    targetPlatform = super.stdenv.targetPlatform // {
-                        parsed = {
-                            cpu = { name = "wasm32"; };
-                            vendor = {name = "unknown";};
-                            kernel = {name = "unknown";};
-                            abi = {name = "unknown";};
-                        };
-                    };
-                };
-            }).overrideAttrs (attrs: {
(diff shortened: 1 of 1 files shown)


DieracDelta (approved):
LGTM!

alxiong:
oh this is interesting, so the wasm target comes by default with the `rust-overlay`? 
(@sveitser how did you even find out?)

sveitser:
I'm not sure. I had a vague memory that when I first tright I wasn't using clang v15 but something older, so figured I would try without the overlay.

DieracDelta:
I'm confused. It looks like the wasm target still needs to be specified? Or do you mean that `clang` has a wasm backend?

alxiong:
inside the nix-shell, if I run `rustc --print target-list | grep wasm` I got:
```
wasm32-unknown-emscripten
wasm32-unknown-unknown
wasm32-wasi
wasm64-unknown-unknown
```

So, it seems that by default, `rust-overlay` already includes these targets as [shown in `targets.nix`](https://github.com/oxalica/rust-overlay/blob/65c3f2655f52a81e1b3e629d4c07df4873d0f2bb/manifests/targets.nix)

make sense!

DieracDelta:
Ah, I see what you mean! This make sense, seems like it [matches the docs](https://doc.rust-lang.org/rustc/targets/built-in.html). The asterisk is that `core` for each target isn't included unless specified. I guess that's what the `targets` attribute is doing.

Issue #258, opened 2023-05-15
choosing u32 or u64 backends for x25519


alxiong:
As part of work in #254, I choose `crypto_box = { version = "0.8.1", default-features = false, features = ["alloc", "u32_backend"] }` as default since we want to compiled down to wasm32 (for browser support)

Meanwhile, as requested by @mike1729, we should support choosing `u64_backend` as well. But as suggested in this comment: https://github.com/RustCrypto/nacl-compat/issues/85#issuecomment-1399502973, the best action now is to wait until `x25519-dalek` and `crypto_box`'s next release to completely remove these two feature flags, and configure using `cfg` (env var) from top level to avoid the additive feature flag problem.

pmikolajczyk41:
do you have any estimates on when next `x25519-dalek` release will be ready/published?

pmikolajczyk41:
if I'm not mistaken, their 4.0.0-rc.2 is already without these backend features

alxiong:
you are right, @pmikolajczyk41  we can point to `crypto_box-v0.9.0-rc.0` right now it seems. (as seen [here](https://github.com/RustCrypto/nacl-compat/blob/crypto_box-v0.9.0-rc.0/crypto_box/Cargo.toml#L23))

I will create a quick PR for this.

PR #270, opened 2023-05-25
remove crypto_box and support choosing curve25519 backend using rustflags

alxiong:
# Description

closes: #258

(see more discussion in https://github.com/RustCrypto/nacl-compat/issues/124)

Changes include:
- switch the underlying AEAD from `crypto_box` to `crypto_kx` + `chacha20poly1305` without changing Jellyfish's own API
  - the latest `crypto_box` no longer supports associated data since it strive to restore compatibility with libsodium
  - we want the latest `crypto_box` because we want to use `--cfg curve25519_dalek_backend="u32_backend"` to select backend instead of feature flag which is additive, (see the linked issue for details)
  - therefore, I decide to move away from it, and just use `crypto_kx` to establish shared secret, and use symmetric AEAD like `xchacha20poly1305` (with extended nonce), see the RustCrypto issue linked above for more details
- Removed `CanonicalSerde` and directly use `serde::Serde` for our structs. (cuz I don't see any benefits of adding the extra layer of indirection here)
- Remove one unit test named `fn test_conversion()`
  - previously, when using `crypto_box`, any random bytes are guaranteed to be converted back and forth into key structs, but now with `crypto_kx::Keypair` which is essentially `curve25519` key pair, there's some canonical formatting, some potential mod modulus operations etc. Thus there's no point to have that unit test any more. 

---

write a little note why.

- [x] ~Wrote unit tests~

cc @pmikolajczyk41

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,7 +70,7 @@ jobs:
 
       - name: Check no_std support and WASM compilation
         env:
-          RUSTFLAGS: '-C target-cpu=generic --cfg curve25519_dalek_backend="u32_backend"'
+          RUSTFLAGS: '-C target-cpu=generic --cfg curve25519_dalek_backend="u32"'
         run: |
           cargo check --no-default-features
           cargo build --target wasm32-unknown-unknown --no-default-features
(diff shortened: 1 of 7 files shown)


mrain (approved):
LGTM. 👍

Issue #272, opened 2023-05-25
`MerkleTree::verify` should not take `self`


ggutoski:
`MerkleTree::verify` currently takes `&self`: 

https://github.com/EspressoSystems/jellyfish/blob/3eb332acfb40de36677b9ae9dbaed0b0a899d0c1/primitives/src/merkle_tree/macros.rs#L98-L102

This means that the caller needs the entire merkle tree just to verify a proof.

Looking at the code, `self` is barely used at all. We could eliminate the need for `self` if we instead include `self.root.value()` inside the `MerkleProof` struct.

ggutoski:
The new member added to `MerkleProof` needs serde help:
```rust
    #[serde(with = "canonical")]
    pub root: T,
```

PR #274, opened 2023-05-26
feat(api)!: MerkleTreeScheme::verify do not take self, pos

ggutoski:
## Description

closes: #272 

API change:

- Convert `MerkleTreeScheme::verify` from method to associated function. (ie. do not take `self`.)
- `MerkleTreeScheme::verify` do not take `pos`. Why? This argument is redundant. Currently, the only use of `pos` is to compare it against `proof.pos` and throw an error if they don't match. By removing `pos` we do not force the user to know anything about `pos` in order to verify a proof.
- **[EDIT]** `MerkleTreeScheme::verify` take `root` arg as per https://github.com/EspressoSystems/jellyfish/pull/274#discussion_r1206884154
- **[EDIT]** New `VerificationResult` return type, used in `MerkleTreeScheme::verify` as per https://github.com/EspressoSystems/jellyfish/pull/274#discussion_r1206866522
- Core API changes in `mod.rs`, `macros.rs`, `internal.rs`. All other modified files merely reflect these API changes.

---

write a little note why.

- [ ] Wrote unit tests (existing unit tests updated to reflect API change) 

### DIFF
--- a/primitives/src/merkle_tree/internal.rs
+++ b/primitives/src/merkle_tree/internal.rs
@@ -137,9 +137,6 @@ where
     pub pos: I,
     /// Nodes of proof path, from root to leaf
     pub proof: MerklePath<E, I, T>,
-    /// Root
-    #[serde(with = "canonical")]
-    pub root: T,
 
     /// Place holder for Arity
     _phantom_arity: PhantomData<Arity>,
@@ -156,11 +153,10 @@ where
         self.proof.len()
     }
 
-    pub fn new(pos: I, proof: MerklePath<E, I, T>, root: T) -> Self {
+    pub fn new(pos: I, proof: MerklePath<E, I, T>) -> Self {
         MerkleProof {
             pos,
             proof,
-            root,
             _phantom_arity: PhantomData,
         }
     }
--- a/primitives/src/merkle_tree/internal.rs
+++ b/primitives/src/merkle_tree/internal.rs
@@ -9,7 +9,7 @@ use core::{marker::PhantomData, ops::AddAssign};
 use super::{
     DigestAlgorithm, Element, Index, LookupResult, MerkleCommitment, NodeValue, ToTraversalPath,
 };
-use crate::errors::PrimitivesError;
+use crate::errors::{PrimitivesError, VerificationResult};
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 use ark_std::{
     borrow::Borrow, boxed::Box, format, iter::Peekable, string::ToString, vec, vec::Vec,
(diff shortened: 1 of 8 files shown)


primitives/src/merkle_tree/mod.rs
```diff
@@ -220,15 +220,10 @@ pub trait MerkleTreeScheme: Sized {
...
-    /// * `pos` - zero-based index of the leaf in the tree
     /// * `proof` - a merkle tree proof
     /// * `returns` - Ok(true) if the proof is accepted, Ok(false) if not. Err()
     ///   if the proof is not well structured, E.g. not for this merkle tree.
-    fn verify(
-        &self,
-        pos: impl Borrow<Self::Index>,
-        proof: impl Borrow<Self::MembershipProof>,
-    ) -> Result<bool, PrimitivesError>;
+    fn verify(proof: impl Borrow<Self::MembershipProof>) -> Result<bool, PrimitivesError>;
```
mrain, line 226: Should we change the return type to `Result<(), PrimitivesError>` as well?

mrain: How about this `pub type VerificationResult = Result<(), ()>`? Just to keep the code cleaner.

primitives/src/merkle_tree/internal.rs
```diff
@@ -137,6 +137,9 @@ where
     pub pos: I,
     /// Nodes of proof path, from root to leaf
     pub proof: MerklePath<E, I, T>,
+    /// Root
+    #[serde(with = "canonical")]
+    pub root: T,
```
mrain, line 142: I feel like the root value shouldn't be included in the Merkle proof structure.
The usual scenario is, party A would compare its locally maintained root value against the root value computed from party B's provided Merkle proof. It's weird if they are bind together.

ggutoski: Makes sense. So I guess instead we should modify `verify` to take a new `root: T` argument? I'll do that.

ggutoski: Done in https://github.com/EspressoSystems/jellyfish/pull/274/commits/57b35014d08d069d9a08a8566b146d93eb9a637c, https://github.com/EspressoSystems/jellyfish/pull/274/commits/2fb45633c60f09d733f5884201352886787cc858

primitives/src/merkle_tree/mod.rs
ggutoski: Done in https://github.com/EspressoSystems/jellyfish/pull/274/commits/37da39a17bb95b23c29c07b3bcd028069e1f2fe4, https://github.com/EspressoSystems/jellyfish/pull/274/commits/27d5e8e962271761852e5b905b1f88f9390daf78

PR #287, opened 2023-06-01
fix: invalid pos is not en error, so return Ok(_) instead of Err(_)

ggutoski:
## Description

one-line addendum to #286 

`MerkleTreeScheme::verify` should not return `Err(_)` if the `pos` arg is inconsistent with the `proof`. It just means that `proof` is invalid for this `pos`, so verification fails. This matters for downstream users who use `?` to return early from a real internal error before checking whether verification fails.

---

write a little note why.

- [ ] Wrote unit tests


mrain (approved):
LGTM

PR #290, opened 2023-06-01
NMT trait and hash function integration

nomaxg:
Introduces the NMT trait and a namespaced hasher

Close https://github.com/EspressoSystems/espresso-sequencer/issues/365

### DIFF
--- a/primitives/src/merkle_tree/namespaced_merkle_tree.rs
+++ b/primitives/src/merkle_tree/namespaced_merkle_tree.rs
@@ -247,21 +247,28 @@ mod nmt_tests {
 
     #[test]
     fn test_namespaced_hash() {
-        let first_leaf = Leaf::new(1);
-        let second_leaf = Leaf::new(5);
+        let num_leaves = 5;
+        let leaves: Vec<Leaf> = (0..num_leaves).map(|i| Leaf::new(i)).collect();
 
         //  Ensure that leaves are digested correctly
-        let first_hash = Hasher::digest_leaf(&0, &first_leaf);
-        let second_hash = Hasher::digest_leaf(&1, &second_leaf);
-        assert_eq!((first_hash.min_namespace, first_hash.max_namespace), (1, 1));
+        let mut hashes: Vec<NamespacedHash<Sha3Node, u64>> = leaves
+            .iter()
+            .enumerate()
+            .map(|(idx, leaf)| Hasher::digest_leaf(&(idx as u64), leaf))
+            .collect();
+        assert_eq!((hashes[0].min_namespace, hashes[0].max_namespace), (0, 0));
 
         // Ensure that sorted internal nodes are digested correctly
-        let hash = Hasher::digest(&[first_hash, second_hash]);
-        assert_eq!((hash.min_namespace, hash.max_namespace), (1, 5));
+        let hash = Hasher::digest(&hashes);
+        assert_eq!(
+            (hash.min_namespace, hash.max_namespace),
+            (0, num_leaves - 1)
+        );
 
         // Ensure that digest errors when internal nodes are not sorted by namespace
         // Digest will turn a result when https://github.com/EspressoSystems/jellyfish/issues/275 is addressed
-        let res = catch_unwind(|| Hasher::digest(&[second_hash, first_hash]));
+        hashes[0] = hashes[hashes.len() - 1];
+        let res = catch_unwind(|| Hasher::digest(&hashes));
         assert!(res.is_err());
     }
 }
--- a/primitives/src/merkle_tree/namespaced_merkle_tree.rs
+++ b/primitives/src/merkle_tree/namespaced_merkle_tree.rs
@@ -17,7 +17,7 @@ use super::{AppendableMerkleTreeScheme, DigestAlgorithm, Element, Index, NodeVal
 /// Namespaced Merkle Tree where leaves are sorted by a namespace identifier.
 pub trait NamespacedMerkleTreeScheme: AppendableMerkleTreeScheme
 where
-    Self::Element: Namespaced<Self::NamespaceId>,
+    Self::Element: Namespaced,
 {
     /// Namespace proof type
     type NamespaceProof: Clone;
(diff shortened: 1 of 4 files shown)


primitives/src/merkle_tree/namespaced_merkle_tree.rs
```diff
@@ -0,0 +1,262 @@
...
+}
+
+/// Trait indicating that a leaf has a namespace
+pub trait Namespaced<N: Namespace> {
+    /// Returns the namespace of the leaf
+    fn get_namespace(&self) -> N;
+}
+
+/// Trait indicating that a digest algorithm can bind namespaces
+pub trait BindNamespace<E, I, T, N>: DigestAlgorithm<E, I, T>
```
nomaxg, line 68: Not sure this is the best name - we need this trait so that we can generate a hash that binds an internal hash to its namespace range

primitives/src/merkle_tree/namespaced_merkle_tree.rs
```diff
@@ -0,0 +1,262 @@
...
+
+impl<T, N> Default for NamespacedHash<T, N>
+where
+    N: Namespace,
+    T: NodeValue,
+{
+    fn default() -> Self {
+        Self {
+            hash: T::default(),
+            max_namespace: <N as Namespace>::min(),
```
nomaxg, line 129: Min/max is swapped like this so that hashing a namespace hash `NamespaceHash::default()` preserves the namespace range

primitives/src/merkle_tree/namespaced_merkle_tree.rs
```diff
@@ -0,0 +1,262 @@
...
+{
+    phantom1: PhantomData<H>,
+    phantom2: PhantomData<E>,
+    phantom3: PhantomData<I>,
+    phantom4: PhantomData<T>,
+    phantom5: PhantomData<N>,
+}
+
+/// Trait indicating that a leaf has a namespace
+pub trait Namespaced<N: Namespace> {
```
nomaxg, line 62: I'll change this to an associated type

primitives/src/merkle_tree/namespaced_merkle_tree.rs
```diff
@@ -0,0 +1,262 @@
...
+
+impl<E, H, T, I, N> DigestAlgorithm<E, I, NamespacedHash<T, N>> for NamespacedHasher<H, E, I, T, N>
+where
+    E: Element + Namespaced<N>,
+    I: Index,
+    N: Namespace,
+    T: NodeValue,
+    H: DigestAlgorithm<E, I, T> + BindNamespace<E, I, T, N>,
+{
+    fn digest(data: &[NamespacedHash<T, N>]) -> NamespacedHash<T, N> {
```
philippecamacho, line 158: We assume the nodes are already namespaced sorted?

primitives/src/merkle_tree/namespaced_merkle_tree.rs
```diff
@@ -0,0 +1,262 @@
...
+        if data.is_empty() {
+            return NamespacedHash::default();
+        }
+        let first_node = data[0];
+        let min_namespace = first_node.min_namespace;
+        let mut max_namespace = first_node.max_namespace;
+        let mut nodes = vec![H::generate_namespaced_commitment(first_node)];
+        for node in data {
+            // Ensure that namespaced nodes are sorted
+            if node.min_namespace < max_namespace {
```
philippecamacho, line 168: Should not it be `node.min_namespace > max_namespace`.?

primitives/src/merkle_tree/namespaced_merkle_tree.rs
philippecamacho: Looks like this is the case. Maybe add a comment.

primitives/src/merkle_tree/namespaced_merkle_tree.rs
```diff
@@ -0,0 +1,262 @@
...
+{
+    fn digest(data: &[NamespacedHash<T, N>]) -> NamespacedHash<T, N> {
+        if data.is_empty() {
+            return NamespacedHash::default();
+        }
+        let first_node = data[0];
+        let min_namespace = first_node.min_namespace;
+        let mut max_namespace = first_node.max_namespace;
+        let mut nodes = vec![H::generate_namespaced_commitment(first_node)];
+        for node in data {
```
philippecamacho, line 166: Should not we skip the first element in the loop? That is:  `for node in data.skip(1)` ?

primitives/src/merkle_tree/namespaced_merkle_tree.rs
```diff
@@ -0,0 +1,262 @@
...
+        }
+    }
+
+    impl<E, I, N> BindNamespace<E, I, Sha3Node, N> for Sha3Digest
+    where
+        E: Element + CanonicalSerialize,
+        I: Index,
+        N: Namespace,
+    {
+        fn generate_namespaced_commitment(
```
philippecamacho, line 227: As discussed, in practice we should be careful about how we hash min,max and content together. Maybe keeping min and max value without hashing them is fine.

primitives/src/merkle_tree/namespaced_merkle_tree.rs
nomaxg: I am actually not sure about this - removed this for now but will revisit it when we start implementing proofs

primitives/src/merkle_tree/namespaced_merkle_tree.rs
nomaxg: If they aren’t hashed then I believe a dishonest prover can modify namespace ranges and return incomplete namespaces

primitives/src/merkle_tree/namespaced_merkle_tree.rs
nomaxg: added!

primitives/src/merkle_tree/namespaced_merkle_tree.rs
philippecamacho: What I meant is that we keep bot min and max value as part of the node, which means they are conceptually hashed, the hash being `(min,max,Hash(...) )` . The output of this hash is larger but it is still a (collision-resistant) hash.

primitives/src/merkle_tree/namespaced_merkle_tree.rs
philippecamacho: The name does not looks bad to me.

PR #295, opened 2023-06-06
Update run_tests.sh for cargo version upgrade

mrain:
## Description

Dealing with failure in CI: such as https://github.com/EspressoSystems/jellyfish/actions/runs/5184434049/jobs/9343327682.

See discussion on [Zulip](https://espresso.zulipchat.com/#narrow/stream/311693-Jellyfish/topic/.E2.9C.94.20CI.20failure/near/363127797).

### DIFF
--- a/scripts/run_tests.sh
+++ b/scripts/run_tests.sh
@@ -4,7 +4,7 @@ set -e
 # We want the code to panic if there is an integer overflow
 export RUSTFLAGS="-C overflow-checks=on"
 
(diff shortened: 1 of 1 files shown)


alxiong (approved):
The original intention of having both stable and nightly toolchain is using stable to compile, and only use nightly for formatting. 

And using nightly for tests should also be fine, since we already check correct compilation in earlier steps via `cargo check --workspace --all-features`. 

I'm good with this.

Issue #275, opened 2023-05-26
Change the return type for DigestAlgorithm


mrain:
Wrap the return type with `Result`.

https://github.com/EspressoSystems/jellyfish/blob/57b09836c79181a8f3a5f36cefa33e705a5e628b/primitives/src/merkle_tree/mod.rs#L130 and https://github.com/EspressoSystems/jellyfish/blob/57b09836c79181a8f3a5f36cefa33e705a5e628b/primitives/src/merkle_tree/mod.rs#L133

PR #299, opened 2023-06-06
DigestAlgorithm API small refactor

mrain:
## Description

closes: #275

`DigestAlgorithm` now returns a `Result` type.

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/primitives/src/merkle_tree/internal.rs
+++ b/primitives/src/merkle_tree/internal.rs
@@ -203,20 +203,20 @@ where
                 let children = chunk
                     .map(|(pos, elem)| {
                         let pos = I::from(pos as u64);
-                        Box::new(MerkleNode::Leaf {
-                            value: H::digest_leaf(&pos, elem.borrow()),
+                        Ok(Box::new(MerkleNode::Leaf {
+                            value: H::digest_leaf(&pos, elem.borrow())?,
                             pos,
                             elem: elem.borrow().clone(),
-                        })
+                        }))
                     })
-                    .pad_using(Arity::to_usize(), |_| Box::new(MerkleNode::Empty))
-                    .collect_vec();
-                Box::new(MerkleNode::<E, I, T>::Branch {
-                    value: digest_branch::<E, H, I, T>(&children),
+                    .pad_using(Arity::to_usize(), |_| Ok(Box::new(MerkleNode::Empty)))
+                    .collect::<Result<Vec<_>, PrimitivesError>>()?;
+                Ok(Box::new(MerkleNode::<E, I, T>::Branch {
+                    value: digest_branch::<E, H, I, T>(&children)?,
                     children,
-                })
+                }))
             })
-            .collect_vec();
+            .collect::<Result<Vec<_>, PrimitivesError>>()?;
         for _ in 1..height {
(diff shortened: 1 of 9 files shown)


primitives/src/merkle_tree/internal.rs
```diff
@@ -327,15 +327,17 @@ where
...
 where
     E: Element,
     H: DigestAlgorithm<E, I, T>,
     I: Index,
     T: NodeValue,
 {
     // Question(Chengyu): any more efficient implementation?
-    let data = data.iter().map(|node| node.value()).collect_vec();
+    let data = data.iter().map(|node| node.value()).collect::<Vec<_>>();
     H::digest(&data)
```
ggutoski, line 341: [out of scope for this PR] Regarding the question: we could change the `DigestAlgorithm` trait so that `digest` takes an `IntoIterator` instead of `&[T]`. Then you could just pass `data.iter().map(|node| node.value())` directly to `digest` with no mem copy or heap alloc.

primitives/src/merkle_tree/universal_merkle_tree.rs
```diff
@@ -44,7 +44,8 @@ where
         let traversal_path = pos.to_traversal_path(self.height);
         let ret = self
             .root
-            .update_internal::<H, Arity>(self.height, pos, &traversal_path, elem);
+            .update_internal::<H, Arity>(self.height, pos, &traversal_path, elem)
+            .expect("Internal error caused by hash failure.");
```
ggutoski, line 48: Is there a reason not to return `Result` here, too?

ggutoski (commented):
minor comments only

primitives/src/merkle_tree/internal.rs
mrain: Good idea. Let me do it

mrain: A tiny issue: `RescueCRHF::sponge_no_padding` is taking a slice of field elements `&[F]`. We still need some memcopy/heap alloc right now. Of course we can change that API in the future.

Do you think it's still worth changing it now?

ggutoski: Up to you. It's not high priority. If you don't feel like doing it now then please post a github issue and tag it "low priority" so we don't forget about this.

I have a similar issue with 

https://github.com/EspressoSystems/jellyfish/blob/6641b3ff97bf12d764d7384bc632cdc84c3f8209/utilities/src/conversion.rs#L128

It should take `IntoIterator` as input and return an iterator as output.

I guess we should both post an issue. 😉

ggutoski: I think perhaps we should move toward a design pattern of taking `IntoIterator` as input and returning an iterator as output whenever possible. @alxiong any opinion?

nomaxg (approved):
thanks!

mrain:
Merged with namespace Merkle tree from the main branch.
Please checking again @nomaxg @ggutoski

ggutoski (approved):
lgtm. don't forget to merge with main again.

ggutoski (approved):

Issue #140, opened 2022-11-18
Batch (non)membership proof for merkle tree


mrain:

nomaxg:
This will compress the size of NMT proofs, which currently rely on a separate merkle proof for every namespace leaf

PR #305, opened 2023-06-07
Basic NMT implementation

nomaxg:
## Description

A basic NMT implementation that wraps a `MerkleTree` and stores namespace range metadata to generate proofs. Right now, the proof just verifies that each leaf in the namespace has the correct namespace id and is part of the tree. I will address namespace completeness in a follow-up PR. 
---

write a little note why.

- [x ] Targeted PR against correct branch (main)
- [ x] Linked to GitHub issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [x ] Wrote unit tests
- [x ] Updated relevant documentation in the code

### DIFF
--- a/primitives/src/merkle_tree/namespaced_merkle_tree.rs
+++ b/primitives/src/merkle_tree/namespaced_merkle_tree.rs
@@ -8,11 +8,16 @@
 use alloc::vec;
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 use ark_std::vec::Vec;
-use core::{fmt::Debug, hash::Hash, marker::PhantomData};
+use core::{borrow::Borrow, fmt::Debug, hash::Hash, marker::PhantomData, ops::Range};
+use hashbrown::{hash_map::Entry, HashMap};
+use typenum::Unsigned;
 
-use crate::errors::PrimitivesError;
+use crate::errors::{PrimitivesError, VerificationResult};
 
-use super::{AppendableMerkleTreeScheme, DigestAlgorithm, Element, Index, NodeValue};
+use super::{
+    append_only::MerkleTree, AppendableMerkleTreeScheme, DigestAlgorithm, Element, Index,
+    LookupResult, MerkleCommitment, MerkleTreeScheme, NodeValue,
+};
 
 /// Namespaced Merkle Tree where leaves are sorted by a namespace identifier.
 /// The data structure supports namespace inclusion proofs.
@@ -27,15 +32,12 @@ where
 
     /// Returns the entire set of leaves corresponding to a given namespace and
     /// a completeness proof.
-    fn get_namespace_leaves_and_proof(
-        &self,
-        namespace: Self::NamespaceId,
-    ) -> (Vec<Self::Element>, Self::NamespaceProof);
+    fn get_namespace_proof(&self, namespace: Self::NamespaceId) -> Self::NamespaceProof;
 
     /// Verifies the completeness proof for a given set of leaves and a
     /// namespace.
     fn verify_namespace_proof(
-        leaves: &[Self::Element],
+        &self,
         proof: Self::NamespaceProof,
         namespace: Self::NamespaceId,
     ) -> Result<(), PrimitivesError>;
@@ -44,6 +46,7 @@ where
 /// NamespacedHasher wraps a standard hash function (implementer of
 /// DigestAlgorithm), turning it into a hash function that tags internal nodes
 /// with namespace ranges.
+#[derive(Debug)]
 pub struct NamespacedHasher<H, E, I, T, N>
 where
     H: DigestAlgorithm<E, I, T>,
@@ -156,9 +159,15 @@ where
         let mut max_namespace = first_node.max_namespace;
         let mut nodes = vec![H::generate_namespaced_commitment(first_node)];
         for node in &data[1..] {
+            if node == &NamespacedHash::default() {
+                continue;
+            }
             // Ensure that namespaced nodes are sorted
             if node.min_namespace < max_namespace {
-                panic!("leaves are out of order")
+                panic!(
+                    "leaves are out of order: Max namespace is {:?} but candidate range is {:?}:{:?}",
+                    max_namespace, node.min_namespace, node.max_namespace
+                );
             }
             max_namespace = node.max_namespace;
             nodes.push(H::generate_namespaced_commitment(*node));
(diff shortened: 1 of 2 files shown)


primitives/src/merkle_tree/namespaced_merkle_tree.rs
```diff
@@ -181,10 +188,387 @@ where
...
+    E: Element + Namespaced<Namespace = N>,
+    T: NodeValue,
+    N: Namespace,
+    Arity: Unsigned,
+{
+    fn extend(
+        &mut self,
+        elems: impl IntoIterator<Item = impl core::borrow::Borrow<Self::Element>>,
+    ) -> Result<(), PrimitivesError> {
+        // TODO: update namespace metadata
```
mrain, line 302: Tracking issue for these TODOs?

primitives/src/merkle_tree/namespaced_merkle_tree.rs
```diff
@@ -181,10 +188,387 @@ where
...
+        self.inner.extend(elems)
+    }
+
+    fn push(
+        &mut self,
+        elem: impl core::borrow::Borrow<Self::Element>,
+    ) -> Result<(), PrimitivesError> {
+        // TODO: update namespace metadata
```
mrain, line 310: Same as above.

primitives/src/merkle_tree/namespaced_merkle_tree.rs
```diff
@@ -181,10 +188,387 @@ where
...
+        self.inner.push(elem)
+    }
+}
+
+#[derive(Clone, Debug)]
+enum NamespaceProofType {
```
mrain, line 316: Better document this type.

primitives/src/merkle_tree/namespaced_merkle_tree.rs
```diff
@@ -20,29 +26,27 @@ where
     Self::Element: Namespaced,
 {
     /// Namespace proof type
-    type NamespaceProof: Clone;
+    type NamespaceProof;
     /// Namespace type
     type NamespaceId: Namespace;
 
     /// Returns the entire set of leaves corresponding to a given namespace and
```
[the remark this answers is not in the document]
nomaxg: yeah good idea I'll add a `NamespaceProof` interface

primitives/src/merkle_tree/namespaced_merkle_tree.rs
```diff
@@ -20,29 +26,27 @@ where
...
         proof: Self::NamespaceProof,
         namespace: Self::NamespaceId,
-    ) -> Result<(), PrimitivesError>;
+    ) -> Result<VerificationResult, PrimitivesError>;
 }
 
 /// NamespacedHasher wraps a standard hash function (implementer of
 /// DigestAlgorithm), turning it into a hash function that tags internal nodes
 /// with namespace ranges.
+#[derive(Debug, Clone)]
```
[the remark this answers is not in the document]
nomaxg: Not strictly necessary at the moment, but `Clone + Debug` tend to be reasonable defaults for any struct (especially `Debug` to help debug tests)

primitives/src/merkle_tree/namespaced_merkle_tree.rs
```diff
@@ -181,10 +188,297 @@ where
     }
 }
 
+type InnerTree<E, H, T, N, Arity> =
+    MerkleTree<E, NamespacedHasher<H, E, u64, T, N>, u64, Arity, NamespacedHash<T, N>>;
+
+#[derive(Debug, Clone)]
```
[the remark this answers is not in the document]
nomaxg: Yes but we'll probably implement a custom serialization scheme so that we don't send the whole tree over the wire, tracked here: https://github.com/orgs/EspressoSystems/projects/20/views/4?pane=issue&itemId=28910685

primitives/src/merkle_tree/namespaced_merkle_tree.rs
```diff
@@ -181,10 +188,387 @@ where
...
+        root: &NamespacedHash<T, N>,
+        namespace: N,
+    ) -> Result<VerificationResult, PrimitivesError> {
+        if let Some(boundary_proof) = self.left_boundary_proof.as_ref() {
+            // If there is a leaf to the left of the namespace range, check that it is less
+            // than the target namespace
+            if boundary_proof.elem().unwrap().get_namespace() >= namespace
+                || *boundary_proof.index() != self.indices[0] - 1
+            {
+                return Ok(Err(()));
```
philippecamacho, line 379: This is confusing. Why do we need to wrap the result into another result?

primitives/src/merkle_tree/namespaced_merkle_tree.rs
```diff
@@ -181,10 +188,387 @@ where
...
+        namespace: N,
+    ) -> Result<VerificationResult, PrimitivesError> {
+        if namespace < root.min_namespace || namespace > root.max_namespace {
+            // Easy case where the namespace isn't covered by the range of the tree root
+            return Ok(Ok(()));
+        } else {
+            // Harder case: Find an element whose namespace is greater than our
+            // target and show that the namespace to the left is less than our
+            // target
+            let left_proof = &self.left_boundary_proof.as_ref().unwrap();
```
jbearer, line 438: We should return an error if `left_boundary_proof` is `None`, not panic. The proof is untrusted data, the prover should not be able to make the verifier panic.

primitives/src/merkle_tree/namespaced_merkle_tree.rs
```diff
@@ -181,10 +188,387 @@ where
...
+            let right_proof = &self.proofs[0];
```
jbearer, line 439: Similar thing here, we shouldn't panic if `self.proofs` is empty

primitives/src/merkle_tree/namespaced_merkle_tree.rs
```diff
@@ -181,10 +188,387 @@ where
...
+    T: NodeValue,
+    H: DigestAlgorithm<E, u64, T> + BindNamespace<E, u64, T, N>,
+    N: Namespace,
+    Arity: Unsigned,
+{
+    proof_type: NamespaceProofType,
+    proofs: Vec<MerkleProof<E, u64, NamespacedHash<T, N>, Arity>>,
+    left_boundary_proof: Option<MerkleProof<E, u64, NamespacedHash<T, N>, Arity>>,
+    right_boundary_proof: Option<MerkleProof<E, u64, NamespacedHash<T, N>, Arity>>,
+    indices: Vec<u64>,
```
jbearer, line 335: I believe these are required to be sequential, and the same length as `proofs`? If so could we replace this with just the first index?

jbearer (changes requested):
We should clean up all the unwraps/panics/unchecked indexes, especially in the verifier -- a malicious prover shouldn't be able to make the verifier panic.

primitives/src/merkle_tree/namespaced_merkle_tree.rs
nomaxg: Actually, I think we can directly return `VerificationResult` here. @ggutoski correct me if I'm wrong, but I believe the nested pattern is used to distinguish user input error (incorrect serialization, incorrect parameters, etc.) from the actual proof verification result.

ggutoski: I agree it looks confusing at first glance but there is a reason for it. See discussion https://github.com/EspressoSystems/jellyfish/pull/274#discussion_r1206866522 and rustdoc:

https://github.com/EspressoSystems/jellyfish/blob/466a7604f00a6d5b142ae1b3b7aabcd1111f06df/primitives/src/errors.rs#L18-L23

The nested result pattern is well-described here: https://sled.rs/errors

I've been slowly migrating code to conform to this pattern, though no explicit declaration has been made to make it a policy for jellyfish code. We can discuss further if you're interested.

In this particular case, the method is infallible so technically we can return `VerificationResult` directly as @nomaxg suggests. Is this infallibility an explicit promise we wish to make forever? If not then perhaps we should keep the nested result return type for future-proofness.

nomaxg: Since this is a helper function, I think its ok to return `VerificationResult` directly. However, the outer `verify` will need the nested result since it makes a call to the internal merkle `verify` fn.

Personally, I'm a fan of this pattern because it gives a nice separation between different classes of errors.

primitives/src/merkle_tree/namespaced_merkle_tree.rs
nomaxg: @ggutoski would it be more appropriate to return a `PrimitivesError` here or a falsey `VerificationResult`? The context is that the proof being verified is malformed.

primitives/src/merkle_tree/namespaced_merkle_tree.rs
philippecamacho: Thank @ggutoski @nomaxg , I was not aware of this error handling pattern. Makes sense.

primitives/src/merkle_tree/namespaced_merkle_tree.rs
```diff
@@ -181,10 +188,387 @@ where
...
+pub struct NamespaceProof<E, T, Arity, N, H>
+where
+    E: Element + Namespaced<Namespace = N>,
+    T: NodeValue,
+    H: DigestAlgorithm<E, u64, T> + BindNamespace<E, u64, T, N>,
+    N: Namespace,
+    Arity: Unsigned,
+{
+    proof_type: NamespaceProofType,
+    proofs: Vec<MerkleProof<E, u64, NamespacedHash<T, N>, Arity>>,
```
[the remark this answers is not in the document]
nomaxg: I added a note to this issue https://github.com/EspressoSystems/jellyfish/issues/140

primitives/src/merkle_tree/namespaced_merkle_tree.rs
```diff
@@ -181,10 +188,387 @@ where
...
+    /// Return the set of leaves associated with this Namespace proof as an
+    /// iterator
+    pub fn get_namespace_leaves(&self) -> impl Iterator<Item = &E> {
+        let num_leaves = match self.proof_type {
+            NamespaceProofType::Presence => self.proofs.len(),
+            NamespaceProofType::Absence => 0,
+        };
+        self.proofs
+            .iter()
+            .map(|proof| proof.elem().unwrap())
```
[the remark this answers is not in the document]
nomaxg: Assuming the namespace proof is valid, this unwrap should be safe (the proof checks that each elem/leaf is accounted for). We could verify this here as well, but I think that verifying the namespace proof before fetching the elements is a reasonable precondition.

primitives/src/merkle_tree/namespaced_merkle_tree.rs
nomaxg: added

primitives/src/merkle_tree/namespaced_merkle_tree.rs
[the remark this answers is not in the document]
nomaxg: done 👍

primitives/src/merkle_tree/namespaced_merkle_tree.rs
```diff
@@ -181,10 +209,460 @@ where
...
+/// set
+#[derive(Clone, Debug)]
+enum NamespaceProofType {
+    Presence,
+    Absence,
+}
+
+#[derive(Clone, Debug)]
+/// Namespace Proof
+pub struct NaiveNamespaceProof<E, T, Arity, N, H>
```
mrain, line 346: Should we also derive `CanonicalSerialize` and `CanonicalDeserialize` to this type?
We might also want `Eq`, `PartialEq`, etc.

nomaxg: yeah - I guess we can derive Serialize/Deserialize and have canonical serialize bounds for `E, T`?

jbearer: Not sure about `Canonical*` but we will definitely need serde `Serialize` and `Deserialize`, for the query service

primitives/src/merkle_tree/namespaced_merkle_tree.rs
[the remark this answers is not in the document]
jbearer: we could consider removing `get_namespace_leaves` (or making it private) and having `verify` return the leaves on success. So you have to verify in order to get the leaves

primitives/src/merkle_tree/namespaced_merkle_tree.rs
nomaxg: done 👍

primitives/src/merkle_tree/namespaced_merkle_tree.rs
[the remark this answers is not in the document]
nomaxg: good idea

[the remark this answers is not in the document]
nomaxg: Looks this would require refactoring `VerificationResult` from a `Result<(),()>` to a `Result<T, ()>`, which is think is a reasonable improvement (cc. @ggutoski) but probably something to tackle in another PR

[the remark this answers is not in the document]
jbearer: Yes almost always a custom result type alias should be generic over the `Ok` type. But yeah we can save this for future work

jbearer (approved):
LGTM 👍

mrain (approved):

philippecamacho (approved):
LGMT

Issue #261, opened 2023-05-19
Reed-Solomon decoder accept FFT domain


mrain:
As requested by VID implementation.

ggutoski:
this issue ported from https://github.com/EspressoSystems/hotshot-primitives/issues/21

PR #309, opened 2023-06-08
feat: Reed-Solomon decoder accept FFT domain

ggutoski:
# Description

close: #261 

- Builds upon #256. Merge target is [`rs-code-fft`](https://github.com/EspressoSystems/jellyfish/tree/rs-code-fft) instead of `main` so you can see the stuff I added to #256. If desired we could change the merge target to `main` and close #256.
- House cleaning: change `reed_solomon_encode` into an iterator adaptor. (ie. It now returns an iterator.) It's not obvious/easy to do the same for `reed_solomon_decode` so I didn't do that.

# Warning

Naive implementation of `reed_solomon_erasure_decode_rou` computes each eval point in linear time. Ideally we would like constant-time random access to domain points without allocating and populating a vec for all the points. How best to achieve that?

https://github.com/EspressoSystems/jellyfish/blob/8baf5e0098088eab3c7fda2d60bb2b69b9aeb90e/primitives/src/reed_solomon_code/mod.rs#L157-L158

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/primitives/src/reed_solomon_code/mod.rs
+++ b/primitives/src/reed_solomon_code/mod.rs
@@ -12,10 +12,13 @@ use ark_poly::{EvaluationDomain, Radix2EvaluationDomain};
 use ark_std::{format, vec, vec::Vec};
 use core::borrow::Borrow;
 
-/// Encode `data` into `data.len() + parity_size` shares.
+/// Erasure-encode `data` into `data.len() + parity_size` shares.
+///
 /// Treating the input data as the coefficients of a polynomial,
 /// Returns the evaluations of this polynomial over [1, data.len() +
-/// parity_size]. If `F` is a `FftField`, the encoding can be done using FFT on
+/// parity_size].
+///
+/// If `F` is a [`FftField`], the encoding can be done using FFT on
 /// a `GeneralEvaluationDomain` (E.g. when num_shares = 3):
 /// ```
 /// use ark_poly::{EvaluationDomain, GeneralEvaluationDomain};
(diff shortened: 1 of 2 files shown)


primitives/src/reed_solomon_code/mod.rs
```diff
@@ -131,23 +136,51 @@ where
...
+where
+    F: FftField,
+    D: IntoIterator,
+    D::Item: Borrow<(usize, F)>,
+    D::IntoIter: ExactSizeIterator + Clone,
+{
+    let domain_shares = shares.into_iter().map(|share| {
+        let &(index, eval) = share.borrow();
+        // TODO(Gus) nth runtime is linear in index!
+        (domain.elements().nth(index).unwrap(), eval)
```
mrain, line 158: We might first collect all elements and then access them here?

mrain:
Generally LGTM. I think we can close the previous PR and directly merge this one to the main branch.

primitives/src/reed_solomon_code/mod.rs
ggutoski: I thought of that, and it might be the least bad solution. My initial worry is that most elements will never be accessed, so you'll allocate and compute a bunch of elements that are never used. eg. if the code rate is 1/4 then only 25% of elements will be accessed.

## Other ideas

- Another stupid idea is to sort the shares by index so that only a single pass through `elements` is needed, but sorting is ridiculous and you still compute all the elements.
- Another idea is to compute `nth` via log-time scalar multiplication. Still not great but perhaps better than scanning through all elements.

ggutoski: Actually the scaler mall solution is n log n as opposed to oven if we collect. No good answers here

mrain: I prefer the collection idea over n log n for the following reason:
- Code is cleaner and more understandable, and you don't have to touch the low level details of the underlying `EvaluationDomain`.
- Efficiency isn't much an issue here. It's only going to be called once during the decoding phase, and (even considering the memory allocation) the overhead here is insignificant compared to other part.
- The collected evaluation points might be useful in the faster RS decoding.

ggutoski: agreed. fixed in https://github.com/EspressoSystems/jellyfish/pull/309/commits/08e870f3ac01fa5fa6bf0f3a3be1d1de7f369b6a

PR #316, opened 2023-06-13
feat: Finish NMT Absence proofs

nomaxg:
## Description

Constructs an absence proof when the target namespace falls within the tree range by providing two adjacent leaves that bound the target namespace. 

---

write a little note why.

- [ x] Targeted PR against correct branch (main)
- [x ] Linked to GitHub issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [x ] Wrote unit tests
- [x ] Updated relevant documentation in the code

### DIFF
--- a/primitives/src/merkle_tree/namespaced_merkle_tree.rs
+++ b/primitives/src/merkle_tree/namespaced_merkle_tree.rs
@@ -652,8 +652,11 @@ where
             // i - 1
             if let Some((_, range)) = self.namespace_ranges.range(namespace..).next() {
                 let i = range.start;
-                left_boundary_proof = Some(self.lookup_proof(i - 1));
-                right_boundary_proof = Some(self.lookup_proof(i));
+                // If i == 0, the target namespace is less than the tree's minimum namespace
+                if i > 0 {
+                    left_boundary_proof = Some(self.lookup_proof(i - 1));
+                    right_boundary_proof = Some(self.lookup_proof(i));
+                }
             }
         }
         NaiveNamespaceProof {
@@ -827,7 +830,7 @@ mod nmt_tests {
             .is_err());
 
         // Check the simple absence proof case when the namespace falls outside of the
-        // tree range
+        // tree range (namespace > root.max_namespace)
         let absence_proof = tree.get_namespace_proof(last_ns + 1);
         let leaves: Vec<Leaf> = absence_proof
             .get_namespace_leaves()
@@ -836,6 +839,20 @@ mod nmt_tests {
             .collect();
         assert!(tree
             .verify_namespace_proof(&absence_proof, last_ns + 1)
+            .unwrap()
+            .is_ok());
+        assert_eq!(leaves, []);
+
+        // Check the simple absence proof case when the namespace falls outside of the
+        // tree range (namespace < root.min_namespace)
+        let absence_proof = tree.get_namespace_proof(first_ns - 1);
+        let leaves: Vec<Leaf> = absence_proof
+            .get_namespace_leaves()
+            .into_iter()
+            .cloned()
+            .collect();
+        assert!(tree
+            .verify_namespace_proof(&absence_proof, first_ns - 1)
             .unwrap()
             .is_ok());
         assert_eq!(leaves, []);
--- a/primitives/src/merkle_tree/namespaced_merkle_tree.rs
+++ b/primitives/src/merkle_tree/namespaced_merkle_tree.rs
@@ -650,8 +650,8 @@ where
             // namespace at some index i, prove that the
             // target namespace is empty by providing proofs of leaves at index i and
             // i - 1
-            if let Some((namespace, _)) = self.namespace_ranges.range(namespace..).next() {
-                let i = self.namespace_ranges.get(namespace).unwrap().start;
+            if let Some((_, range)) = self.namespace_ranges.range(namespace..).next() {
+                let i = range.start;
                 left_boundary_proof = Some(self.lookup_proof(i - 1));
                 right_boundary_proof = Some(self.lookup_proof(i));
             }
--- a/primitives/src/merkle_tree/namespaced_merkle_tree.rs
+++ b/primitives/src/merkle_tree/namespaced_merkle_tree.rs
@@ -5,11 +5,13 @@
 // along with the Jellyfish library. If not, see <https://mit-license.org/>.
 
 //! Implementation of a Namespaced Merkle Tree.
-use alloc::vec;
+use alloc::{
+    collections::{btree_map::Entry, BTreeMap},
+    vec,
+};
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 use ark_std::{string::ToString, vec::Vec};
 use core::{borrow::Borrow, fmt::Debug, hash::Hash, marker::PhantomData, ops::Range};
-use hashbrown::{hash_map::Entry, HashMap};
 use itertools::Itertools;
 use serde::{Deserialize, Serialize};
 use typenum::Unsigned;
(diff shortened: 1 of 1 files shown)


primitives/src/merkle_tree/namespaced_merkle_tree.rs
```diff
@@ -612,35 +634,35 @@ where
...
             }
         } else {
-            proof_type = NamespaceProofType::Absence
-            // TODO: This only handles the simple absence proof case where the
-            // namespace is outside of the root's namespace range
+            proof_type = NamespaceProofType::Absence;
+            let ranges: Vec<&N> = self.namespace_ranges.keys().collect();
+            let closest_ns_idx = ranges
+                .binary_search_by(|leaf_ns| (**leaf_ns).cmp(&namespace))
+                .expect_err("The namespace should not be in the range");
```
jbearer, line 652: Instead of collecting a BST into a vec and then searching the vec, consider using [`range`](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.range), e.g. `self.namespace_ranges.range(namespace..).next()` will give you 
* the entry for `namespace` if `namespace` is in the map (should not be possible in this case)
* the first entry greater than `namespace` if there is one
* `None` if `namespace` is greater than all the keys in the map

jbearer (approved):
Looks good, I had one take-it-or-leave-it comment on the use of BTreeMap

primitives/src/merkle_tree/namespaced_merkle_tree.rs
mrain: Suddenly I'm confused about the necessity of this `namesapce_ranges` as a `BTreeMap`. Isn't merkle tree itself already served as a similar data structure? Why should we have another BTreeMap.

nomaxg: @mrain the `BTreeMap` just enables efficient lookup of namespace range, we could do without it but then we'd need to search the leaves linearly to generate namespace proofs.

nomaxg: also @jbearer nice that looks cleaner, I'll try it out

mrain: But the namespace part of the Merkle tree already forms a [segment tree](https://en.wikipedia.org/wiki/Segment_tree), right? It does support efficient lookup, with similar complexity of `BTreeMap`.

nomaxg: I'm not sure I'm following, I guess with the leaves tagged as namespaces the Merkle tree implicitly represents a segment tree, but we still need to store the ranges somewhere right?

mrain: Not really. You don't have to store the ranges, but you can easily implements the [`range`](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.range) operation in time complexity `O(D)` where `D` is the tree height, usually in `log(N)` which is similar to `BTreeMap`.

Well, this can be an `improvement` since it reduces the storage overhead. We don't have to do it right now.

jbearer: I think the issue is that the ranges aren't actually stored in each internal node, they're computed dynamically from the leaves up when hashing. Is that right @nomaxg ?

nomaxg: Oh gotcha - I see what you're saying. We might still be better off because the complexity of the BTreeMap is in the # of namespaces (which is strictly less than N). But yeah, if each namespace is sparse I guess we would be adding extra storage overhead for a similar lookup.

primitives/src/merkle_tree/namespaced_merkle_tree.rs
```diff
@@ -646,20 +646,12 @@ where
...
             // target namespace is empty by providing proofs of leaves at index i and
             // i - 1
-            if closest_ns_idx > 0 && closest_ns_idx < ranges.len() {
-                let i = self
-                    .namespace_ranges
-                    .get(ranges[closest_ns_idx])
-                    .unwrap()
-                    .start;
+            if let Some((namespace, _)) = self.namespace_ranges.range(namespace..).next() {
+                let i = self.namespace_ranges.get(namespace).unwrap().start;
```
jbearer, line 654: Isn't `i` here the same as the second element of the entry, which is ignored above? If so we could grab it there and avoid this extra lookup/unwrap

primitives/src/merkle_tree/namespaced_merkle_tree.rs
mrain: Ahh @jbearer , IIUC the range information is stored in the `NamespacedHash`. There are 2 fields, `min_namespace` and `max_namespace`.

jbearer (approved):

primitives/src/merkle_tree/namespaced_merkle_tree.rs
```diff
@@ -825,5 +839,35 @@ mod nmt_tests {
             .unwrap()
             .is_ok());
         assert_eq!(leaves, []);
+
+        // Check absence proof case when the namespace falls inside of the tree range
+        let absence_proof = tree.get_namespace_proof(3);
```
philippecamacho, line 844: I tried to compute an absence proof for ns=first_ns -1=0 but I get an error. Is this normal?

```rust
let absence_proof = tree.get_namespace_proof(first_ns-1);

attempt to subtract with overflow
thread 'merkle_tree::namespaced_merkle_tree::nmt_tests::test_nmt' panicked at 'attempt to subtract with overflow', primitives/src/merkle_tree/namespaced_merkle_tree.rs:655:62
stack backtrace:
   0: rust_begin_unwind
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/std/src/panicking.rs:579:5
   1: core::panicking::panic_fmt
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/core/src/panicking.rs:64:14
   2: core::panicking::panic
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/core/src/panicking.rs:114:5
   3: <jf_primitives::merkle_tree::namespaced_merkle_tree::NMT<E,H,Arity,N,T> as jf_primitives::merkle_tree::namespaced_merkle_tree::NamespacedMerkleTreeScheme>::get_namespace_proof
             at ./src/merkle_tree/namespaced_merkle_tree.rs:655:62
   4: jf_primitives::merkle_tree::namespaced_merkle_tree::nmt_tests::test_nmt
             at ./src/merkle_tree/namespaced_merkle_tree.rs:831:29
   5: jf_primitives::merkle_tree::namespaced_merkle_tree::nmt_tests::test_nmt::{{closure}}
             at ./src/merkle_tree/namespaced_merkle_tree.rs:778:19
   6: core::ops::function::FnOnce::call_once
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/core/src/ops/function.rs:250:5
   7: core::ops::function::FnOnce::call_once
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

philippecamacho: Anyways it would be good to have a test for an absence proof of a namespace smaller than `root.min_namespace`.

nomaxg: Good catch - we need to ensure that `i > 0`

nomaxg: @philippecamacho fixed

philippecamacho (approved):
LGTM

Issue #314, opened 2023-06-12
Implement `extend` and `push` for NMT 


nomaxg:
We need to update the namespace metadata when we push leaves via these functions.

PR #321, opened 2023-06-14
feat: Implement `extend` + `push` for NMT

nomaxg:
…ng namespace data

## Description

Refactor logic that handles namespace metadata into a helper function that can be used for `from_elems`, `push`, and `extend`. 

closes: #314 

---

write a little note why.

- [ x] Targeted PR against correct branch (main)
- [ x] Linked to GitHub issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [ x] Wrote unit tests
- [ x] Updated relevant documentation in the code

### DIFF
--- a/primitives/src/merkle_tree/namespaced_merkle_tree/mod.rs
+++ b/primitives/src/merkle_tree/namespaced_merkle_tree/mod.rs
@@ -9,7 +9,7 @@ use alloc::collections::{btree_map::Entry, BTreeMap};
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 use ark_std::vec::Vec;
 use core::{borrow::Borrow, fmt::Debug, hash::Hash, marker::PhantomData, ops::Range};
-use serde::{Deserialize, Serialize};
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
 
 use typenum::Unsigned;
 
@@ -104,7 +104,7 @@ pub trait Namespace:
     + Hash
     + Ord
     + Serialize
-    + for<'a> Deserialize<'a>
+    + DeserializeOwned
 {
     /// Returns the minimum possible namespace
     fn min() -> Self;
@@ -226,9 +226,7 @@ where
         &mut self,
         elem: impl core::borrow::Borrow<Self::Element>,
     ) -> Result<(), PrimitivesError> {
-        let cloned = elem.borrow().clone();
-        NMT::<E, H, Arity, N, T>::update_namespace_metadata(&mut self.namespace_ranges, [elem])?;
-        self.inner.push(cloned)
+        self.extend([elem])
     }
 }
 
--- a/primitives/src/merkle_tree/namespaced_merkle_tree/mod.rs
+++ b/primitives/src/merkle_tree/namespaced_merkle_tree/mod.rs
@@ -93,7 +93,7 @@ where
     fn generate_namespaced_commitment(namespaced_hash: NamespacedHash<T, N>) -> T;
 }
 
-/// Trait indiciating that a struct can act as an orderable namespace
+/// Trait indicating that a struct can act as an orderable namespace
 pub trait Namespace:
     Debug
     + Clone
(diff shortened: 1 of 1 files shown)


primitives/src/merkle_tree/namespaced_merkle_tree/mod.rs
```diff
@@ -93,7 +95,16 @@ where
...
+    Debug
+    + Clone
+    + CanonicalDeserialize
+    + CanonicalSerialize
+    + Default
+    + Copy
+    + Hash
+    + Ord
+    + Serialize
+    + for<'a> Deserialize<'a>
```
jbearer, line 107: You should use `DeserializeOwned` instead (`serde::de::DeserializeOwned`). I believe it's just an alias for this bound but serde has the right to change it in the future

primitives/src/merkle_tree/namespaced_merkle_tree/mod.rs
```diff
@@ -221,22 +217,24 @@ where
...
+        self.inner.extend(leaves)
     }
 
     fn push(
         &mut self,
         elem: impl core::borrow::Borrow<Self::Element>,
     ) -> Result<(), PrimitivesError> {
-        // TODO(#314): update namespace metadata
-        self.inner.push(elem)
+        let cloned = elem.borrow().clone();
```
jbearer, line 229: At first I was confused why we need to clone here but not in `extend`. I wonder if it would be clearer to `let leaves = NMT::update_namespace_metadata(...)` and then `self.inner.push(leaves[0])` or `self.inner.extend(leaves)` (even though `leaves` only has 1 element). To more closely mirror `extend`. Definitely not a big deal either way, though

sveitser: I wonder if it would make sense to implement this as `self.extend([elem])` instead?

jbearer: Yeah that would be quite elegant

nomaxg: Great idea

philippecamacho: In practice why do we need to update the tree? Is it not enough to build the tree from the leaves only once? Or do we do this in order to keep the consistency of the API?

nomaxg: The current Hotshot Block trait requires us to push one tx at a time

nomaxg: @jbearer it would be ideal if we could construct the HotShot block in one go @jbearer does something about consensus preclude us from doing this?

jbearer: It comes from validating consensus, where the builder would push one transaction at a time, checking if it made the block inconsistent, as a way to filter out invalid transactions from the proposal. I think for sequencing consensus the builder could theoretically collect all transactions into a `Vec` as they are submitted, and then build the block all at once. You'll have to talk to the consensus team to see if an API change is possible

primitives/src/merkle_tree/namespaced_merkle_tree/mod.rs
```diff
@@ -93,7 +95,16 @@ where
 
 /// Trait indiciating that a struct can act as an orderable namespace
```
philippecamacho, line 96: indiciating -> indicating

primitives/src/merkle_tree/namespaced_merkle_tree/mod.rs
```diff
@@ -421,14 +451,40 @@ mod nmt_tests {
...
+    }
+
+    fn test_nmt_with_build_type(build_type: BuildType) {
         let namespaces = [1, 2, 2, 2, 4, 4, 4, 5];
         let first_ns = namespaces[0];
         let last_ns = namespaces[namespaces.len() - 1];
         let internal_ns = namespaces[1];
         let leaves: Vec<Leaf> = namespaces.iter().map(|i| Leaf::new(*i)).collect();
-        let tree = TestNMT::from_elems(3, leaves.clone()).unwrap();
+        let tree = match build_type {
```
philippecamacho, line 473: Do we expect the trees to be the same regardless of how they are built?

If this is the case maybe a dedicated test to check that would be worth it.  Because the current test can pass with different trees, for example if we replace lines 476 and 477 by

```rust
     let mut nmt = TestNMT::from_elems(3, &[leaves[0],leaves[1]]).unwrap();
     nmt.extend([leaves[2],leaves[3],leaves[4],leaves[5]]).unwrap(); // No elements leaves[6], leaves[7]
```
 the test still passes.

nomaxg: yes I believe that trees should be same as long as leaves are inserted in the same order @mrain is this a fair assumption? If that is the case, I can write a dedicated test that compares the roots and the namespace ranges for consistency

nomaxg:
@philippecamacho added a test for tree consistency

philippecamacho (approved):
LGTM

PR #337, opened 2023-07-03
feat(vid): port VID from another repo

ggutoski:
## Description

closes: #336 

A direct port of `vid` from hotshot-primitives repo. Nontrivial changes:
- Feature-gate tests with `test-srs` feature. (This was much easier from another repo because `test-srs` could be specified in a dev-dependency in `Cargo.toml`, but there seems to be no idiomatic way to enable a feature of the present crate only for test/bench code 😢 .)
- Support `no_std`. Remove use of `thiserror` crate, which is not `no_std` compatible, and tweak associated error-handling code.

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/plonk/src/circuit/plonk_verifier/gadgets.rs
+++ b/plonk/src/circuit/plonk_verifier/gadgets.rs
@@ -207,7 +207,7 @@ where
     }
     let mut transcript_var = RescueTranscriptVar::new(circuit);
     if let Some(msg) = extra_transcript_init_msg {
-        let msg_fs = bytes_to_field_elements::<_, F>(msg);
+        let msg_fs = bytes_to_field_elements::<_, F>(<&Vec<u8> as AsRef<[u8]>>::as_ref(&msg));
         let msg_vars = msg_fs
             .iter()
             .map(|x| circuit.create_variable(*x))
--- a/plonk/src/circuit/plonk_verifier/gadgets.rs
+++ b/plonk/src/circuit/plonk_verifier/gadgets.rs
@@ -207,7 +207,7 @@ where
     }
     let mut transcript_var = RescueTranscriptVar::new(circuit);
     if let Some(msg) = extra_transcript_init_msg {
-        let msg_fs = bytes_to_field_elements::<_, F>(<&Vec<u8> as AsRef<[u8]>>::as_ref(&msg));
+        let msg_fs = bytes_to_field_elements::<_, F>(msg);
         let msg_vars = msg_fs
             .iter()
             .map(|x| circuit.create_variable(*x))
--- a/primitives/src/vid/mod.rs
+++ b/primitives/src/vid/mod.rs
@@ -1,0 +1,69 @@
+//! Trait and implementation for a Verifiable Information Retrieval (VID).
+//!
+/// See <https://arxiv.org/abs/2111.12323> section 1.3--1.4 for intro to VID semantics.
+use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
+use ark_std as std; // needed for thiserror crate
+use ark_std::{fmt::Debug, string::String, vec::Vec};
+
+pub mod advz;
+
+/// The error type for `VidScheme` methods.
+///
+/// # Use of both `thiserror` and `anyhow`
+/// This library is both a producer and consumer of errors.
+/// It provides a custom error `VidError` for consumers of this library, aided by `thiserror`.
+/// Moreover, it is a consumer of errors from lower-level libraries, aided by `anyhow`.
+/// We have yet to settle on a preferred error handling philosophy.
+#[derive(thiserror::Error, Debug)]
+pub enum VidError {
+    /// Caller provided an invalid argument
+    #[error("invalid arguments: {0}")]
+    Argument(String),
+    /// Internal error
+    #[error(transparent)]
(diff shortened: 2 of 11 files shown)


primitives/tests/advz.rs
```diff
@@ -0,0 +1,41 @@
...
+use jf_primitives::{
+    pcs::{checked_fft_size, prelude::UnivariateKzgPCS, PolynomialCommitmentScheme},
+    vid::advz::Advz,
+};
+use sha2::Sha256;
+
+mod vid;
+
+#[test]
+fn round_trip() {
```
alxiong, line 13: a strange thing is this integration test is [not run inside CI](https://github.com/EspressoSystems/jellyfish/actions/runs/5449191040/jobs/9913183995#step:13:223). idk why.

primitives/src/vid/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+    E: ark_std::fmt::Display + Debug + Send + Sync + 'static,
+{
+    VidError::Internal(anyhow::anyhow!(e))
+}
+
+/// Convenience [`Result`] wrapper for [`VidError`].
+pub type VidResult<T> = Result<T, VidError>;
+
+/// VID: Verifiable Information Dispersal
+pub trait VidScheme {
```
alxiong, line 39: I wonder if we should add two associated constants as the config for an instantiation of this VID scheme, namely `payload_chunk_size` and `num_storage_nodes`?

alxiong (commented):
Thanks Gus! generally LGTM, left a few comments.

primitives/src/vid/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+/// # No-std support
+/// `no_std` mode requires `.map_err(vid)` to convert from a non-`anyhow` error
+/// as per [`anyhow` docs](https://docs.rs/anyhow/latest/anyhow/index.html#no-std-support),
+fn vid<E>(e: E) -> VidError
+where
+    E: ark_std::fmt::Display + Debug + Send + Sync + 'static,
+{
+    VidError::Internal(anyhow::anyhow!(e))
+}
+
```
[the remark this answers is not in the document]
ggutoski: We could. I chose `vid` because you need to put `.map_err(vid)` everywhere, so I want a short name. I agree that `vid` on its own is not helpful, but `.map_err(vid)` seems clear to me.

Do you, as a fresh observer, prefer `vid_internal_err`? If so then I'm happy to change it.

Aside: sucks that this is state-of-the-art for error handling in Rust. 😕

primitives/tests/advz.rs
ggutoski: Hmm. My guess is it's because `tests/advz.rs` has `#![cfg(feature = "test-srs")]` on line 1. I did this because tests do not compile without it:
```
no function or associated item named `gen_srs_for_testing` found for struct `UnivariateKzgPCS` in the current scope
```
`gen_srs_for_testing` is feature-gated according to `#[cfg(any(test, feature = "test-srs"))]`. Seems like `test` is having no effect in this integration, so I need to gate behind `test-srs`. wat do 🤷 

Aside: yet another basic usability issue in Rust. I wish they'd fix this stuff already.

primitives/src/vid/mod.rs
ggutoski: I can imagine applications where the user might want to change these values dynamically at runtime, in which case they cannot be compile-time constants. (eg. number of storage nodes might change with time, and `payload_chunk_size` is often a function of `num_storage_nodes`.)

In any case, this topic is outside the scope of this PR. We could post a separate issue if we don't want to forget about it in the future.

plonk/src/circuit/plonk_verifier/gadgets.rs
```diff
@@ -207,7 +207,7 @@ where
     }
     let mut transcript_var = RescueTranscriptVar::new(circuit);
     if let Some(msg) = extra_transcript_init_msg {
-        let msg_fs = bytes_to_field_elements::<_, F>(msg);
+        let msg_fs = bytes_to_field_elements::<_, F>(<&Vec<u8> as AsRef<[u8]>>::as_ref(&msg));
```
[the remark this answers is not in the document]
ggutoski: Sorry about that. Reverted this change in https://github.com/EspressoSystems/jellyfish/pull/337/commits/9a366f1b64696624b633afea3f5889d9e124b176 . My local state was screwed up. (`Cargo.lock` again!)

primitives/tests/advz.rs
ggutoski: Added `test-srs` feature flag to test script in https://github.com/EspressoSystems/jellyfish/pull/337/commits/639e51d0990f62ef3c2402715b23a241bfaa01cf . This test [now runs in CI](https://github.com/EspressoSystems/jellyfish/actions/runs/5456171771/jobs/9928625023?pr=337#step:13:370). 🎉

primitives/src/vid/mod.rs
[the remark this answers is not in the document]
alxiong: > because you need to put .map_err(vid) everywhere, so I want a short name.

yeah, I also guessed that's the reason. But yes, I do prefer clearer names here. Thanks for indulging me :)

alxiong (approved):

Issue #375, opened 2023-09-20
VID dispersal, commitment should take an iterator over `u8` instead of `&[u8]`


ggutoski:
https://github.com/EspressoSystems/jellyfish/blob/f6866c5395405f807e62516b5d56d9dc390aeb61/primitives/src/vid/mod.rs#L56-L59

This forces the caller to marshal its data as a contiguous `u8` slice, which is a waste of time and memory.

ggutoski:
# Problem

`bytes_to_field_elements` and `bytes_from_field_elements` also need to change to take an iterator:
https://github.com/EspressoSystems/jellyfish/blob/f6866c5395405f807e62516b5d56d9dc390aeb61/utilities/src/conversion.rs#L129-L131

Unfortunately, current impls of these functions need to know the size up front. We could do it if we require the iterator to be `ExactSizeIterator`, but the expected use case is an iterator formed by chaining a bunch of other iterators. (Example: `Vec<Vec<u8>>`). Sadly, chained iterators can never be `ExactSizeIterator`, even when composed of iterators that are: https://github.com/rust-lang/rust/pull/66531

Thus, any solution to this issue requires a change to `bytes_[to|from]_field_elements` that does not require `ExactSizeIterator`. That's annoying because we need to find a new way to keep the invertibility guarantee without `ExactSizeIterator`. 😕

ggutoski:
duplicate of #300 . But the [above comment](https://github.com/EspressoSystems/jellyfish/issues/375#issuecomment-1728502055) is new.

ggutoski:
# Candidate solution

As described in #300, `bytes_[to|from]_field_elements` should be an "iterator adaptor".

Currently we prefix the output with a field element that encodes the byte length of the input. Don't do this because it requires knowing the byte length at the beginning. Instead, append an additional field element *at the end* of the iterator that encodes the number of bytes to discard from the final field element.

Example: Field elements take 32 bytes, but the input byte length is 14 mod 32, so we append an additional field element that encodes the number 14, indicating that only the first 14 bytes of the final field element are part of the input and so the remaining 18 bytes should be discarded.

This solution requires `peekable` (the ability to peek 1 item ahead). Fortunately, `peekable` is part of the `Iterator` trait, so all iterators have it (including chained iterators).

Issue #300, opened 2023-06-06
bytes_to_field_elements should accept and return an iterator


ggutoski:
# Problem

_Originally posted by @ggutoski in https://github.com/EspressoSystems/jellyfish/pull/299#discussion_r1220149576_

Currently `bytes_to_field_elements` takes `AsRef<[u8]>` as input and returns `Vec<F>` as output. This can lead to unnecessary mem copy or heap alloc if the caller doesn't already have a `AsRef<[u8]>` or wants the output in some form other than `Vec<F>`.

Example: here the caller actually wants a `Vec<Vec<F>>`, and needs to copy/allocate a whole 2D vec for this purpose.

https://github.com/EspressoSystems/hotshot-primitives/blob/52b2f713eea897778eb8a73f401c9b1b4a2dde19/src/vid/advz.rs#L158-L159

# Solution

`bytes_to_field_elements` should accept `IntoIterator` instead of `AsRef<[u8]>` and should return an iterator over `F` instead of `Vec<F>`.

We should do the same for `bytes_from_field_elements`, though that will be more complicated to implement.

This solution is an example of the *iterator adaptor* pattern:
- [Processing a Series of Items with Iterators - The Rust Programming Language](https://doc.rust-lang.org/book/ch13-02-iterators.html#methods-that-produce-other-iterators)
- [std::iter - Rust](https://doc.rust-lang.org/stable/std/iter/index.html#adapters)

amit0365:
Hello I am planning to work on this. Are you close to implementing the solution @ggutoski ?

ggutoski:
> Hello I am planning to work on this. Are you close to implementing the solution @ggutoski ?

Hello @amit0365 . I have spent no time on this issue. Feel free to work on it! ❤️

PR #381, opened 2023-10-05
feat: VID take iterator instead of slice

ggutoski:
## Description

closes: #375
closes: #300 

- `conversion.rs`: add new `bytes_to_field` and `field_to_bytes` that act on iterators instead of slices. The code is more complex than I like. But if you trust the tests then you don't need to get too deep into it. The cause of this complexity is that we need to append an extra field element at the end to encode some length information in order to enable inversion of this process.
  - These methods would be much simpler if we did not need to encode this metadata. But that would make these functions useless because the caller must know in advance how many bytes are in the iterator so he can truncate the decoded bytes accordingly. Since we're now using iterators, not slices, it seems unreasonable for the caller to know this.
  - These methods work only on `PrimeField`, not `Field`. We could extend it to `Field` but I don't want to take extra time for that now.
- I did not delete the existing functions `bytes_to_field_elements`, `bytes_from_field_elements` that act on slices. But we could delete them if needed. (Currently they're still used in plonk.)
- Update VID API to take iterators instead of slices, use the new conversion methods.
- Add a new `elems_len` field to `Common` struct in VID. This data is needed because recovery adds extra zero elements via RS decoding. Previously we did not need this because the existing `bytes_to_field_elements` encodes the *total* byte length in the output. Thus, `bytes_from_field_elements` automatically drops any extra zero elements added by RS decoding. By contrast, the new `bytes_to_field` encodes only the byte length of the *final* element (not the total byte length). So now VID needs to keep track of how many zero elems were added by RS decoding.

## Other stuff snuck into this PR

- clean itertools dependency
- add new derivations for `VidDisperse` struct that are useful downstream

---

write a little note why.

- [x] Wrote unit tests


primitives/src/vid/advz.rs
```diff
@@ -140,17 +140,21 @@ where
...
 // 1,2: `Polynomial` is univariate: domain (`Point`) same field as range
 // (`Evaluation'). 3,4: `Commitment` is (convertible to/from) an elliptic curve
 // group in affine form. 5: `H` is a hasher
+//
+// `PrimeField` needed only because `bytes_to_field` needs it.
+// Otherwise we could relax to `FftField`.
 impl<P, T, H, V> VidScheme for GenericAdvz<P, T, H, V>
 where
     P: UnivariatePCS<Point = <P as PolynomialCommitmentScheme>::Evaluation>,
-    P::Evaluation: FftField,
```
mrain, line 153: Any specific reason to bound it on `PrimeField`?

primitives/src/vid/advz.rs
```diff
@@ -163,15 +167,22 @@ where
...
 
-        // TODO perf: DenseUVPolynomial::from_coefficients_slice copies the slice.
-        // We could avoid unnecessary mem copies if bytes_to_field_elements returned
-        // Vec<Vec<F>>
-        let elems = bytes_to_field_elements(payload);
+        // TODO perf: is it possible to avoid collect() here?
+        let elems: Vec<_> = bytes_to_field::<_, P::Evaluation>(payload).collect();
+
         for coeffs in elems.chunks(self.payload_chunk_size) {
+            // TODO perf: DenseUVPolynomial::from_coefficients_slice copies the slice.
```
mrain, line 181: Is follow-up issue created?

primitives/src/vid/advz.rs
ggutoski: I was in a rush. 😊 I'm happy to extend it to `Field` if you request it. That gives me an excuse to allocate the time to do it.

ggutoski: The answer to your question is here: https://github.com/EspressoSystems/jellyfish/blob/474fcdf6a40102ce0acfc5c3729ed539562c5051/primitives/src/vid/advz.rs#L152-L153

This https://github.com/EspressoSystems/jellyfish/pull/381#discussion_r1349016039 is an explanation for why I restricted `bytes_to_field` to `PrimeField`.

mrain: Got it. We can settle with `PrimeField` for now, and add a follow-up issue for the future improvement.

mrain (approved):
LGTM. We can merge it after some follow-up issues and CI fixes.

primitives/src/vid/advz.rs
ggutoski: https://github.com/EspressoSystems/jellyfish/issues/382

Issue #388, opened 2023-10-26
T2: VID hello-world namespace support


ggutoski:
Make a feature-complete (but rudimentary) implementation of VID namespace support for ADVZ scheme.

## Goals

- Get an API we can use downstream. We will need to iterate on this API.
- Get something working. It'll be slow at first. After we have feature-completeness we can pick some low-hanging optimization fruit such as #339 #387 .

PR #389, opened 2023-10-30
feat: hello-world namespace support for ADVZ VID scheme

ggutoski:
closes: #388 

# Revised 2023-11-06

Force-pushed update incorporating comments from previous reviews and offline discussion.

- Add `PayloadProver` subtrait for `VidScheme`. It's a generic trait with two implementations for ADVZ. See rustdoc: https://github.com/EspressoSystems/jellyfish/blob/93e396ebed1421458e2647d6461932dffb44680d/primitives/src/vid/advz/payload_prover.rs#L9-L15
- Change `VidScheme` to use `&[u8]` for payload type as per offline discussion.
- move `bytes_to_field` from `conversion.rs` to a new `bytes_to_field.rs` file owned by `advz`.

# Excerpts from original description

- change the bytes-to-field conversion in `conversion.rs`. It's much simpler now because I decided we don't need to add metadata into the output field elements. Instead, it's the responsibility of the caller to know where exactly their bytes begin/end in the encoded field elements.
- `advz.rs` has a large diff but it's mostly just moving code around.
- fix a bug in encoding payload bytes to polynomials wherein FFT and inverse-FFT were accidentally swapped.
- fix some nix stuff 

---

write a little note why.

- [x] Wrote unit tests


primitives/src/vid/namespace.rs
```diff
@@ -0,0 +1,53 @@
...
+    /// Verify a proof for `payload` for data index range `start..start+len-1`.
+    ///
+    /// See TODO in `namespace_verify` on `payload`.
+    fn data_verify(
+        &self,
+        payload: &Self::Payload,
+        start: usize,
+        len: usize,
+        commit: &Self::Commit,
+        common: &Self::Common,
```
jbearer, line 35: Are `commit` and `common` always together? Would it make sense to just put all the common stuff in `commit`?

primitives/src/vid/namespace.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    /// Compute a proof for `payload` for data index range `start..start+len-1`.
+    ///
+    /// See TODO in `namespace_verify` on `payload`.
+    fn data_proof(
+        &self,
+        payload: &Self::Payload,
+        start: usize,
+        len: usize,
+    ) -> VidResult<Self::DataProof>;
```
jbearer, line 24: The prover (e.g. query server) also needs some way to get the actual data. So I think this should be something like `VidResult<(Vec<u8>, Self::DataProof)>`, or `Self::DataProof` should have some trait bound that allows me to extract the data from the proof

primitives/src/vid/namespace.rs
```diff
@@ -0,0 +1,53 @@
...
+    ) -> VidResult<Result<(), ()>>;
+
+    /// Verify the `payload` namespace indexed by `namespace_index` against
+    /// `commit`, `common`.
+    ///
+    /// TODO: We prefer not to include the whole `payload`. But the namespace
+    /// proof needs a few payload bytes from outside the namespace. In the
+    /// future `payload` should be replaced by a payload subset that includes
+    /// only the bytes needed to verify a namespace.
+    fn namespace_verify(
```
jbearer, line 46: I thought we discussed not having any namespace-specific interface in Jellyfish, but having the sequencer itself use the namespace table to convert from namespace index to byte range, and then just use the above APIs to read bytes?

primitives/src/vid/namespace.rs
ggutoski: `commit` is supposed to be an extremely small item (like a 32-byte hash).

`common` is metadata sent to all storage nodes such as KZG commitments to all the polynomials in a block payload. It's size is not bounded. I view it more as part of a proof to be verified against `commit` than as part of `commit` itself.

primitives/src/vid/namespace.rs
ggutoski: My understanding is that the query server will already have `payload` in full. Thus, it already has the actual data---it could grab its own subslice of `payload`.

But this raises a related issue: Due to bytes-to-field encoding, the user needs a few bytes outside the tx/namespace in order to verify the proof. Where should these bytes go? I guess it makes sense to put those few bytes inside the `DataProof`.

But it's not clear to me that the actual data should also go into the `DataProof`.

primitives/src/vid/namespace.rs
ggutoski: My takeaway from that discussion is the rule "no namespace-specific stuff in jellyfish, instead just give me a byte range" applies only to  *transactions*.

Things are a bit different for a full namespace proof. I do suspect that `namespace_verify` in jellyfish will never be used except in tests for correctness. So basically this function is a reference implementation against which to test smart contract verification of a namespace. wdyt?

primitives/src/vid/advz.rs
```diff
@@ -194,22 +198,17 @@ where
...
-        I::Item: Borrow<u8>,
-    {
+    fn commit_only(&self, payload: &Self::Payload) -> VidResult<Self::Commit> {
+        // Can't use `Self::poly_commits_hash()`` here because `P::commit()`` returns
+        // `Result<P::Commitment,_>`` instead of `P::Commitment`.
+        // There's probably an idiomatic way to do this using eg.
+        // itertools::process_results() but the code is unreadable.
         let mut hasher = H::new();
         let elems_iter = bytes_to_field::<_, P::Evaluation>(payload);
         for coeffs_iter in elems_iter.chunks(self.payload_chunk_size).into_iter() {
```
chancharles92, line 210: `coeffs_iter` -> `evals_iter`? (the elements are no longer coefficients.)

primitives/src/vid/advz.rs
```diff
@@ -218,124 +217,15 @@ where
...
-        I: IntoIterator,
-        I::Item: Borrow<P::Evaluation>,
-    {
         // partition payload into polynomial coefficients
         // and count `elems_len` for later
-        let elems_iter = payload.into_iter().map(|elem| *elem.borrow());
-        let mut elems_len = 0;
+        let elems_iter = bytes_to_field::<_, P::Evaluation>(payload).map(|elem| *elem.borrow());
         let mut polys = Vec::new();
         for coeffs_iter in elems_iter.chunks(self.payload_chunk_size).into_iter() {
```
chancharles92, line 227: Same here, maybe better to change the variable name of `coeffs_iter` to `evals_iter`.

primitives/src/vid/advz.rs
```diff
@@ -388,20 +278,10 @@ where
...
-                (poly_commit as &P::Commitment)
-                    .serialize_uncompressed(&mut hasher)
-                    .map_err(vid)?;
-            }
-            hasher.finalize()
+            bytes_len,
         };
 
+        let commit = Self::poly_commits_hash(common.poly_commits.iter())?;
         let pseudorandom_scalar = Self::pseudorandom_scalar(&common)?;
```
chancharles92, line 285: Can we derive `pseudorandom_scalar` from `commit` + `all_evals_digest`? There can be redundant hash calls if we derive from `poly_commits`.

primitives/src/vid/advz/namespace.rs
```diff
@@ -0,0 +1,355 @@
...
+        // check args:
+        // TODO TEMPORARY: forbid requests that span multiple polynomials
+        if len_namespace > 1 {
+            return Err(VidError::Argument(format!(
+                "request spans {} polynomials, expect 1",
+                len_namespace
+            )));
+        }
+
+        // grab the `start_namespace`th polynomial
```
chancharles92, line 69: Optional TODO in the future PR: save this computation.

primitives/src/vid/advz/namespace.rs
```diff
@@ -0,0 +1,355 @@
...
+    H: Digest + DynDigest + Default + Clone + Write,
+    V: MerkleTreeScheme<Element = Vec<P::Evaluation>>,
+    V::MembershipProof: Sync + Debug,
+    V::Index: From<u64>,
+{
+    // TODO should be P::Proof, not Vec<P::Proof>
+    // https://github.com/EspressoSystems/jellyfish/issues/387
+    type DataProof = Vec<P::Proof>;
+
+    fn data_proof(
```
chancharles92, line 39: In the namespace application, does the prover need to generate these proofs online? Can it just batchly preprocess all of the evaluation proofs offline, and in the online phase, whenever a client asks, it just forwards proofs without doing any computations?
cc @ggutoski @jbearer

ggutoski: Yes, it could. I suspect servers will opt to pre-process everything if they get sufficient traffic. Seems to me that the the priority is to optimize data on the wire, so KZG batch proofs should be quite helpful here.

primitives/src/vid/namespace.rs
jbearer: Gotcha

primitives/src/vid/namespace.rs
jbearer: Yeah, putting the extra bytes in `DataProof` and having the payload data separate seems fine. But the query server still needs some interface for actually reading bytes from the full payload. I don't see that here since `Self::Payload` doesn't have a trait bound that would allow me to read bytes out of it

ggutoski: Good point. I'll add something for that.

ggutoski: As discussed offline I'll remove the `Payload` assoc type and just enforce `&[u8]` payload. It hurts my soul to be so opinionated in the API, but we discussed that downstream users are assumed to bear the cost of serializing to a 1D byte slice, and this assumption makes our life much easier.

primitives/src/vid/namespace.rs
ggutoski: As discussed offline, both namespace and tx proofs will have the exact same API. I'll make a generic trait and implement it twice: once for namespace proofs (snark-friendly) and once for tx proofs (requires a pairing). There will be no namespace-specific stuff in jellyfish. Instead the API is just "give me a index range and I'll give you a proof."

primitives/src/vid/advz.rs
ggutoski: changed `coeffs_iter` -> `evals_iter` because it's (an iterator over) the evaluations of a polynomial.

primitives/src/vid/advz.rs
ggutoski: Yep. Done in commit 8f2240b3f9d01c29f77142c8b2dd8698c0be79fe

primitives/src/vid/advz/namespace.rs
ggutoski: #397

primitives/src/vid/advz/namespace.rs
ggutoski: #397

primitives/src/vid/payload_prover.rs
```diff
@@ -0,0 +1,31 @@
...
+    /// Compute a proof for a sub-slice of payload data.
+    fn payload_proof<B>(&self, payload: B, range: Range<usize>) -> VidResult<PROOF>
+    where
+        B: AsRef<[u8]>;
+
+    /// Verify a proof made by `payload_proof`.
+    fn payload_verify<B>(
+        &self,
+        chunk: B,
+        commit: &Self::Commit,
```
jbearer, line 25: I think `range` should also be a public input here. The statement being verified is not just `chunk` is a slice from `commit`, but `chunk` is _the_ slice at position `range` in `commit`.

primitives/src/vid/advz/payload_prover.rs
```diff
@@ -0,0 +1,505 @@
+// Copyright (c) 2023 Espresso Systems (espressosys.com)
+// This file is part of the Jellyfish library.
+
+// You should have received a copy of the MIT License
+// along with the Jellyfish library. If not, see <https://mit-license.org/>.
+
+//! Implementations of [`PayloadProver`] for `Advz`.
+//!
+//! Two implementations:
```
jbearer, line 9: It would be nice to give type aliases for these different proof types in terms of the high level properties/use cases, that make the contrasts obvious. I don't have any great ideas, though. Maybe something like `GranularPayloadProof`/`RangePayloadProof` or something?

primitives/src/vid/advz/payload_prover.rs
```diff
@@ -0,0 +1,505 @@
...
+    AffineRepr, Debug, DenseUVPolynomial, Digest, DynDigest, GenericAdvz, MerkleTreeScheme,
+    PolynomialCommitmentScheme, PrimeField, UnivariatePCS, Vec, VidResult, Write,
+};
+use crate::{
+    alloc::string::ToString,
+    vid::{payload_prover::PayloadProver, vid, VidError, VidScheme},
+};
+use ark_std::{format, ops::Range};
+
+impl<P, T, H, V> PayloadProver<Proof<P::Proof>> for GenericAdvz<P, T, H, V>
```
jbearer, line 30: Something slightly annoying about this interface is as a downstream user I'm probably going to have to write out whatever `P::Proof` is in order to help the typechecker figure out which kind of proof I want, since there are two different implementations with the same interface. In practice I'm probably going to be referring to the VID scheme itself through a type alias with specific instantiations of `<P, T, H, V>`. Makes it kind of annoying to click through rust docs to try and figure out what `P` is and then what `P::Proof` is so I can name this `Proof<P::Proof>` type.

Two ideas for making this a bit more ergonomic:

1. Bring back the monolothic prover trait with two pairs of methods with identical interfaces, but different names. That way it's obvious from the name of the method which type of proof I'm asking for, I don't have to actually write down the type. However, this would enshrine in the trait interface the idea that we have these two separate types of proofs, which might not be true for every conceivable VID scheme.
2. Introduce a marker type which we use to denote the kind of proof we're looking for abstractly, decoupled from the specific type we use to implement that proof. The marker type can be used to select the trait implementation I want and the actual type of proof can be an associated type of that trait. Something roughly like this:

```rust
trait PayloadProver<ProofType>: VidScheme {
  type Proof;
  fn payload_proof(...) -> VidResult<Self::Proof>;
}

struct SnarkFriendlyProof;
struct GranularProof;

impl PayloadProver<GranularProof> for GenericAdvz<P, T, H, V> {
  type Proof = Proof<P::Proof>;
  ...
}

impl PayloadProver<GranularProof> for GenericAdvz<P, T, H, V> {
  type Proof = CommitRecovery<P::Evaluation>;
  ...
}
```

Use like
```rust
PayloadProver::<GranularProof>::payload_proof(&scheme, payload, range);
PayloadProver::<SnarkFriendlyProof>::payload_proof(&scheme, payload, range);
```

primitives/src/pcs/univariate_kzg/mod.rs
```diff
@@ -92,8 +92,9 @@ impl<E: Pairing> PolynomialCommitmentScheme for UnivariateKzgPCS<E> {
         poly: &Self::Polynomial,
     ) -> Result<Self::Commitment, PCSError> {
         let prover_param = prover_param.borrow();
-        let commit_time =
-            start_timer!(|| format!("Committing to polynomial of degree {} ", poly.degree()));
+        // let commit_time =
```
chancharles92, line 95: Minor: we can uncomment it but add a flag-feature so that it doesn't execute when we do disperse benchmark.

primitives/src/vid/payload_prover.rs
ggutoski: It's assumed that the `range` data is included in `proof`. Both impls for this trait include a `chunk_range: Range<usize>,` field. It's sufficient to pass tests. I prefer it this way to reduce clutter in the args. Sound ok to you?

primitives/src/vid/advz.rs
```diff
@@ -380,45 +294,48 @@ where
...
-                    .serialize_uncompressed(&mut hasher)
-                    .map_err(vid)?;
-            }
-            hasher.finalize()
+            bytes_len: payload_len,
         };
+        end_timer!(common_timer);
 
-        let pseudorandom_scalar = Self::pseudorandom_scalar(&common)?;
+        let misc_timer = start_timer!(|| {
```
chancharles92, line 311: We can omit this timer.

primitives/src/vid/advz/payload_prover.rs
ggutoski: I like the marker type idea and I think we should use it if this is indeed a problem. But are you sure it will be so bad for downstream users? My tests work with the following:
```rust
let data_proof: Proof<_> = advz.payload_proof(payload, range)?;
let chunk_proof: CommitRecovery<_> = advz.payload_proof(payload, range)?;
```
Admittedly the names `Proof` and `CommitRecovery` kinda suck. It's hard to come up with good names without mentioning "namespace" or "transaction". 😉

primitives/src/vid/advz/payload_prover.rs
chancharles92: Agree, it's better to distinguish these two types of implementations/APIs.

primitives/src/vid/payload_prover.rs
jbearer: I think morally it should be an argument because it's part of the statement being proven, conceptually, not part of the proof. Otherwise as a verifier, I need some way to extract the `chunk_range` from the proof to check it against the expected range, which needs to be incorporated into a trait somehow to make sense with the rest of the trait-based interface. And, we introduce a footgun where the verifier forgets to do this extra check, which is much harder to do if the expected range is a required parameter.

primitives/src/vid/advz/payload_prover.rs
jbearer: Ah the `<_>` is a nice trick. I think that's fine then

primitives/src/pcs/univariate_kzg/mod.rs
ggutoski: Oops! Forgot to uncomment this. Add `kzg-print-trace` feature flag in https://github.com/EspressoSystems/jellyfish/pull/389/commits/a0678336d0061f4f6fb2406401a5cc2179191c2e . I added it only to those timers that were causing me noise. I could add it everywhere else, too, but it does pollute the code.

primitives/src/vid/advz.rs
ggutoski: done in https://github.com/EspressoSystems/jellyfish/pull/389/commits/a0678336d0061f4f6fb2406401a5cc2179191c2e

chancharles92 (commented):
LGTM, can approve after addressing Jeb's concern about APIs.

primitives/src/vid/payload_prover.rs
ggutoski: New `Statement` struct in https://github.com/EspressoSystems/jellyfish/pull/389/commits/f399a4fbc7903b3b9d1c39df86ae7e248b6cd893

It's not needed in `payload_proof` but it does improve the readability of `payload_verify`.

I'll leave it to you to resolve this conversation when you're satisfied.

primitives/src/vid/advz/payload_prover.rs
ggutoski: Renamed to `SmallRangeProof`, `LargeRangeProof` and tidied in https://github.com/EspressoSystems/jellyfish/pull/389/commits/eb8fde47d2e88a7dd27c2e8f88c9df6e75f2564d .

Unfortunately it's not clear how to give a useful type alias, as I would need to specify the type parameter `P`. But the whole file is generic over `P` so any type alias would need to hard-code a specific choice for `P`, which I don't want to do.

jbearer: You could do `type SmallRangeProof<P> = Proof<<P as Trait>::Proof>`. But renaming is also good 👍

primitives/src/vid/advz/payload_prover.rs
```diff
@@ -23,11 +24,37 @@ use super::{
...
 };
 use ark_std::{format, ops::Range};
 
-impl<P, T, H, V> PayloadProver<Proof<P::Proof>> for GenericAdvz<P, T, H, V>
+/// A proof intended for use on small payload subslices.
+///
+/// KZG batch proofs and accompanying metadata.
+///
+/// TODO use batch proof instead of `Vec<P>` <https://github.com/EspressoSystems/jellyfish/issues/387>
+pub struct SmallRangeProof<P> {
```
jbearer, line 39: The proof types should at least derive `Clone`, `Debug`, and serde traits

primitives/src/vid/payload_prover.rs
```diff
@@ -23,13 +23,51 @@ pub trait PayloadProver<PROOF>: VidScheme {
     /// `chunk` is the payload sub-slice for which a proof was generated via
```
jbearer, line 23: Comment needs to be updated,  we no longer have `chunk`

jbearer (approved):
Looks good. I think we might be missing some important traits on the proof types, particularly serdes. But if you want we can get this merged and add those later. Or add them now. Either way

primitives/src/vid/advz/payload_prover.rs
ggutoski: Definitely true. I'm going to merge now anyway because (i) this PR is already too big, (ii) don't want to initiate another round of approval, (iii) there will certainly be back-and-forth over issues like this as we integrate into the sequencer.

PR #390, opened 2023-10-31
feat: Reproducible KZG benches

tessico:
## Description

The existing KZG bench only samples one polynomial and one query point per size parameter (num vars): this leads to quite a lot of variability across runs (try running it twice, you'll get significant discrepancies!).

This PR fixes the above issue by sampling multiple polys and opening points per parameter set, and also uses criterion framework for more detailed and comparable results.

closes: #XXXX

---

write a little note why.

- [ ] Wrote unit tests


chancharles92 (commented):
Thanks @tessico , @Antonio95 , LGTM. Could you also fix the build? Thanks.

primitives/benches/pcs.rs
```diff
@@ -4,100 +4,136 @@
...
+
+    let start = Instant::now();
+    assert!(MultilinearKzgPCS::verify(&vk, &commitment, &point, &value, &proof).unwrap());
+    start.elapsed()
+}
+
+fn kzg_254(c: &mut Criterion) {
+    bench_pcs_method::<Bn254>(
+        c,
+        (MIN_NUM_VARS..MAX_NUM_VARS).step_by(2).collect(),
```
alxiong, line 115: Instead of collecting it into a Vec, can we change this function to accept [Range](https://doc.rust-lang.org/std/ops/struct.Range.html) ?

alxiong (commented):
LGTM! Thanks a lot for the PR! Just a minor comment.

primitives/benches/pcs.rs
Antonio95: Good call, thank you. We tried that at some point and it failed because `step_by` returns a `StepBy` object which, oddly enough, cannot immediately be converted into `Range`.

However, `StepBy` does implement `Iterator` and so I have changed the argument to `bench_pcs_method` to expect an implementor of that - which is also more general.

primitives/benches/pcs.rs
```diff
@@ -4,100 +4,136 @@
...
+        "verify_kzg_range_BN_254",
+        verify::<Bn254>,
+    );
+}
+
+criterion_group! {
+    name = pcs_benches;
+    config = Criterion::default();
+    targets = kzg_254
```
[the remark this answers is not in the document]
Antonio95: Done! For time as well as size benches

PR #398, opened 2023-11-08
test: Expose more internal functions for test

alxiong:
These are the changes I made during writing tests for https://github.com/EspressoSystems/espresso-sequencer/pull/722, Most changes are under `test-api`, so should be safe

### DIFF
--- a/plonk/src/proof_system/structs.rs
+++ b/plonk/src/proof_system/structs.rs
@@ -60,26 +60,26 @@ pub type OpenKey<E> = UnivariateVerifierParam<E>;
 #[derivative(PartialEq, Hash(bound = "E:Pairing"))]
 pub struct Proof<E: Pairing> {
     /// Wire witness polynomials commitments.
-    pub(crate) wires_poly_comms: Vec<Commitment<E>>,
+    pub wires_poly_comms: Vec<Commitment<E>>,
 
     /// The polynomial commitment for the wire permutation argument.
-    pub(crate) prod_perm_poly_comm: Commitment<E>,
+    pub prod_perm_poly_comm: Commitment<E>,
 
     /// Splitted quotient polynomial commitments.
-    pub(crate) split_quot_poly_comms: Vec<Commitment<E>>,
+    pub split_quot_poly_comms: Vec<Commitment<E>>,
 
     /// (Aggregated) proof of evaluations at challenge point `zeta`.
-    pub(crate) opening_proof: Commitment<E>,
+    pub opening_proof: Commitment<E>,
 
     /// (Aggregated) proof of evaluation at challenge point `zeta * g` where `g`
     /// is the root of unity.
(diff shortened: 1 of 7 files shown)


plonk/src/proof_system/structs.rs
```diff
@@ -60,26 +60,26 @@ pub type OpenKey<E> = UnivariateVerifierParam<E>;
 #[derivative(PartialEq, Hash(bound = "E:Pairing"))]
 pub struct Proof<E: Pairing> {
     /// Wire witness polynomials commitments.
-    pub(crate) wires_poly_comms: Vec<Commitment<E>>,
+    pub wires_poly_comms: Vec<Commitment<E>>,
```
chancharles92, line 63: Minor & double-check: Would a user be able to get and modify it if it's not under `test_api`?

chancharles92 (commented):
LGTM, one minor question.

plonk/src/proof_system/structs.rs
alxiong: they would be able to get and modify, still, I don't see a problem there. 

My lines of reasoning is this:
- mark visibility to private or pub(crate) to start with
- when need cross-crate access, ask if this field/methods is so internal that doesn't make much sense to other modules
- if the field/method is _unsurprising_ even externally, then free to mark it as `pub`, as long as its type is also public (i.e. not some internal repr)
- finally, on more foolproof API principle (where we should limit internal changing surface so that downstream accidentally mess up/mutate an object in unintended way)
  - first, realize that they are just bytes, and even internal fields are marked private, if the downstream want to modify it (e.g. via desserializing from carefully designed bytes), they most likely can.
  - restrictive visibility mostly protect unintended read, rust's `mut` checker already protect against unintended mutation

chancharles92: The only concern I have is that if we change the struct field in the future, that would be breaking changes to the users that uses those struct fields in their code. Usually the best practice is only to set functions to public, or  set those struct fields that are very stable to public.

alxiong: I totally agree. 
I'm not a good judge of how stable these fields will be. they all look like quite unsurprising and standard for me (if I imagine myself being someone looking for a plonk impl, with the exception of `is_merged` field)

The alternative of defining getters for these fields is also no better, any changes to struct's field would still be breaking changes. (another word, getter is just a syntax sugar in our case, since we need to access most internal fields)


There is another possibility: define special getter/setter methods with documentation that these are for certain purposes and not for general public use? 

OTOH, we might overthink this, even arkworks's API can undergo such heavy refactoring, what's stopping us at our maturity level?

chancharles92: Yeah, fair enough. I think it's fine to keep them public then. Can you update the branch? I'll then approve.

chancharles92 (approved):

PR #409, opened 2023-11-11
fix: Incorrect aggregation of KZG benchmark results

tessico:
## Description

`commit` and `open` are unaffected much as they don't require precomputation to run the benchmarks. `verify` needs to first call `commit` and then `open` on the same poly, and the bench was calculating the total time to execute the method (i.e. commit + open + verify), not the elapsed time spent only in `verify`.

---

write a little note why.

- [ ] Wrote unit tests


primitives/benches/pcs.rs
```diff
@@ -39,7 +39,13 @@ pub fn bench_pcs_method<E: Pairing>(
             BenchmarkId::from_parameter(num_vars),
             &num_vars,
             |b, num_vars| {
-                b.iter(|| method(&pp, *num_vars));
+                b.iter_custom(|i| {
```
chancharles92, line 42: What's the semantic difference here? And what is i? (Maybe add some comments for the reader to better understand the logic).

tessico: I've added a comment about why we need `iter_custom`, trying to keep it brief. Some more info: https://bheisler.github.io/criterion.rs/book/user_guide/timing_loops.html.

mrain (approved):
LGTM. Feel free to merge after CI.

Issue #424, opened 2023-11-22
ADVZ PayloadProver support requests that span multiple polynomial


ggutoski:
# tl;dr

ADVZ PayloadProver currently fails if a proof request spans multiple polynomials. Remove this restriction.

https://github.com/EspressoSystems/jellyfish/blob/265eaaa059e0a501824e5aadf7ac5f32d9e4a6b8/primitives/src/vid/advz/payload_prover.rs#L380-L385

# Background

Distilled from [zulip](https://espresso.zulipchat.com/#narrow/stream/330967-Sequencer/topic/Namespace.20boundaries.20in.20jellyfish):

- In the past we vaguely intended that each namespace would correspond to 1 polynomials in the ADVZ payload partition. In this case the sequencer must take into account these polynomial boundaries when it builds the block payload bytes. ie. sequencer must leave zero padding between namespaces so that each namespace starts a new ADVZ polynomial.
- Complication: how to handle small namespaces? The vague intuition is to pack many small namespaces into a single polynomial. I guess it would be the sequencer who decides which namespaces get packed and which get their own polynomial.
- It's hard to do this without breaking abstraction: an abstract VID scheme doesn't know anything about polynomial boundaries.
- A hacky solution: defining an abstract notion of "alignment" in a general VID scheme. So we could have `next_alignment_boundary(index) -> Option<usize>`. A VID scheme that doesn't care about alignment can just return `None`, ADVZ can return the next polynomial boundary.
- Alternative: abandon the rule of 1 namespace per polynomial and just accept the fact that jellyfish namespace proofs might not align well with polynomial boundaries.

## Conclusion

Ignore polynomial boundaries for now. We'll revisit later if we run into perf issues.

PR #438, opened 2023-11-29
feat: ADVZ PayloadProver support requests that span multiple polynomial

ggutoski:
## Description

closes: #424 

Naive implementation:
- `SmallRangeProof`: concatenate individual KZG proofs from multiple polynomials, do not aggregate
- `LargeRangeProof`: rebuild and verify multiple KZG commitments. There's not much more we could do to improve here. But verification inside a smart contract is now more complex because it might need to compute multiple KZG commitments instead of one.

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/primitives/src/vid/advz/payload_prover.rs
+++ b/primitives/src/vid/advz/payload_prover.rs
@@ -89,6 +89,8 @@ where
         let range_elem_byte = self.range_elem_to_byte_clamped(&range_elem, payload.len());
         let range_poly_byte = self.range_poly_to_byte_clamped(&range_poly, payload.len());
         let offset_elem = self.offset_poly_to_elem(range_poly.start, range_elem.start);
+        let final_points_range_end =
+            self.final_poly_points_range_end(range_elem.len(), offset_elem);
 
         // prepare list of input points
         // perf: we might not need all these points
@@ -107,7 +109,7 @@ where
                 start: if i == 0 { offset_elem } else { 0 },
                 // final polynomial? stop at the end of the proof range
                 end: if i == range_poly.len() - 1 {
-                    (range_elem.len() + offset_elem - 1) % points.len() + 1
+                    final_points_range_end
                 } else {
                     points.len()
                 },
@@ -158,6 +160,8 @@ where
         let range_elem = self.range_byte_to_elem(&proof.chunk_range);
         let range_poly = self.range_elem_to_poly(&range_elem);
         let offset_elem = self.offset_poly_to_elem(range_poly.start, range_elem.start);
+        let final_points_range_end =
+            self.final_poly_points_range_end(range_elem.len(), offset_elem);
 
         // prepare list of input points
         // perf: we might not need all these points
@@ -174,7 +178,7 @@ where
                 start: if i == 0 { offset_elem } else { 0 },
                 // final polynomial? stop at the end of the proof range
                 end: if i == range_poly.len() - 1 {
-                    (range_elem.len() + offset_elem - 1) % points.len() + 1
+                    final_points_range_end
                 } else {
                     points.len()
                 },
@@ -319,6 +323,9 @@ where
             self.payload_chunk_size * elem_byte_capacity::<KzgEval<E>>(),
         );
         range_elem_start - index_coarsen(start_poly_byte, elem_byte_capacity::<KzgEval<E>>())
+    }
+    fn final_poly_points_range_end(&self, range_elem_len: usize, offset_elem: usize) -> usize {
+        (range_elem_len + offset_elem - 1) % self.payload_chunk_size + 1
     }
 
     // arg check helpers
--- a/primitives/src/vid/advz/payload_prover.rs
+++ b/primitives/src/vid/advz/payload_prover.rs
@@ -28,10 +28,11 @@ use crate::{
         vid, VidError, VidScheme,
     },
 };
+use anyhow::anyhow;
 use ark_ec::pairing::Pairing;
 use ark_poly::EvaluationDomain;
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
-use ark_std::{format, ops::Range};
+use ark_std::{format, ops::Range, println};
 use itertools::Itertools;
 use jf_utils::canonical;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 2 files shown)


chancharles92:
> * `LargeRangeProof`: rebuild and verify multiple KZG commitments. There's not much more we could do to improve here. But verification inside a smart contract is now more complex because it might need to compute multiple KZG commitments instead of one.

Can the smart contract just generate a random scalar and randomly combine the field vectors? Finally, it only needs to compute a single KZG commitment and compare it with the random linear combinations of the original commitments.

ggutoski:
> Can the smart contract just generate a random scalar and randomly combine the field vectors? Finally, it only needs to compute a single KZG commitment and compare it with the random linear combinations of the original commitments.

Interesting idea. Basically it replaces subsequent MSMs with a faster hash of the field vectors.

In any case I don't want to do it in this PR. 😉

primitives/src/vid/advz/payload_prover.rs
```diff
@@ -479,54 +475,95 @@ mod tests {
...
+                    let stmt = Statement {
+                        payload_subslice: &payload[range.clone()],
+                        range: range.clone(),
+                        commit: &d.commit,
+                        common: &d.common,
+                    };
+
+                    let small_range_proof: SmallRangeProof<_> =
+                        advz.payload_proof(&payload, range.clone()).unwrap();
+                    advz.payload_verify(stmt.clone(), &small_range_proof)
```
chancharles92, line 506: Add one more bad path where a wrong proof gets rejected

primitives/src/vid/advz/payload_prover.rs
```diff
@@ -479,54 +475,95 @@ mod tests {
...
+                        .unwrap()
+                        .unwrap();
+
+                    let large_range_proof: LargeRangeProof<_> =
+                        advz.payload_proof(&payload, range.clone()).unwrap();
+                    advz.payload_verify(stmt, &large_range_proof)
```
chancharles92, line 512: Same here, add a bad path test.

primitives/src/vid/advz/payload_prover.rs
```diff
@@ -84,31 +86,40 @@ where
...
-
-        // grab the polynomial that contains `range`
-        // TODO allow precomputation: https://github.com/EspressoSystems/jellyfish/issues/397
-        let polynomial = self.polynomial(
-            bytes_to_field::<_, KzgEval<E>>(payload[start_namespace_byte..].iter())
-                .take(self.payload_chunk_size),
-        );
+        let range_poly_byte = self.range_poly_to_byte_clamped(&range_poly, payload.len());
+        let offset_elem = self.offset_poly_to_elem(range_poly.start, range_elem.start);
+        let final_points_range_end =
```
chancharles92, line 92: For the case where the range goes across n >= 3 polys, are there any advantages of using `SmallRangeProof`? 
My argument is that if the range covers at least n-2 >= 1 *full* polys, compared to `LargeRangeProof`, the proof size and the verification time of `SmallRangeProof` won't be much more efficient *even if we use KZG multiproof in the future*. This is because the proof size is at least the number of evaluations (even if the KZG proof is O(1)-sized), and the verification is more than O(n*|poly_deg|) G-ops. (Recall that the KZG multi-verification still needs O(|eval_points|) group operations.)

chancharles92 (changes requested):

chancharles92:
> Interesting idea. Basically it replaces subsequent MSMs with a faster hash of the field vectors.
> 
> In any case I don't want to do it in this PR. 😉
In the interactive smart contract setting, we can even generate the randomness via oracles without doing Fiat-Shamir. But yes, there's no need to implement that for now.

primitives/src/vid/advz/payload_prover.rs
ggutoski: > ...the proof size is at least the number of evaluations (even if the KZG proof is O(1)-sized)

Why is that? A `SmallRangeProof` with constant-size KZG proof is constant-sized. The remaining fields are all constant-size. (The length of `prefix_bytes` and `suffix_bytes` is always less than 32.)

> ...the verification is more than O(n*|poly_deg|) G-ops.

Yes but I wonder whether the n-factor could be eliminated by aggregating multiproofs across multiple polynomials as described [here](https://dankradfeist.de/ethereum/2021/06/18/pcs-multiproofs.html). But that's a cutting edge optimization that will need to wait until later.

In any case, I doubt there is any advantage to the use of `SmallRangeProof` for large ranges--that's what `LargeRangeProof` is for. If your range spans multiple polynomials then `LargeRangeProof` is probably a better choice.

Is there a terminology misunderstanding? "Small" here refers to the size of the *range* of payload bytes that will be proven, not the size of the *proof* itself. (ie. [small range] proof vs. small [range proof] 😛 ) We tried to choose names that generalize beyond the narrow tx-namespace application. The intention is that `Small` should be used for tx proofs (typically only a small number of bytes) in a setting where a pairing is feasible (such as a off-chain light client). By contrast, `Large` should be used for namespace proofs (typically a large number of bytes) in a setting where a pairing is impractical (such as on-chain light client). Perhaps we could change these names.

primitives/src/vid/advz/payload_prover.rs
ggutoski: Done. See https://github.com/EspressoSystems/jellyfish/pull/438#discussion_r1416200681

primitives/src/vid/advz/payload_prover.rs
chancharles92: > Is there a terminology misunderstanding?

Oh the understanding of terminology is correct. My thought was: if `range` in `SmallRangeProof::payload_proof()` is not small (e.g., if it goes across >= 3 polynomials), the API should directly return an error and ask the user to use `LargeRangeProof::payload_proof()` instead (because there is no advantage of using `SmallRangeProof::payload_proof()` in this case). While in the current code, it is still dealing with this case and returns a valid but inefficient proof.

> Why is that? A SmallRangeProof with constant-size KZG proof is constant-sized. The remaining fields are all constant-size. (The length of prefix_bytes and suffix_bytes is always less than 32.)

Because anyway you need to send the txs content to the verifier (my understanding of the so-called `proof` is `data + auxiliary_proof`). E.g., if the tx data is exactly a polynomial, in the `LargeRangeProof` case, you don't need to send any extra info (as the tx itself is already a proof); while here you need to additionally send an O(1)-sized KZG multiproof.

chancharles92 (commented):
LGTM, will approve after addressing the comment in https://github.com/EspressoSystems/jellyfish/pull/438#discussion_r1416277670

primitives/src/vid/advz/payload_prover.rs
ggutoski: Thanks for clarifying. 

> My thought was: if `range` in `SmallRangeProof::payload_proof()` is not small (e.g., if it goes across >= 3 polynomials), the API should directly return an error and ask the user to use `LargeRangeProof::payload_proof()` instead (because there is no advantage of using `SmallRangeProof::payload_proof()` in this case). While in the current code, it is still dealing with this case and returns a valid but inefficient proof.

An alternative is to add another impl of `PayloadProver` that intelligently selects the proof type based on the user's input. The return type would be an enum with `Small` and `Large` variants. The existing impls for `Small` and `Large` remain as-is so that the user retains the ability to enforce a preference for `Small` or `Large`. (Example: perhaps the user always wants `Large` because it's pairing-free. I can't think of a reason to enforce `Small` at this time but you never know.)

> ... in the `LargeRangeProof` case, you don't need to send any extra info (as the tx itself is already a proof)

A `LargeRangeProof` needs enough auxiliary data to complete a polynomial. This auxiliary data could be quite large if for example the user's range just barely crosses into another polynomial. But yes, if the data range coincides with a polynomial then there's near-zero overhead.

# Proposal

I propose we merge as-is and punt this question to a new issue. This discussion is about performance but our current goal is feature-completeness. Any opinions?

chancharles92 (approved):

primitives/src/vid/advz/payload_prover.rs
ggutoski: New issue #441

PR #448, opened 2023-12-11
fix: ADVZ VID perf improvements and security patches (see linked issues)

### LINKED ISSUE #400: ADVZ VID parallelize computation of commitments
blocked by #399 

subtask of #394 ?

- Use `batch_commit` to compute KZG commitments in parallel
- Add a `bach_commit_iter` method to `UnivariateKzgPCS` that takes an iterator over polynomials instead of a slice. That's what I start with in the ADVZ use case; I would need to collect it into a vec just for this call anyway. Use [ParallelBridge in rayon::iter - Rust](https://docs.rs/rayon/latest/rayon/iter/trait.ParallelBridge.html) for the parallel iterator.

### LINKED ISSUE #443: Remove `chunk_range` from `Small` and `LargeRangeProof`
`chunk_range` is useless and wastes space in the proof. We require a copy of this range as part of `Statement`, and we only reference `chunk_range` to check for equality with the range given in the `Statement`. This issue spawned from https://github.com/EspressoSystems/espresso-sequencer/pull/796#discussion_r1417577302

### LINKED ISSUE #444: ADVZ add payload bytes length to commitment
Currently the ADVZ VID commitment includes only the KZG polynomial commitments derived from the payload bytes, but not the payload byte length. An attacker could modify `Common::bytes_len` and cause the payload recovery process to recover a payload whose byte length differs from the original. Also, downstream users might want cryptographic assurance of the precise payload byte length.

### LINKED ISSUE #445: ADVZ PayloadProver::payload_verify should check `stmt.range` against payload length
Currently payload_verify does not verify that `stmt.range` is within bounds of `common.bytes_len`. Otherwise it's possible to make a proof that passes verification even if `stmt.range` is outside the payload. Needs #444 to be secure.

ggutoski:
# Description

close #400 
close #443 
close #444 
close #445

- Each issue is tagged in the git commit message that closes it.
- #400 use `UnivariateKzgPCS::batch_commit()`, which already has parallelism
  - ~2x speed-up! See benches below.
  - `commit_only()` now use `derive_commit()` just like `disperse()`.
- #443 minor communication improvement
- #444 
- #445 + minor refactoring
- no new tests
- Question: `Common` has a member `bytes_len` of type `usize`. This struct is serialized and sent over the wire. Should we avoid use of `usize` in such structs? Should we instead use something like `u32` where we can enforce a fixed size?

# Benches for `commit_only`

How to run:
```bash
cargo test --release --package jf-primitives --lib --features test-srs --features print-trace -- vid::advz::tests::commit_only_timer --exact --nocapture
```
Be sure to uncomment `#[test]` for `commit_only_timer` test.

## Before parallelism

```
Start:   KZG10::Setup with prover degree 256 and verifier degree 1
··Start:   Generating powers of G
··End:     Generating powers of G ..................................................5.894ms
End:     KZG10::Setup with prover degree 256 and verifier degree 1 .................9.540ms
Start:   VID commit_only 1048576 payload bites to 512 nodes
End:     VID commit_only 1048576 payload bites to 512 nodes ........................520.926ms
test vid::advz::tests::commit_only_timer ... ok
```

## After parallelism

```
Start:   KZG10::Setup with prover degree 256 and verifier degree 1
··Start:   Generating powers of G
··End:     Generating powers of G ..................................................4.751ms
End:     KZG10::Setup with prover degree 256 and verifier degree 1 .................9.268ms
Start:   batch commit 133 polynomials
End:     batch commit 133 polynomials ..............................................239.522ms
test vid::advz::tests::commit_only_timer ... ok
```

---

write a little note why.

- [ ] Wrote unit tests


chancharles92 (approved):
LGTM

primitives/src/vid/advz/payload_prover.rs
```diff
@@ -237,7 +232,6 @@ where
             suffix_elems,
             prefix_bytes: payload[range_elem_byte.start..range.start].to_vec(),
             suffix_bytes: payload[range.end..range_elem_byte.end].to_vec(),
-            chunk_range: range,
         })
     }
```
chancharles92, line 237: Should we rename the file `payload_prover` if it also contains verification APIs?

ggutoski: I'm open to that but I can't think of a name I like more. `payload_prover_and_verifier.rs` is too long. Any suggestion? I thought maybe "prover" in the name suggests that there's also a verifier included, but I admit that's not necessarily obvious.

I have another PR coming with a naming challenge of its own. Perhaps it makes sense to do a rename in the next PR. I'll merge now that I have approval.

Issue #460, opened 2024-01-15
`VidScheme::Common` expose getter for number of storage nodes


ggutoski:
It's convenient to use existing network traffic to reconstruct VID scheme parameters (ie. number of storage nodes, etc). To this end we should store this info in `Common` and provide a getter for it. Something like this except for number of storage nodes:
https://github.com/EspressoSystems/jellyfish/blob/8d071e2df74e5093d2b3c35c32773b09370c2452/primitives/src/vid.rs#L75-L76

PR #472, opened 2024-01-24
feat: new trait method VidScheme::get_num_storage_nodes

ggutoski:
## Description

closes: #460 

Does what it says on the tin.

# Question

Currently data such as [payload byte length], [number of storage nodes] is stored in `Common` as `usize`. But `Common` is a serializable struct. I'm uneasy about putting `usize` in such a struct because the guarantees you get are less clear and it could waste space. This struct will be sent over the wire and every byte counts. Shall we pick a fixed word size here like `u16`, `u32`? (We could even use 3 bytes.)

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/primitives/src/vid/advz.rs
+++ b/primitives/src/vid/advz.rs
@@ -384,7 +384,8 @@ where
                 common.poly_commits.len()
             )));
         }
-        if common.num_storage_nodes != self.num_storage_nodes.try_into().map_err(vid)? {
+        let num_storage_nodes: u32 = self.num_storage_nodes.try_into().map_err(vid)?; // pacify cargo check --target wasm32-unknown-unknown --no-default-features
+        if common.num_storage_nodes != num_storage_nodes {
             return Err(VidError::Argument(format!(
                 "common num_storage_nodes differs from self ({},{})",
                 common.num_storage_nodes, self.num_storage_nodes
@@ -448,7 +449,8 @@ where
                 self.payload_chunk_size
             )));
         }
-        if common.num_storage_nodes != self.num_storage_nodes.try_into().map_err(vid)? {
+        let num_storage_nodes: u32 = self.num_storage_nodes.try_into().map_err(vid)?; // pacify cargo check --target wasm32-unknown-unknown --no-default-features
+        if common.num_storage_nodes != num_storage_nodes {
             return Err(VidError::Argument(format!(
                 "common num_storage_nodes differs from self ({},{})",
                 common.num_storage_nodes, self.num_storage_nodes
--- a/primitives/src/vid/advz.rs
+++ b/primitives/src/vid/advz.rs
@@ -27,7 +27,9 @@ use ark_poly::{DenseUVPolynomial, EvaluationDomain, Radix2EvaluationDomain};
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 use ark_std::{
     borrow::Borrow,
-    end_timer, format,
+    end_timer,
+    fmt::Debug,
+    format,
     marker::PhantomData,
     ops::{Add, Mul},
     start_timer, vec,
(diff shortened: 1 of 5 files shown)


primitives/src/vid/advz.rs
```diff
@@ -190,7 +190,8 @@ where
     #[serde(with = "canonical")]
     all_evals_digest: KzgEvalsMerkleTreeNode<E, H>,
 
-    bytes_len: usize, // TODO don't use usize in serializable struct?
+    bytes_len: usize,         // TODO don't use usize in serializable struct?
+    num_storage_nodes: usize, // TODO don't use usize in serializable struct?
```
jbearer, line 194: I wonder if we could infer either/both of these from the number of polynomial commitments and number of evals (height and width of matrix?)

primitives/src/vid/advz.rs
```diff
@@ -483,6 +485,10 @@ where
     fn get_payload_byte_len(common: &Self::Common) -> usize {
         common.bytes_len
     }
+
+    fn get_num_storage_nodes(common: &Self::Common) -> usize {
+        common.num_storage_nodes
```
jbearer, line 490: Don't we have to verify somewhere that this is consistent with the actual matrix encoding of the data? I would expect to see changes somewhere in `verify_share` or something. Same question also for `bytes_len`

primitives/src/vid/advz.rs
jbearer: And yes, best to avoid `usize` in serialized structs. I don't think we're at the point where we're fighting for bytes in a succinct fixed-size type like this, `u64` is probably fine. Although in practice, both of these will probably never be bigger than a `u32`

ggutoski: > I wonder if we could infer either/both of these from the number of polynomial commitments and number of evals (height and width of matrix?)

Possible. Number of polynomial commitments is readily available in `Common`. Number of evals... is this info readily available to all nodes?

Note that in the near future we will allow number of polynomials to decouple from number of evals for optimization reasons. Of course, the relevant ratio should be known to all but the calculation gets a bit more obfuscated.

ggutoski: > Number of evals... is this info readily available to all nodes?

Yes. In hotshot, at least, it equals the number of quorum nodes. So everybody knows it.

ggutoski: But note that the number of evals is not readily available from a `Common` or a single `Share`. So it's up to the downstream user to do the accounting required for this.

primitives/src/vid/advz.rs
ggutoski: Good questions.

# Number of storage nodes

We can check this against the VID scheme params in any method that takes both `self` and `common`. Currently that's `verify_share` and `recover_payload`. I'll add this.

I just noticed that this quantity should be hashed into the commitment like we already do for `bytes_len`:
https://github.com/EspressoSystems/jellyfish/blob/bf1a6dfb63e2ce9b409a682d2793b836f1659097/primitives/src/vid/advz.rs#L543-L547
I'll add this.

# Payload byte length

The exact byte length would be lost in `disperse` if we did not save it in `Common`. The matrix gives you only an upper bound on the byte length, as we pad the payload with zero-bytes to fill the matrix. It is used within jellyfish only in `recover_payload` so that we can truncate the recovered payload to the correct length.

I don't see anything we could do here to enforce a declaration by the disperser of the "correct" byte length.

primitives/src/vid/advz.rs
jbearer: Ah ok, we don't want this

primitives/src/vid/advz.rs
jbearer: Ah I see. The correct byte length is just a convention, so we don't have to enforce it, we just need agreement. Maybe we should enforce that it is plausible, though (ie less than or equal to the padded byte length)

jbearer: Sorry, this is out of scope for this PR. I'll approve and you can decide whether to add this stuff now or in a separate PR

jbearer (approved):

primitives/src/vid/advz.rs
ggutoski: > Ah I see. The correct byte length is just a convention, so we don't have to enforce it, we just need agreement. Maybe we should enforce that it is plausible, though (ie less than or equal to the padded byte length)

The only place where we could execute this enforcement is in `verify_share` by rejecting a share if the claimed payload byte len is larger than what's possible given [num polynomials] and [polynomial degree]. But is this really worth it? An alternative is to simply *define* the payload byte length as the *minimum* of (i) what's possible given [num polynomials] and [polynomial degree], and (ii) the declared byte len in `Common`. ie. If it's too big then `truncate` does nothing. Complexity is reduced.

ggutoski: Punted to #473 . I'm happy as-is for now, at least.

ggutoski:
[CI build failure](https://github.com/EspressoSystems/jellyfish/actions/runs/7645456432/job/20832078339) I did not notice locally because I don't build locally with `print-trace` feature 😬 .

philippecamacho (approved):
LGTM

Issue #465, opened 2024-01-16
Add `leaves()` function for `UniversalMerkleTree`


jbearer:
`NMT` has [this function](https://jellyfish.docs.espressosys.com/jf_primitives/merkle_tree/namespaced_merkle_tree/struct.NMT.html#method.leaves) which allows us to iterate over the leaves. In the sequencer, we use this to [serialize](https://github.com/EspressoSystems/espresso-sequencer/blob/main/sequencer/src/block.rs#L293-L294) an NMT as just a list of leaves, and [reconstruct the tree structure](https://github.com/EspressoSystems/espresso-sequencer/blob/main/sequencer/src/block.rs#L303-L305) upon deserialization. This has two advantages:
1. The serialization is much smaller since we aren't including all the internal nodes
2. The deserializer doesn't trust the serializer to provide a well-formed Merkle tree

We would like to do something similar with `UniversalMerkleTree` for the fee state. For this we need a way to iterate over the leaves.

cc @tbro

mrain:
Will deal with https://github.com/EspressoSystems/espresso-sequencer/pull/989#discussion_r1453715569 as well

PR #475, opened 2024-01-29
feat: ergonomics changes for merkle tree

mrain:
## Description

closes: #465 

- Remove the unnecessary trait bound `From<u64>` to the merkle tree index
- A new `update_with` interface for the universal merkle tree as requested here
  - https://github.com/EspressoSystems/espresso-sequencer/pull/989#discussion_r1453715569
- A new `iter()` interface for Merkle tree scheme, allows user to iterate through all elements that's in memory.

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/primitives/src/merkle_tree/append_only.rs
+++ b/primitives/src/merkle_tree/append_only.rs
@@ -25,7 +25,7 @@ use num_traits::pow::pow;
 use serde::{Deserialize, Serialize};
 use typenum::Unsigned;
 
-impl_merkle_tree_scheme!(MerkleTree, build_tree_internal);
+impl_merkle_tree_scheme!(MerkleTree);
 impl_forgetable_merkle_tree_scheme!(MerkleTree);
 
 impl<E, H, I, Arity, T> AppendableMerkleTreeScheme for MerkleTree<E, H, I, Arity, T>
@@ -36,6 +36,19 @@ where
     Arity: Unsigned,
     T: NodeValue,
 {
+    fn from_elems(
+        height: usize,
+        elems: impl IntoIterator<Item = impl Borrow<Self::Element>>,
+    ) -> Result<Self, PrimitivesError> {
+        let (root, num_leaves) = build_tree_internal::<E, H, I, Arity, T>(height, elems)?;
+        Ok(Self {
+            root,
+            height,
+            num_leaves,
+            _phantom: PhantomData,
+        })
+    }
+
     fn push(&mut self, elem: impl Borrow<Self::Element>) -> Result<(), PrimitivesError> {
         <Self as AppendableMerkleTreeScheme>::extend(self, [elem])
     }
--- a/primitives/src/merkle_tree/append_only.rs
+++ b/primitives/src/merkle_tree/append_only.rs
@@ -10,7 +10,8 @@ use core::ops::AddAssign;
 
 use super::{
     internal::{
-        build_tree_internal, MerkleNode, MerkleProof, MerkleTreeCommitment, MerkleTreeIter,
+        build_tree_internal, MerkleNode, MerkleProof, MerkleTreeCommitment, MerkleTreeIntoIter,
+        MerkleTreeIter,
     },
     AppendableMerkleTreeScheme, DigestAlgorithm, Element, ForgetableMerkleTreeScheme, Index,
     LookupResult, MerkleCommitment, MerkleTreeScheme, NodeValue, ToTraversalPath,
--- a/primitives/src/merkle_tree/append_only.rs
+++ b/primitives/src/merkle_tree/append_only.rs
@@ -6,8 +6,6 @@
 
 //! Implementation of a typical append only merkle tree
 
-use core::ops::AddAssign;
-
 use super::{
     internal::{
         build_tree_internal, MerkleNode, MerkleProof, MerkleTreeCommitment, MerkleTreeIntoIter,
@@ -35,7 +33,7 @@ impl<E, H, I, Arity, T> MerkleTree<E, H, I, Arity, T>
 where
     E: Element,
     H: DigestAlgorithm<E, I, T>,
-    I: Index + From<u64> + AddAssign + ToTraversalPath<Arity>,
+    I: Index,
     Arity: Unsigned,
     T: NodeValue,
 {
@@ -48,7 +46,15 @@ where
             _phantom: PhantomData,
         }
     }
+}
 
+impl<E, H, Arity, T> MerkleTree<E, H, u64, Arity, T>
+where
+    E: Element,
+    H: DigestAlgorithm<E, u64, T>,
+    Arity: Unsigned,
+    T: NodeValue,
+{
     /// Construct a new Merkle tree with given height from a data slice
     /// * `height` - height of the Merkle tree, if `None`, it will calculate the
     ///   minimum height that could hold all elements.
@@ -58,7 +64,7 @@ where
         height: Option<usize>,
         elems: impl IntoIterator<Item = impl Borrow<E>>,
     ) -> Result<Self, PrimitivesError> {
-        let (root, height, num_leaves) = build_tree_internal::<E, H, I, Arity, T>(height, elems)?;
+        let (root, height, num_leaves) = build_tree_internal::<E, H, Arity, T>(height, elems)?;
         Ok(Self {
             root,
             height,
@@ -68,11 +74,10 @@ where
     }
 }
 
-impl<E, H, I, Arity, T> AppendableMerkleTreeScheme for MerkleTree<E, H, I, Arity, T>
+impl<E, H, Arity, T> AppendableMerkleTreeScheme for MerkleTree<E, H, u64, Arity, T>
 where
     E: Element,
-    H: DigestAlgorithm<E, I, T>,
-    I: Index + From<u64> + AddAssign + ToTraversalPath<Arity>,
+    H: DigestAlgorithm<E, u64, T>,
     Arity: Unsigned,
     T: NodeValue,
 {
@@ -86,10 +91,11 @@ where
     ) -> Result<(), PrimitivesError> {
         let mut iter = elems.into_iter().peekable();
 
-        let traversal_path = I::from(self.num_leaves).to_traversal_path(self.height);
+        let traversal_path =
+            ToTraversalPath::<Arity>::to_traversal_path(&self.num_leaves, self.height);
         self.num_leaves += self.root.extend_internal::<H, Arity>(
             self.height,
-            &I::from(self.num_leaves),
+            &self.num_leaves,
             &traversal_path,
             true,
             &mut iter,
@@ -105,7 +111,6 @@ where
 
 // TODO(Chengyu): extract a merkle frontier
 
-// TODO(Chengyu): unit tests
 #[cfg(test)]
 mod mt_tests {
     use crate::{
@@ -183,7 +188,7 @@ mod mt_tests {
             unreachable!()
         }
 
-        let result = RescueMerkleTree::<F>::verify(&root, 0u64, &bad_proof);
+        let result = RescueMerkleTree::<F>::verify(&root, 0, &bad_proof);
         assert!(result.unwrap().is_err());
 
         let mut forge_proof = MerkleProof::new(2, proof.proof);
@@ -198,7 +203,7 @@ mod mt_tests {
         } else {
             unreachable!()
         }
-        let result = RescueMerkleTree::<F>::verify(&root, 2u64, &forge_proof);
+        let result = RescueMerkleTree::<F>::verify(&root, 0, &forge_proof);
         assert!(result.unwrap().is_err());
     }
 
@@ -242,7 +247,7 @@ mod mt_tests {
             unreachable!()
         }
 
-        let result = mt.remember(0u64, elem, &bad_proof);
+        let result = mt.remember(0, elem, &bad_proof);
         assert!(result.is_err());
 
         let mut forge_proof = MerkleProof::new(2, proof.proof.clone());
@@ -257,7 +262,7 @@ mod mt_tests {
         } else {
             unreachable!()
         }
-        let result = mt.remember(2u64, elem, &forge_proof);
+        let result = mt.remember(2, elem, &forge_proof);
         assert!(result.is_err());
 
         assert!(mt.remember(0, elem, &proof).is_ok());
@@ -312,7 +317,7 @@ mod mt_tests {
         assert_eq!(mt.num_leaves(), 3);
         // Leaves that are forgotten doesn't appear here
         let leaves = mt.into_iter().collect::<Vec<_>>();
-        assert_eq!(leaves, [(0u64, F::from(0u64)), (2u64, F::from(2u64))]);
+        assert_eq!(leaves, [(0, F::from(0u64)), (2, F::from(2u64))]);
 
         let kv_set = [
             (BigUint::from(64u64), F::from(32u64)),
--- a/primitives/src/merkle_tree/append_only.rs
+++ b/primitives/src/merkle_tree/append_only.rs
@@ -172,7 +172,7 @@ mod mt_tests {
         let mut mt = RescueMerkleTree::<F>::from_elems(2, &[F::from(3u64), F::from(1u64)]).unwrap();
         let root = mt.commitment().digest();
         let (lookup_elem, lookup_proof) = mt.lookup(0).expect_ok().unwrap();
-        let lookup_elem = lookup_elem.clone();
+        let lookup_elem = *lookup_elem;
         let (elem, proof) = mt.forget(0).expect_ok().unwrap();
         assert_eq!(lookup_elem, elem);
         assert_eq!(lookup_proof, proof);
--- a/primitives/src/merkle_tree/append_only.rs
+++ b/primitives/src/merkle_tree/append_only.rs
@@ -9,7 +9,9 @@
 use core::ops::AddAssign;
 
 use super::{
-    internal::{build_tree_internal, MerkleNode, MerkleProof, MerkleTreeCommitment},
+    internal::{
+        build_tree_internal, MerkleNode, MerkleProof, MerkleTreeCommitment, MerkleTreeIter,
+    },
     AppendableMerkleTreeScheme, DigestAlgorithm, Element, ForgetableMerkleTreeScheme, Index,
     LookupResult, MerkleCommitment, MerkleTreeScheme, NodeValue, ToTraversalPath,
 };
@@ -259,5 +261,19 @@ mod mt_tests {
             *node,
             bincode::deserialize(&bincode::serialize(node).unwrap()).unwrap()
         );
+    }
+
+    #[test]
+    fn test_mt_iter() {
+        test_mt_iter_helper::<Fq254>();
+        test_mt_iter_helper::<Fq377>();
+        test_mt_iter_helper::<Fq381>();
+    }
+
+    fn test_mt_iter_helper<F: RescueParameter>() {
+        let mt =
+            RescueMerkleTree::<F>::from_elems(2, &[F::from(0u64), F::from(1u64), F::from(2u64)])
+                .unwrap();
+        assert!(mt.iter().all(|(index, elem)| { elem == &F::from(*index) }));
     }
 }
--- a/primitives/src/merkle_tree/append_only.rs
+++ b/primitives/src/merkle_tree/append_only.rs
@@ -40,10 +40,10 @@ where
     T: NodeValue,
 {
     fn from_elems(
-        height: usize,
+        height: Option<usize>,
         elems: impl IntoIterator<Item = impl Borrow<Self::Element>>,
     ) -> Result<Self, PrimitivesError> {
-        let (root, num_leaves) = build_tree_internal::<E, H, I, Arity, T>(height, elems)?;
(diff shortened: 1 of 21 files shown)


primitives/src/merkle_tree/internal.rs
```diff
@@ -544,73 +544,105 @@ where
...
-    pub(crate) fn update_internal<H, Arity>(
+    pub(crate) fn update_with_internal<H, Arity, F>(
         &mut self,
         height: usize,
         pos: impl Borrow<I>,
         traversal_path: &[usize],
-        elem: impl Borrow<E>,
-    ) -> Result<LookupResult<E, (), ()>, PrimitivesError>
+        f: F,
+    ) -> Result<(i64, LookupResult<E, (), ()>), PrimitivesError>
```
jbearer, line 553: Document what this `i64` is

primitives/src/merkle_tree/internal.rs
```diff
@@ -544,73 +544,105 @@ where
...
+            } => match f(Some(node_elem)) {
+                Some(elem) => {
+                    let result = (0i64, LookupResult::Ok(node_elem.clone(), ()));
+                    *value = H::digest_leaf(pos, &elem)?;
+                    *node_elem = elem;
+                    Ok(result)
+                },
+                None => {
+                    *self = MerkleNode::Empty;
+                    Ok((-1i64, LookupResult::NotFound(())))
```
jbearer, line 574: Not clear what the `LookupResult` represents either. Here it looks like the result refers to the new tree (element is `NotFound` because we just deleted it) but above we returned the old value (`node_elem`)

primitives/src/merkle_tree/macros.rs
```diff
@@ -105,6 +99,14 @@ macro_rules! impl_merkle_tree_scheme {
...
                 proof.borrow().verify_membership_proof::<H>(root.borrow())
             }
+
+            fn iter(&self) -> MerkleTreeIter<Self::Element, Self::Index, Self::NodeValue> {
+                MerkleTreeIter::new(&self.root)
+            }
+
+            fn into_iter(self) -> MerkleTreeIntoIter<Self::Element, Self::Index, Self::NodeValue> {
+                MerkleTreeIntoIter::new(self.root)
+            }
```
jbearer, line 109: Slightly more idiomatic would be to implement `IntoIter<Item = (&I, &E)> for &Self` and `IntoIter<Item = (I, E)> for Self`, similar to how `Vec` does it

primitives/src/merkle_tree/internal.rs
mrain: Oh this is a silly bug. Will document it and fix it

primitives/src/merkle_tree/internal.rs
```diff
@@ -895,3 +927,89 @@ where
...
+
+impl<'a, E, I, T> Iterator for MerkleTreeIter<'a, E, I, T>
+where
+    E: Element,
+    I: Index,
+    T: NodeValue,
+{
+    type Item = (&'a I, &'a E);
+
+    fn next(&mut self) -> Option<Self::Item> {
```
tbro, line 951: Do I understand correct usage which would be analogous to previous `fn leaves()` (for serialization) would be to `match` the returned `Option` to filter empty fields? Or, since it is an `Iter` we would do `mt.iter().filter_map(|item|item)`. I'm specifically thinking of the case of the sparse merkle tree.

primitives/src/merkle_tree/macros.rs
mrain: Good idea, but will keep `iter()` because it's not supported by any traits. [937c573](https://github.com/EspressoSystems/jellyfish/pull/475/commits/937c57379424b074bfe5caf61d3b7538a8b8deb8)

primitives/src/merkle_tree/internal.rs
mrain: [937c573](https://github.com/EspressoSystems/jellyfish/pull/475/commits/937c57379424b074bfe5caf61d3b7538a8b8deb8)

primitives/src/merkle_tree/internal.rs
mrain: Documented in [937c573](https://github.com/EspressoSystems/jellyfish/pull/475/commits/937c57379424b074bfe5caf61d3b7538a8b8deb8)

primitives/src/merkle_tree/internal.rs
mrain: It's an `Iter`. So just use `mt.iter().collect()` or `mt.into_iter().collect()` depending on whether you want to consume the merkle tree ownership.
All empty leaves are filtered out by default.

primitives/src/merkle_tree/internal.rs
```diff
@@ -544,73 +544,111 @@ where
...
-                // If the branch containing the update was not in memory, the update failed and
-                // nothing was changed, so we can short-circuit without recomputing this node's
-                // value.
-                if res == LookupResult::NotInMemory {
-                    return Ok(res);
+                if matches!(*children[branch], MerkleNode::ForgettenSubtree { .. }) {
+                    // If the branch containing the update was not in memory,
+                    // the update failed and nothing was
+                    // changed, so we can short-circuit without recomputing this
+                    // node's value.
```
ggutoski, line 595: My confusion: if something is "forgotten" or "not in memory" is that because it was removed by the user? Or is this element still in the tree but it's been swapped to cache? My guess is the former. In that case, maybe use "deleted/removed by user" or something.

ggutoski (commented):
I did only a quick skim but looks great! Minor comment below, plus the request I posted offline. Thank yoU!!

primitives/src/merkle_tree/internal.rs
mrain: Change to "forgotten by user".
The element remains in the tree (for commitment), but, like frontier, we do not track it in this MT structure.

ggutoski: Now I'm even more confused. 🤪 Let C be the commitment. I remove a node but C does not change, correct? Then I serialize and deserialize, so "forgotten" nodes are now truly gone. Then I expect the commitment to change to something different from C.  Shouldn't the commitment be consistent across serialization?

mrain: `forget` is something that you remove a not from memory, but not from the tree.
(De)serialization is w.r.t the whole tree, information won't get lost

ggutoski (dismissed):
Approve with disclaimer: my review was not deep. If you want deeper review from me then it will need to wait until tomorrow. 👋 🙂

primitives/src/merkle_tree/internal.rs
ggutoski: I think perhaps I've overlooked some code comments / rusdoc somewhere that explains this concept of "forget". Apologies and please point me to the right place.

My understanding so far:
- User may delete an item from the tree. Under the hood, such an item is "forgotten", but it's hash still contributes to the merkle root so that the root does not change even after the deletion.
- (de)serialization must remember these "forgotten" items so that the tree can be rebuilt to the same merkle root.

This seems counterintuitive to me. I expect a MT with leaves [A,B,C] to have a unique, deterministic root, independent of whether that MT was built directly from [A,B,C] vs [A,B,C,D] + delete D.

primitives/src/merkle_tree/mod.rs
```diff
@@ -196,11 +198,8 @@ pub trait MerkleTreeScheme: Sized {
     /// Tree arity
     const ARITY: usize;
 
-    /// Construct a new merkle tree with given height from a data slice
-    fn from_elems(
-        height: usize,
-        elems: impl IntoIterator<Item = impl Borrow<Self::Element>>,
-    ) -> Result<Self, PrimitivesError>;
+    /// Initialize an empty Merkle tree
+    fn new(height: usize) -> Self;
```
alxiong, line 203: ## Open to discussion/debate

A big consequence of removing `from_elems` is that we can no longer construct a struct that _only_ implements `MerkleTreeScheme` trait, which makes this trait useless. 
Sometimes, we just want a fixed-non-appendable MT. 

I understand that removing this originates from the intention to remove `Index: From<u64>` trait bound which is not pretty, but compared to the non-implementable trait, I prefer to swallow this pill.

primitives/src/merkle_tree/mod.rs
```diff
@@ -241,11 +240,25 @@ pub trait MerkleTreeScheme: Sized {
...
 /// append-only vector.
 pub trait AppendableMerkleTreeScheme: MerkleTreeScheme {
+    /// Construct a new Merkle tree with given height from a data slice
+    /// * `height` - height of the Merkle tree, if `None`, it will calculate the
+    ///   minimum height that could hold all elements.
+    /// * `elems` - an iterator to all elements
+    /// * `returns` - A constructed Merkle tree, or `Err()` if errors
+    fn from_elems(
+        height: Option<usize>,
+        elems: impl IntoIterator<Item = impl Borrow<Self::Element>>,
```
alxiong, line 259: since we are going to compute the required `height` regardless of if it's None or Some, why ask? (are we meant to indicate an upperbound `capacity`?)

primitives/src/merkle_tree/mod.rs
```diff
@@ -291,14 +304,37 @@ pub trait UniversalMerkleTreeScheme: MerkleTreeScheme {
...
+    /// * `returns` - Err() if any error occurs internally. Ok(result) if the
+    ///   update is success or the given leaf is not in memory.
+    fn update_with<F>(
+        &mut self,
+        pos: impl Borrow<Self::Index>,
+        f: F,
+    ) -> Result<LookupResult<Self::Element, (), ()>, PrimitivesError>
+    where
+        F: FnOnce(Option<&Self::Element>) -> Option<Self::Element>;
+
```
alxiong, line 337: why do we assume the update function only takes the old value? would it be safer to assume that it's a function of lots many other tree info, 
e.g. I only increment the fifth slot if we have more than 10 leaves with odd value. (i admit, a contrived logic)

my point being, why restrict ourselves to this API, why couldn't we just accept the `new_value: Self::Element` and let the caller figure out the derivation of the new value?

primitives/src/merkle_tree/universal_merkle_tree.rs
```diff
@@ -285,6 +299,30 @@ mod mt_tests {
...
+            mt.update_with(F::from(i as u64), |elem| match elem {
+                Some(elem) => Some(*elem),
+                None => Some(F::from(i as u64)),
+            })
+            .unwrap();
+        }
+        assert_eq!(mt.num_leaves(), 10);
+        // test lookup at index 7
+        let (val, proof) = mt.universal_lookup(F::from(7u64)).expect_ok().unwrap();
+        assert_eq!(val, &F::from(7u64));
```
alxiong, line 312: why don't we test all pos that are updated?
also how are sure they are new value? I thought leaf at pos `i` was originally also initiated as `F::from(i)`?

primitives/src/merkle_tree/universal_merkle_tree.rs
```diff
@@ -285,6 +299,30 @@ mod mt_tests {
...
+                .unwrap()
+                .is_ok()
+        );
+
+        // Remove index 8
+        mt.update_with(F::from(8u64), |_| None).unwrap();
+        assert!(mt
+            .universal_lookup(F::from(8u64))
+            .expect_not_found()
+            .is_ok());
```
alxiong, line 325: should we also check other `mt.num_leaves()` is also correctly updated?

primitives/src/merkle_tree/append_only.rs
```diff
@@ -247,4 +266,20 @@ mod mt_tests {
...
+        test_mt_iter_helper::<Fq381>();
+    }
+
+    fn test_mt_iter_helper<F: RescueParameter>() {
+        let mt = RescueMerkleTree::<F>::from_elems(
+            Some(2),
+            &[F::from(0u64), F::from(1u64), F::from(2u64)],
+        )
+        .unwrap();
+        assert!(mt.iter().all(|(index, elem)| { elem == &F::from(*index) }));
```
alxiong, line 283: this test is slightly too hardcoded? can we maybe have 10 test for a randomly generated list of some size (randomly ranged from [10, 50], sth like that), 

also should we add test for MT that contains empty subtree, or forgotten subtree, and showcase what our iterator will do?

primitives/src/merkle_tree/mod.rs
mrain: Yes for capacity.
You could start with some less elements and eventually have a large tree

primitives/src/merkle_tree/universal_merkle_tree.rs
mrain: Index 7 is not updated previously

primitives/src/merkle_tree/append_only.rs
mrain: IMHO we don't need such random test as the code doesn't manipulate the value. But sure I'll add a test for forgotten subtree.

primitives/src/merkle_tree/universal_merkle_tree.rs
mrain: added [bd2bc19](https://github.com/EspressoSystems/jellyfish/pull/475/commits/bd2bc19dcc34a7b0647794c15cc820ff0fa27c79)

primitives/src/merkle_tree/append_only.rs
mrain: strengthened test [bd2bc19](https://github.com/EspressoSystems/jellyfish/pull/475/commits/bd2bc19dcc34a7b0647794c15cc820ff0fa27c79)

primitives/src/merkle_tree/mod.rs
mrain: I prefer the new design as it looks cleaner.
Also we could rename the `AppendableMerkleTreeScheme` to `VecMerkleTreeScheme` or something to avoid confusion

primitives/src/merkle_tree/mod.rs
alxiong: ah, right, my mistake, anonymous functions can `move` contextual variable into the body, I forgot.

primitives/src/merkle_tree/mod.rs
alxiong: ask for more opinions here @ggutoski @jbearer @tbro 

(i have a slight preference to keeping `MTScheme::from_elems()`)

ggutoski: I don't know why `from_elems` requires `From<u64>` trait bound on `Index`. Even after moving `from_elems` to `AppendableMerkleTreeScheme`, we still have a mysterious `From<u62>` trait bound. Is there no way to remove this requirement?

# On constructor methods in traits

I have not yet seen a "correct" universal rule on whether/how to put constructor methods in traits.

Example: I had a similar problem with `VidScheme` trait. Different VID schemes require different things for construction. (Example: ADVZ requires polynomial degree & KZG params. Other VID schemes might not need these things.) I did not put a constructor in `VidScheme`.

It's still possible to construct a bare `MerkleTreeScheme`, but you need to isolate construction. Example: have a tiny module `M` that exposes a merkle tree "factory" function with return type `impl MerkleTreeScheme`. Then `M` is the only place that "knows" the underlying implementation of `MerkleTreeScheme`; everything else talks to it via this trait, maintaining modularity/abstraction. It seems I'll need to do something like this for `VidScheme`.

For merkle trees, we're starting to see demand for different constructors. eg. `from_elems`, whether to require a `height` arg, etc. It's hard to accommodate all these variations in a universal constructor method.

ggutoski: Another option: `from_elems` could take an `Iterator<Item = (Self::Index, impl Borrow<Self::Element>)` and force the caller to make `Index`s for you. This does not require `From<u64>`. But it is less ergonomic for the caller. Most of the time I expect caller needs to use `.enumerate()`, which is not terribly ugly.

mrain: Oh just like the `from_kv_set` in `UniversalMerkleTreeScheme`? That's actually a good idea.`

mrain: > Another option: `from_elems` could take an `Iterator<Item = (Self::Index, impl Borrow<Self::Element>)` and force the caller to make `Index`s for you. This does not require `From<u64>`. But it is less ergonomic for the caller. Most of the time I expect caller needs to use `.enumerate()`, which is not terribly ugly.

This is actually a good idea. Just like `from_kv_set` from `UniversalMerkleTreeScheme`.

There's a little problem, it's not easy to implement an efficient tree builder like [the one here](https://github.com/EspressoSystems/jellyfish/blob/34059730ee692a35963732b7b9f909c9546dd5c0/primitives/src/merkle_tree/internal.rs#L178) because you can't assume order on the provided tuples which is generic over `Index`.

ggutoski: Can you elaborate? I don't see where an ordering is needed. Seems that the existing implementation does not use ordering either. It merely uses `From<u64>` as a way to make a new `Index`. But you have no ordering guarantee on these indices. (eg. the `From<u64>` impl could be a hash output or string or anything.)

ggutoski: Maybe you need to add a uniqueness check for each `Index` from the callers iterator...?

mrain: Ordering is implicitly needed in that implementation. As it's just repeatedly aggregate `Arity` nodes to form an internal nodes as their parents. 

Also, what if user provide `[(0, sth), (1, sth), (3, sth)]` as the input?

ggutoski: As discussed offline:
- appendable merkle tree really needs `Index` to be an integer type like `usize`, so we should probably enforce `usize`
- it should be possible to maintain `from_elems` in `MerkleTreeScheme` and remove `From<u64>`. Then modify `AppendableMerkleTree` to enforce `Index = usize`.
- We might need to use `Iterator<Item = (Self::Index, impl Borrow<Self::Element>)` as described [above](https://github.com/EspressoSystems/jellyfish/pull/475#discussion_r1473324565) to implement `from_elems` for tree types that allow generic `Index`.

It would be nice to get these changes into the present PR. But if it's too much extra work then we could push it to a later PR.

alxiong: offline discussion during today's crypto meeting:

remove constructors from traits, functions that generic over traits should be receiving a struct and only use traits for behaviors, not for construction. (by Jeb, and agreed by everyone)

mrain: Already did. Check the newest commit

primitives/src/merkle_tree/append_only.rs
```diff
@@ -31,17 +31,32 @@ use typenum::Unsigned;
 impl_merkle_tree_scheme!(MerkleTree);
 impl_forgetable_merkle_tree_scheme!(MerkleTree);
 
-impl<E, H, I, Arity, T> AppendableMerkleTreeScheme for MerkleTree<E, H, I, Arity, T>
+impl<E, H, I, Arity, T> MerkleTree<E, H, I, Arity, T>
 where
     E: Element,
     H: DigestAlgorithm<E, I, T>,
     I: Index + From<u64> + AddAssign + ToTraversalPath<Arity>,
```
ggutoski, line 38: Whoa, TIL `ToTraversalPath` exists. it's all starting to fall into place. 😉 

I thought the plan is to force `Index = usize`. Also, do you not wish to remove `From<u64>`?

ggutoski (dismissed):
Approve to give you the option to end the discussion. One comment below.

primitives/src/merkle_tree/append_only.rs
mrain: Now we force `Index = u64`. [f7d75e4](https://github.com/EspressoSystems/jellyfish/pull/475/commits/f7d75e428bbd635e53f0eaf864ab54b63a35b9b6)

primitives/src/merkle_tree/internal.rs
```diff
@@ -212,7 +210,7 @@ where
             .map(|chunk| {
                 let children = chunk
                     .map(|(pos, elem)| {
-                        let pos = I::from(pos as u64);
+                        let pos = pos as u64;
```
ggutoski, line 213: Nit. `as` could cause bugs. You could throw an error if `usize`->`u64` conversion fails. (It never will because `usize` is at most 8 bytes... for now.) It's convenient that this method already returns a `Result`, so you don't need to obfuscate your API merely to accommodate shortfalls in Rust.

A bug can only occur if you're on a 32-bit system and someone attempts to create a merkle tree with 4 billion elements. So don't lose sleep over this. 😁

primitives/src/merkle_tree/internal.rs
```diff
@@ -291,12 +288,12 @@ where
             .map(|chunk| {
                 let children = chunk
                     .map(|(pos, elem)| {
-                        Ok(if (pos as u64) < num_leaves - 1 {
+                        let pos = pos as u64;
```
ggutoski, line 291: same

ggutoski (approved):
Awesome. Don't even look at my new comments.

primitives/src/merkle_tree/internal.rs
mrain: Nah, the error happens if I'm on a (>64)-bit system and throwing >2^64 elements to create a MT.

ggutoski: lol oops! I still blame Rust. 😊

Issue #393, opened 2023-11-01
ADVZ allow large polynomial degree


ggutoski:
Premise: low polynomial degree is suboptimal. Performance might improve if we allow polynomials with larger degree. Thus, we need to send multiple evaluations (instead of only one evaluation) from each polynomial to each storage node.

chancharles92:
My (unreliable) guess is that we can achieve the best performance when we set poly degree ~1000 and add parallelism (for doing k RS encoding/poly committing in parallel, where k is the number of polys).

ggutoski:
As a bonus, a fix for this issue also fixes #356 . ie. it reduces the size of `Common`.

PR #480, opened 2024-02-05
feat: Enable support for large polynomials in VID

akonring:
## Description

This PR introduces the concept of multiplicity of the degree for polynomials encoding the payload.
Support for large degree polynomials is done by extending message/codeword length by a factor $\alpha$ while maintaining the same rate $r=\frac{\alpha\cdot m}{\alpha\cdot n}$. Each node in the system will receive $\alpha$ evaluation points (multi-shares) for each polynomial $f_i$  $\forall i\in \{1,\ldots,k\}$.

closes: #393

### DIFF
--- a/primitives/src/vid/advz.rs
+++ b/primitives/src/vid/advz.rs
@@ -37,6 +37,7 @@ use ark_std::{
     Zero,
 };
 use bytes_to_field::{bytes_to_field, field_to_bytes};
+use core::mem;
 use derivative::Derivative;
 use digest::crypto_common::Output;
 use itertools::Itertools;
@@ -122,13 +123,13 @@ where
 
         // Later we will convert to u32.
         // Better to know now whether that conversion will succeed.
-        if <usize as TryInto<u32>>::try_into(num_storage_nodes).is_err() {
+        if u32::try_from(num_storage_nodes).is_err() {
             return Err(VidError::Argument(format!(
                 "num_storage nodes {} should be convertible to u32",
                 num_storage_nodes
             )));
         }
-        if <usize as TryInto<u32>>::try_into(multiplicity).is_err() {
+        if u32::try_from(multiplicity).is_err() {
             return Err(VidError::Argument(format!(
                 "multiplicity {} should be convertible to u32",
                 multiplicity
@@ -367,16 +368,14 @@ where
             if (index + 1) % self.multiplicity == 0 {
                 shares.push(Share {
                     index,
-                    evals: evals.clone(),
-                    aggregate_proofs: proofs.clone(),
+                    evals: mem::take(&mut evals),
+                    aggregate_proofs: mem::take(&mut proofs),
                     evals_proof: all_evals_commit // TODO: check MT lookup for each index
                         .lookup(KzgEvalsMerkleTreeIndex::<E, H>::from(index as u64))
                         .expect_ok()
                         .map_err(vid)?
                         .1,
                 });
-                evals = Vec::new();
-                proofs = Vec::new()
             }
         }
 
@@ -511,10 +510,11 @@ where
                 share.evals.len()
             )));
         }
-        if num_evals % self.multiplicity > 0 {
+        if num_evals != self.multiplicity * common.poly_commits.len() {
             return Err(VidError::Argument(format!(
-                "multiplicity does not divide num_evals: multiplicity {}, num_evals {}",
-                self.multiplicity, num_evals,
+                "num_evals should be (multiplicity * poly_commits): {} but is instead: {}",
+                self.multiplicity * common.poly_commits.len(),
+                num_evals,
             )));
         }
         let chunk_size = self.multiplicity * self.payload_chunk_size;
@@ -523,10 +523,8 @@ where
         let elems_capacity = num_evals * chunk_size;
         let mut elems = Vec::with_capacity(elems_capacity);
 
+        let mut evals = Vec::with_capacity(num_evals);
         for p in 0..num_polys {
-            // each share is guaranteed to contain the same number of evals
-            // but we don't know to which polynomial each eval belongs
-            let mut evals = Vec::new();
             for share in shares {
                 // extract all evaluations for polynomial p from the share
                 for m in 0..self.multiplicity {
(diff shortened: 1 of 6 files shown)


primitives/src/vid/advz.rs
```diff
@@ -323,30 +350,36 @@ where
...
+            evals.extend(all_storage_node_evals[index].iter());
+            proofs.push(aggregate_proofs[index].clone());
+            if (index + 1) % self.multiplicity == 0 {
+                shares.push(Share {
                     index,
-                    evals,
-                    aggregate_proof,
-                    evals_proof: all_evals_commit
+                    evals: evals.clone(),
+                    aggregate_proofs: proofs.clone(),
```
ggutoski, line 371: This might be a good use case for [`Vec::drain`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain). If that works then you could delete lines 378-379.

primitives/src/vid/advz.rs
```diff
@@ -445,30 +493,48 @@ where
...
+        let num_polys = num_evals / self.multiplicity;
 
-        let elems_capacity = num_polys * self.payload_chunk_size;
+        let elems_capacity = num_evals * chunk_size;
         let mut elems = Vec::with_capacity(elems_capacity);
-        for i in 0..num_polys {
+
+        for p in 0..num_polys {
+            // each share is guaranteed to contain the same number of evals
+            // but we don't know to which polynomial each eval belongs
```
ggutoski, line 528: As discussed offline, let's strengthen this check to ensure for all shares the evals len equals `common.poly_commits.len()` times `multiplicity`. Then you can know which polynomial belongs to each eval.

primitives/src/vid/advz.rs
```diff
@@ -682,29 +755,29 @@ mod tests {
     };
     use sha2::Sha256;
 
-    // #[test]
-    #[allow(dead_code)]
+    #[ignore]
```
ggutoski, line 758: ❤️

primitives/src/vid/advz.rs
```diff
@@ -866,9 +939,10 @@ mod tests {
...
             // 1 eval missing from all shares
             shares_missing_evals.last_mut().unwrap().evals.pop();
-            let bytes_recovered = advz
-                .recover_payload(&shares_missing_evals, &common)
-                .expect("recover_payload should succeed when shares have equal eval lengths");
+            let bytes_recovered = advz.recover_payload(&shares_missing_evals, &common).expect(
+                "recover_payload should succeed when shares have
+            equal eval lengths",
+            );
             assert_ne!(bytes_recovered, bytes_random);
```
ggutoski, line 946: As discussed, this test is dumb. Please delete it.

ggutoski (commented):
Looks great! Only minor changes remain as discussed offline.

ggutoski:
[CI failure](https://github.com/EspressoSystems/jellyfish/actions/runs/7803221583/job/21284984932?pr=480) seems to be the same as this other PR: https://github.com/EspressoSystems/jellyfish/pull/481#issuecomment-1928740539 The fix is the same: wait and try again.

primitives/src/vid/advz.rs
```diff
@@ -57,6 +57,7 @@ where
 {
     payload_chunk_size: usize,
```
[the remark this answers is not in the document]
akonring: Yes, went back and forth on this but eventually just left it.
Think my original confusion stemmed from the fact that in coding theory a "chunk" can mean different things (fx see [here](https://en.wikipedia.org/wiki/Erasure_code#Applications_of_erasure_coding_in_storage_systems)).

After grasping the meaning I don't think it is that bad and it is just nice to call `.chunks(chunk_size)` on an iterator :smile:.

Let me know if you have strong opinions on this, then we can change it.

primitives/src/vid/advz.rs
akonring: Clippy is suggesting `std::mem::take(&mut evals)`.
But I'm unsure if introducing this has implications for our WASM compliance?
(Arkworks has a similar `ark_std::mem`)

primitives/src/vid/advz.rs
ggutoski: On second thought we should keep the test but check that payload recovery returns an error

primitives/src/vid/advz.rs
ggutoski: No idea. Tempting as it is to learn the answer, let's not get nerd-sniped by this question. I leave it to your judgement. 😁

primitives/src/vid/advz.rs
```diff
@@ -175,7 +197,7 @@ where
     evals: Vec<KzgEval<E>>,
 
     #[serde(with = "canonical")]
-    aggregate_proof: KzgProof<E>,
+    aggregate_proofs: Vec<KzgProof<E>>,
```
ggutoski, line 200:
```suggestion
    // aggretate_proofs.len() equals self.multiplicity
    // TODO further aggregate into a single KZG proof.
    aggregate_proofs: Vec<KzgProof<E>>,
```
...or some reasonable explainer comment.

ggutoski (approved):
Just one more suggested code comment.

ggutoski (approved):
:shipit:

chancharles92 (commented):
@akonring Good job!

chancharles92:
@ggutoski @akonring Do we want to add some tests for the case where `multiplicity > 1`?

akonring:
> @ggutoski @akonring Do we want to add some tests for the case where `multiplicity > 1`?

Yes, good point! Should prolly have been a part of this PR but can create another issue and make a separate PR.

ggutoski:
> @ggutoski @akonring Do we want to add some tests for the case where `multiplicity > 1`?

LOL thank you @chancharles92 ! 🤦

akonring:
Issue here: https://github.com/EspressoSystems/jellyfish/issues/483
Thanks for keeping me honest, @chancharles92.

Since I was a little too fast on the merge button, if there are any leftover input/comments we can maybe put them in the new issue #483

Issue #483, opened 2024-02-07
Introduce new test in advz for large-degree polynomials


akonring:
**Introduce tests in advz testing different choices of `multiplicity`.**

The feature that allows to modify the polynomial degree encoding the payload (through the parameter `multiplicity`) was implemented in #480.
But no unit tests were added as a part of the PR :

PR #485, opened 2024-02-08
fix: incorrect VID verification/recovery (multiplicity > 1)

akonring:
## Description

#480 introduced a new feature that allows for building polynomials of larger degree when encoding the VID payload by choosing a multiplicity parameter but no tests were submitted with the new features.
 
This PR fixes issues in the VID verification and payload recovery when `multiplicity > 1` and **creates a test** to exercise the new part of the code.

closes: #483

### DIFF
--- a/primitives/src/vid/advz.rs
+++ b/primitives/src/vid/advz.rs
@@ -981,6 +981,30 @@ mod tests {
         }
     }
 
+    #[test]
+    fn disperse_and_recover_with_multiplicity() {
+        let multiplicity = 4;
+        let bytes_per_element = bytes_to_field::elem_byte_capacity::<KzgEval<Bls12_381>>();
+        let payload_size = bytes_per_element * 1 << 7; // 128 field elements
+        let (payload_chunk_size, num_storage_nodes) = (4, 6);
+        let mut rng = jf_utils::test_rng();
+        let srs = init_srs(payload_chunk_size * multiplicity, &mut rng);
+        let advz = Advz::new(payload_chunk_size, num_storage_nodes, multiplicity, srs).unwrap();
+        let bytes_random = init_random_payload(payload_size, &mut rng);
+
+        let disperse: VidDisperse<Advz<Bls12_381, Sha256>> = advz.disperse(&bytes_random).unwrap();
+        let (shares_multiplicity, common) = (disperse.shares, disperse.common);
+        let bytes_per_element = bytes_to_field::elem_byte_capacity::<KzgEval<Bls12_381>>();
+        let bytes_per_poly = payload_chunk_size * multiplicity * bytes_per_element;
+
+        assert_eq!(common.poly_commits.len(), payload_size / bytes_per_poly);
+
+        let bytes_recovered = advz
+            .recover_payload(&shares_multiplicity, &common)
+            .expect("recover_payload should succeed with multiplicity");
+        assert_eq!(bytes_recovered, bytes_random);
+    }
+
     /// Routine initialization tasks.
     ///
     /// Returns the following tuple:
--- a/primitives/src/vid/advz.rs
+++ b/primitives/src/vid/advz.rs
@@ -981,30 +981,6 @@ mod tests {
         }
     }
 
-    #[test]
-    fn disperse_and_recover_with_multiplicity() {
-        let multiplicity = 4;
-        let bytes_per_element = bytes_to_field::elem_byte_capacity::<KzgEval<Bls12_381>>();
-        let payload_size = bytes_per_element * 1 << 7; // 128 field elements
-        let (payload_chunk_size, num_storage_nodes) = (4, 6);
-        let mut rng = jf_utils::test_rng();
-        let srs = init_srs(payload_chunk_size * multiplicity, &mut rng);
-        let advz = Advz::new(payload_chunk_size, num_storage_nodes, multiplicity, srs).unwrap();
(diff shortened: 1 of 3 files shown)


primitives/src/vid/advz.rs
```diff
@@ -976,6 +981,30 @@ mod tests {
         }
     }
 
+    #[test]
+    fn disperse_and_recover_with_multiplicity() {
```
ggutoski, line 985: Thanks for this test. It looks like it's an "integration" test. ie. It talks only to public-facing APIs, needs no access to internal details. I suggest we remove this test and incorporate it into the existing integration test `primitives/tests/advz.rs`. You'll need to tweak the `vid_factory` closure to accept a a new `multiplicity` arg. We should add test cases for multiple distinct multiplicities, maybe all of them 1,2,4,8,16.

primitives/src/vid/advz.rs
```diff
@@ -246,15 +246,14 @@ where
...
-        let code_word_size = self.multiplicity * self.num_storage_nodes;
 
         let polys: Vec<_> = bytes_to_field::<_, KzgEval<E>>(payload)
             .chunks(chunk_size)
             .into_iter()
             .map(|evals_iter| self.polynomial(evals_iter))
             .collect();
         let poly_commits = UnivariateKzgPCS::batch_commit(&self.ck, &polys).map_err(vid)?;
-        Self::derive_commit(&poly_commits, payload.len(), code_word_size)
+        Self::derive_commit(&poly_commits, payload.len(), self.num_storage_nodes)
```
ggutoski, line 256: glad you made this switch. We want num_storage_nodes in the commit.

ggutoski (changes requested):

primitives/src/vid/advz.rs
```diff
@@ -976,6 +981,30 @@ mod tests {
...
+        let payload_size = bytes_per_element * 1 << 7; // 128 field elements
+        let (payload_chunk_size, num_storage_nodes) = (4, 6);
+        let mut rng = jf_utils::test_rng();
+        let srs = init_srs(payload_chunk_size * multiplicity, &mut rng);
+        let advz = Advz::new(payload_chunk_size, num_storage_nodes, multiplicity, srs).unwrap();
+        let bytes_random = init_random_payload(payload_size, &mut rng);
+
+        let disperse: VidDisperse<Advz<Bls12_381, Sha256>> = advz.disperse(&bytes_random).unwrap();
+        let (shares_multiplicity, common) = (disperse.shares, disperse.common);
+        let bytes_per_element = bytes_to_field::elem_byte_capacity::<KzgEval<Bls12_381>>();
```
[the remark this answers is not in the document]
akonring: yes, thanks

primitives/src/vid/advz.rs
akonring: Yes, I see your point but running the current tests combined (cartesian-product style) with multiplicity=1,2,4,8,16 takes 6-7min (local machine). 

I opted for a Frankenstein approach that just picks a random multiplicity for each `(payload_chunk_size, num_storage_nodes)`-pair, .

Let me know if you have something else in mind.

ggutoski: good call. thank you

primitives/tests/advz.rs
```diff
@@ -32,7 +31,13 @@ fn round_trip() {
...
 
     vid::round_trip(
-        |payload_chunk_size, num_storage_nodes| {
+        |payload_chunk_size, num_storage_nodes, multiplicity| {
+            let mut srs_rng = jf_utils::test_rng();
+            let srs = UnivariateKzgPCS::<Bls12_381>::gen_srs_for_testing(
+                &mut srs_rng,
+                num_coeff * multiplicity,
+            )
+            .unwrap();
```
ggutoski, line 40: You don't need to move this inside the VID factory. Instead just do it once at the beginning (like before) but set it to `supported_degree * max_multiplicity` where `max_multiplicity` is the maximum possible multiplicity, something like
```rust
let max_multiplicity = multiplicities.iter().max();
```
This might speed up the test a bit.

primitives/tests/vid/mod.rs
```diff
@@ -1,32 +1,35 @@
...
 /// Correctness test generic over anything that impls [`VidScheme`]
 ///
 /// `pub` visibility, but it's not part of this crate's public API
 /// because it's in an integration test.
 /// <https://doc.rust-lang.org/book/ch11-03-test-organization.html#submodules-in-integration-tests>
 pub fn round_trip<V, R>(
-    vid_factory: impl Fn(usize, usize) -> V,
+    vid_factory: impl Fn(usize, usize, usize) -> V,
     vid_sizes: &[(usize, usize)],
+    multiplicities: &[usize],
```
ggutoski, line 17: This vid_factory has lost all pretence of abstraction now that it also takes a multiplicity arg. We could recover full abstraction here but it's probably not worth the trouble so don't worry about it.

ggutoski (approved):
One nontrivial suggestion and one trivial suggestion. Approving anyway; you can decide whether you want to bother to change it or just merge as-is. Thank you!

primitives/tests/vid/mod.rs
akonring: Yes, weaving in the multiplicity sort of ruined the nice abstraction.

ggutoski (approved):

PR #489, opened 2024-02-14
feat: implement non destructive universal Merkle tree

mrain:
## Description

- Introduce `NonDestructiveUniversalMerkleTreeScheme` trait.
- Refactor internal wrapper of `MerkleNode` from `Box` to `Arc`.
- Add persistent update functionalities to `UniversalMerkleTree`.

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/primitives/src/merkle_tree/internal.rs
+++ b/primitives/src/merkle_tree/internal.rs
@@ -364,6 +364,7 @@ where
 {
     /// Forget a leaf from the merkle tree. Internal branch merkle node will
     /// also be forgotten if all its leafs are forgotten.
+    /// WARN(#495): this method breaks non-membership proofs.
     #[allow(clippy::type_complexity)]
     pub(crate) fn forget_internal(
         &self,
@@ -400,10 +401,12 @@ where
                                 .collect::<Vec<_>>(),
                         });
                         children[traversal_path[height - 1]] = new_child;
-                        if children
-                            .iter()
-                            .all(|child| matches!(**child, MerkleNode::ForgettenSubtree { .. }))
-                        {
+                        if children.iter().all(|child| {
+                            matches!(
+                                **child,
+                                MerkleNode::Empty | MerkleNode::ForgettenSubtree { .. }
+                            )
+                        }) {
                             (
                                 Arc::new(MerkleNode::ForgettenSubtree { value: *value }),
                                 LookupResult::Ok(elem, proof),
--- a/primitives/src/merkle_tree/internal.rs
+++ b/primitives/src/merkle_tree/internal.rs
@@ -363,17 +363,25 @@ where
 {
     /// Forget a leaf from the merkle tree. Internal branch merkle node will
     /// also be forgotten if all its leafs are forgotten.
+    #[allow(clippy::type_complexity)]
     pub(crate) fn forget_internal(
-        &mut self,
+        &self,
         height: usize,
         traversal_path: &[usize],
-    ) -> LookupResult<E, MerklePath<E, I, T>, MerklePath<E, I, T>> {
+    ) -> (
+        Arc<Self>,
+        LookupResult<E, MerklePath<E, I, T>, MerklePath<E, I, T>>,
+    ) {
         match self {
-            MerkleNode::Empty => LookupResult::NotFound(vec![MerkleNode::Empty; height + 1]),
+            MerkleNode::Empty => (
+                Arc::new(self.clone()),
+                LookupResult::NotFound(vec![MerkleNode::Empty; height + 1]),
+            ),
             MerkleNode::Branch { value, children } => {
-                match children[traversal_path[height - 1]]
-                    .forget_internal(height - 1, traversal_path)
-                {
+                let mut children = children.clone();
+                let (new_child, result) = children[traversal_path[height - 1]]
+                    .forget_internal(height - 1, traversal_path);
+                match result {
                     LookupResult::Ok(elem, mut proof) => {
                         proof.push(MerkleNode::Branch {
                             value: T::default(),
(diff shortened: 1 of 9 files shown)


primitives/src/merkle_tree/internal.rs
```diff
@@ -655,6 +655,123 @@ where
...
+            },
+            MerkleNode::Branch { value, children } => {
+                let branch = traversal_path[height - 1];
+                let result = children[branch].persistent_update_with_internal::<H, Arity, _>(
+                    height - 1,
+                    pos,
+                    traversal_path,
+                    f,
+                )?;
+                let mut children = children.clone();
```
jbearer, line 701: I think this ends up doing a deep clone of the whole subtree, since each child is a `Box`. Don't we need to make the children `Arc` or something so we can shallow clone the children that aren't being updated?

jbearer (commented):
The name "persistent" is a bit confusing to me. I'm having a hard time articulating why...I think "persistent" update is not a term I'm familiar with, and it's not really the update that's persistent, it's...the old version of the tree I guess? Whereas like "non-destructive update" or "immutable update" would be terms that are more intuitive. But it's a matter of personal opinion mostly.

Main comment is the issue with cloning `children`.

primitives/src/merkle_tree/internal.rs
mrain: Then should we make everything `Arc`?
Also I learnt before that there are certain differences between `std::sync::Arc` and `async_std::sync::Arc`.

jbearer: Oh no way I thought those were just type aliases lol. For this I don't know that it matters too much which one we use (it's only going to be used internally anyways, not part of the public API).

I think we need to make everything `Arc`, but as I was saying in Zulip, this makes it hard to implement a regular mutable update. So I think we have two choices:
* Make everything `Arc`, and implement the regular mutable update by doing a non-mutable update and then, at the very end, setting the root with `*self = new_root`. This will have slightly worse performance but still O(log(n)) I think
* Have a completely separate node type for immutable and mutable updates

Personally I would lean towards the former as it means we have a lot less code to maintain (and mutable updates are a thin wrapper around immutable updates which is kind of nice). But I don't know how much we care about the performance of mutable updates.

Probably nothing we have now is highly sensitive to the performance of this, and because the mutable and immutable updates are in separate traits, we could always implement a specialized version of the tree that only supports mutable updates later on, if it became important

mrain: Persistent data structure is a common name in competitive programming community. Anyway I change it to `NonDestructive`.

I've made everything `Arc`. Check the newest changes.

mrain: Oh CI fails for wasm.

@alxiong Is there a wasm-compatible `Arc`?

jbearer: > Persistent data structure is a common name in competitive programming community. Anyway I change it to NonDestructive.

You learn something new every day! Thanks for introducing me to the terminology, and feel free to name it whatever you want if you like that better

primitives/src/merkle_tree/internal.rs
```diff
@@ -365,61 +363,85 @@ where
...
                                     }
                                 })
                                 .collect::<Vec<_>>(),
                         });
+                        children[traversal_path[height - 1]] = new_child;
                         if children.iter().all(|child| {
                             matches!(
                                 **child,
-                                MerkleNode::Empty | MerkleNode::ForgettenSubtree { value: _ }
+                                MerkleNode::Empty | MerkleNode::ForgettenSubtree { .. }
```
jbearer, line 405: Am I reading this right, if after doing an update all the children are empty _or_ forgotten, we replace the whole branch with a forgotten subtree? Doesn't this break non-membership proofs for the universal merkle tree? Sometimes we care to distinguish empty branches from forgotten subtrees

mrain: Oh you just find another long standing bug.
Fixed here [5a3229c](https://github.com/EspressoSystems/jellyfish/pull/489/commits/5a3229c9c804721fc6fd1c7e23b84ff0b4bd493f)

jbearer (approved):
LGTM

primitives/src/merkle_tree/internal.rs
mrain: Test fails here: https://github.com/EspressoSystems/jellyfish/blob/9daf7655bb734ab6c856986363b7389fb1237989/primitives/src/merkle_tree/universal_merkle_tree.rs#L471-L474

Now we may have to think more carefully about the behavior of `universal_forget`. To preserve non-membership proofs, if you forget any leaf from an empty merkle tree, it's not trimming down the memory usage, rather it creates a path to record the information that an empty leaf is forgotten.

Maybe the best option would be allowing user to forget a subtree, but current API doesn't expose the concept of the subtree so it's hard to do so.

primitives/src/merkle_tree/universal_merkle_tree.rs
```diff
@@ -146,6 +146,50 @@ where
...
+        pos: impl Borrow<Self::Index>,
+        elem: impl Borrow<Self::Element>,
+    ) -> Result<Self, PrimitivesError> {
+        self.persistent_update_with(pos, |_| Some(elem.borrow().clone()))
+    }
+
+    fn persistent_remove(&self, pos: Self::Index) -> Result<Self, PrimitivesError> {
+        self.persistent_update_with(pos, |_| None)
+    }
+
```
alxiong, line 169: these can be default impl?

primitives/src/merkle_tree/internal.rs
jbearer: Yikes. One idea from a long time ago was to have "pinned" paths. Basically an extra boolean on each node, and you are allowed to forget empty sub-trees _unless_ they are pinned. Not entirely sure if that fixes the problem

I think we can save this for future work, and keep the old behavior of forget for now (where it doesn't preserve non-membership proofs). We aren't using forget right now in the sequencer, and with the low memory usage afforded by the persistent updates, I don't know if we'll ever have to. Plus, if any proof is ever missing, we have the ability to fetch it.

mrain: Rolled back in [6946a32](https://github.com/EspressoSystems/jellyfish/pull/489/commits/6946a320d112ac69e289cc1525bedafaa1219564) and added an issue #495

primitives/src/merkle_tree/universal_merkle_tree.rs
mrain: Yes, also for `update` and `remove` [6946a32](https://github.com/EspressoSystems/jellyfish/pull/489/commits/6946a320d112ac69e289cc1525bedafaa1219564)

jbearer (approved):

Issue #490, opened 2024-02-15
Integrate with ICICLE MSM


alxiong:
As our exploration in #470, the benchmark on icicle is impressive on GPU, this issue aims to integrate into Jellyfish for MSM, NTT and potentially EC-NTT.

Here's an attentive plan:

- [x] setup separate nix devShell with cuda and other deps installed
- [x] workflow diagram (@mrain)
- [x] add optional invocation of icicle's MSM in `UnivariateKZG::commit()` behind a feature flag
- [x] add (equivalence) tests on MSM
- [x] add new benchmark for MSM alone
- [ ] add new benchmark for VID dispersal

mrain:
Please check if it's clear enough: https://drive.google.com/file/d/1iPDYrTxI2EjBf_Ua4CoQklOJInTdgCED/view?usp=sharing

cc @alxiong @chancharles92

philippecamacho:
Draft PR https://github.com/EspressoSystems/jellyfish/pull/498

mrain:
Plan: 
- Merge PR #498 by Mar 6
- Integrate with VID by Mar 8.
- @mrain will help on integration with HotShot by Mar 15.

mrain:
There're some troubles for upstream users now. In order to use GPU, you need to specify the icicle equivalent curves and fields. But if we move the icicle type generic to an associate type, it could be much easier.
```rust
pub trait GPUCommit<E, C>
```
to 
```rust
pub trait GPUCommit<E> {
    type ICICLE_EQUIV;
}
```

mrain:
Because of the messy trait bounds, we are expecting a delay for this project.

updated timeline:

- refactor traits by Mar 13.
- VID integration by Mar 15.
- HotShot integration by Mar 20.

PR #318, opened 2023-06-14
chore: add flamegraph instructions for profiling

alxiong:
## Description

Add `cargo-flamegraph` to nix-shell and add instructions to demonstrate how to run in in README.

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [x] Targeted PR against correct branch (main)
- [x] Linked to GitHub issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [x] ~Wrote unit tests~
- [x] Updated relevant documentation in the code
- [x] ~Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md`~
- [ ] Re-reviewed `Files changed` in the GitHub PR explorer

PR #498, opened 2024-02-27
feat: ICICLE msm integration

alxiong:
## Description

closes: #490 

Unit tests already test the correctness of the MSM results.
Benchmark can be run via `cargo bench --bench msm --features "test-srs icicle"`

- [ ] Fixed cudatoolkit inside nix shell is incompatible with cuda driver problem (causing [`CudaErrorInsufficientDriver`](https://github.com/ingonyama-zk/icicle/blob/e6035698b5e54632f2c44e600391352ccc11cad4/wrappers/golang/cuda_runtime/const.go#L63C2-L63C30) err)
- [x] Tweak around benchmark code and setup
- [x] Update Changelog

## Benchmark

**TL;DR: it's about 50x speedup compared to arkworks!** 🎉

Current `criterion` benchmark:
```
MSM with arkworks/19    time:   [1.5497 s 1.5526 s 1.5554 s]
MSM with arkworks/20    time:   [2.8726 s 2.8801 s 2.8848 s]
MSM with arkworks/21    time:   [5.8268 s 5.8572 s 5.8847 s]
MSM with arkworks/22    time:   [11.618 s 11.641 s 11.665 s]

MSM with ICICLE/19      time:   [26.753 ms 26.841 ms 27.018 ms]
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high severe
MSM with ICICLE/20      time:   [48.716 ms 48.728 ms 48.737 ms]
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) low mild
MSM with ICICLE/21      time:   [87.649 ms 87.671 ms 87.697 ms]
MSM with ICICLE/22      time:   [164.33 ms 164.40 ms 164.45 ms]
```

Individual GPU-accelerated breakdown `MSM(2^20)`

```
Start:   Committing to polynomial of degree 1048576
··Start:   Type Conversion: ark->ICICLE: Group
··End:     Type Conversion: ark->ICICLE: Group .....................................11.500ms
··Start:   Load group elements: CPU->GPU
··End:     Load group elements: CPU->GPU ...........................................5.197ms
··Start:   Type Conversion: ark->ICICLE: Scalar
··End:     Type Conversion: ark->ICICLE: Scalar ....................................5.835ms
··Start:   Load scalars: CPU->GPU
··End:     Load scalars: CPU->GPU ..................................................2.624ms
··Start:   GPU-accelerated MSM
··End:     GPU-accelerated MSM .....................................................20.493ms
··Start:   Load MSM result GPU->CPU
··End:     Load MSM result GPU->CPU ................................................26.000ms
··Start:   Type Conversion: ICICLE->ark: Group
··End:     Type Conversion: ICICLE->ark: Group .....................................27.140µs
End:     Committing to polynomial of degree 1048576  ...............................84.879ms
```
_note_: the "GPU-accelerated MSM" is somewhat misleading, because we use non-blocking async MSM on GPU, the computation actually wasn't finished before our CPU moved on, thus part of "Load MSM result GPU->CPU" is "synchronizing the result on the cuda stream" which means wait for the work to finish. Since we know the loading a single projective group should be instant, the more accurate MSM computation time is `20.49 + 26 = 47 sec` which aligns with the `criterion` output above.

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/primitives/benches/kzg_gpu.rs
+++ b/primitives/benches/kzg_gpu.rs
@@ -6,10 +6,12 @@ use ark_bn254::Bn254;
 #[cfg(feature = "icicle")]
 use ark_ec::models::{short_weierstrass::Affine, CurveConfig};
 use ark_ec::pairing::Pairing;
+#[cfg(feature = "icicle")]
+use ark_ff::PrimeField;
 use ark_poly::{univariate::DensePolynomial, DenseUVPolynomial};
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
 #[cfg(feature = "icicle")]
-use jf_primitives::icicle_deps::*;
+use jf_primitives::icicle_deps::{curves::*, *};
 use jf_primitives::pcs::{
     prelude::{PolynomialCommitmentScheme, UnivariateKzgPCS},
     StructuredReferenceString,
@@ -51,23 +53,26 @@ where
     C: IcicleCurve + MSM<C>,
     C::ScalarField: ArkConvertible<ArkEquivalent = E::ScalarField>,
     C::BaseField: ArkConvertible<ArkEquivalent = <C::ArkSWConfig as CurveConfig>::BaseField>,
+    <C::ArkSWConfig as CurveConfig>::BaseField: PrimeField,
     E: Pairing<G1Affine = Affine<<C as IcicleCurve>::ArkSWConfig>>,
+    UnivariateKzgPCS<E>: GPUCommit<E, C>,
 {
     let mut group = c.benchmark_group("MSM with ICICLE");
     let mut rng = test_rng();
+    let stream = warmup_new_stream().unwrap();
 
     let supported_degree = 2usize.pow(MAX_LOG_DEGREE as u32);
     let pp = UnivariateKzgPCS::<E>::gen_srs_for_testing(&mut rng, supported_degree).unwrap();
-    // TODO: (alex) figure out load longer SRS first, and only use part of it later
-    // currently it will error if the `scalars.len() % points.len() != 0`
-    // while we can tap into a slice behind the reference via
-    // `HostOrDeviceSlice[..]` which gives `&[T]` however msm() doesn't accept
-    // &[T], only `&HostOrDeviceSlice`
+    let (full_ck, _vk) = pp.trim(supported_degree).unwrap();
+    let mut srs_on_gpu = <UnivariateKzgPCS<E> as GPUCommit<E, C>>::load_prover_param_to_gpu(
+        full_ck,
+        supported_degree,
+    )
+    .unwrap();
 
     // setup for commit first
     for log_degree in MIN_LOG_DEGREE..MAX_LOG_DEGREE {
         let degree = 2usize.pow(log_degree as u32);
-        let (ck, _vk) = pp.trim(degree).unwrap();
         let p = <DensePolynomial<E::ScalarField> as DenseUVPolynomial<E::ScalarField>>::rand(
             degree, &mut rng,
         );
@@ -76,7 +81,23 @@ where
             BenchmarkId::from_parameter(log_degree),
             &log_degree,
             |b, _log_degree| {
-                b.iter(|| UnivariateKzgPCS::<E>::commit_with_gpu::<C>(&ck, &p).unwrap())
+                b.iter(|| {
+                    // step-by-step commitment on gpu
+                    let poly_on_gpu =
+                        <UnivariateKzgPCS<E> as GPUCommit<E, C>>::load_poly_to_gpu(&p).unwrap();
+                    let msm_result_on_gpu =
+                        <UnivariateKzgPCS<E> as GPUCommit<E, C>>::commit_on_gpu(
+                            &mut srs_on_gpu,
+                            &poly_on_gpu,
+                            &stream,
+                        )
+                        .unwrap();
+                    let _comm = <UnivariateKzgPCS<E> as GPUCommit<E, C>>::load_commitment_to_host(
+                        msm_result_on_gpu,
+                        &stream,
+                    )
+                    .unwrap();
+                })
             },
         );
     }
@@ -86,7 +107,7 @@ where
 fn kzg_gpu_bn254(c: &mut Criterion) {
     kzg_ark::<Bn254>(c);
     #[cfg(feature = "icicle")]
-    kzg_icicle::<Bn254, icicle_bn254::curve::CurveCfg>(c);
+    kzg_icicle::<Bn254, IcicleBn254>(c);
 }
 
 criterion_group! {
--- a/primitives/benches/kzg_gpu.rs
+++ b/primitives/benches/kzg_gpu.rs
@@ -1,26 +1,28 @@
 //! This benchmark meant for MSM speed comparison between arkworks and
 //! GPU-accelerated code We use `UnivariateKzgPCS::commit()` as a proxy for MSM
 //!
-//! Run `cargo bench --bench msm --features "test-srs icicle"`
+//! Run `cargo bench --bench kzg-gpu --features "test-srs icicle"`
 use ark_bn254::Bn254;
-#[cfg(feature = "icicle")]
-use ark_ec::models::{short_weierstrass::Affine, CurveConfig};
-use ark_ec::pairing::Pairing;
+use ark_ec::{
+    models::{short_weierstrass::Affine, CurveConfig},
+    pairing::Pairing,
+};
 use ark_poly::{univariate::DensePolynomial, DenseUVPolynomial};
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
-#[cfg(feature = "icicle")]
-use jf_primitives::icicle_deps::*;
-use jf_primitives::pcs::{
-    prelude::{PolynomialCommitmentScheme, UnivariateKzgPCS},
-    StructuredReferenceString,
+use jf_primitives::{
+    icicle_deps::*,
+    pcs::{
+        prelude::{PolynomialCommitmentScheme, UnivariateKzgPCS},
+        StructuredReferenceString,
+    },
 };
 use jf_utils::test_rng;
 
-const MIN_LOG_DEGREE: usize = 12;
+const MIN_LOG_DEGREE: usize = 19;
 const MAX_LOG_DEGREE: usize = 23;
 
 /// running MSM using arkworks backend
-pub fn bench_msm_ark<E: Pairing>(c: &mut Criterion) {
+pub fn kzg_ark<E: Pairing>(c: &mut Criterion) {
     let mut group = c.benchmark_group("MSM with arkworks");
     let mut rng = test_rng();
 
@@ -46,7 +48,7 @@ pub fn bench_msm_ark<E: Pairing>(c: &mut Criterion) {
 
 /// running MSM using ICICLE backends
 #[cfg(feature = "icicle")]
-pub fn bench_msm_icicle<E, C>(c: &mut Criterion)
+pub fn kzg_icicle<E, C>(c: &mut Criterion)
 where
     C: IcicleCurve + MSM<C>,
     C::ScalarField: ArkConvertible<ArkEquivalent = E::ScalarField>,
@@ -58,6 +60,11 @@ where
 
     let supported_degree = 2usize.pow(MAX_LOG_DEGREE as u32);
     let pp = UnivariateKzgPCS::<E>::gen_srs_for_testing(&mut rng, supported_degree).unwrap();
+    // TODO: (alex) figure out load longer SRS first, and only use part of it later
+    // currently it will error if the `scalars.len() % points.len() != 0`
+    // while we can tap into a slice behind the reference via
+    // `HostOrDeviceSlice[..]` which gives `&[T]` however msm() doesn't accept
+    // &[T], only `&HostOrDeviceSlice`
 
     // setup for commit first
     for log_degree in MIN_LOG_DEGREE..MAX_LOG_DEGREE {
@@ -78,17 +85,16 @@ where
     group.finish();
 }
 
-fn msm_bn254(c: &mut Criterion) {
-    bench_msm_ark::<Bn254>(c);
-    #[cfg(feature = "icicle")]
-    bench_msm_icicle::<Bn254, icicle_bn254::curve::CurveCfg>(c);
+fn kzg_gpu_bn254(c: &mut Criterion) {
+    // kzg_ark::<Bn254>(c);
+    kzg_icicle::<Bn254, icicle_bn254::curve::CurveCfg>(c);
 }
 
 criterion_group! {
-    name = msm_benches;
+    name = kzg_gpu_benches;
     config = Criterion::default().sample_size(10);
     targets =
-        msm_bn254,
+        kzg_gpu_bn254,
 }
 
-criterion_main!(msm_benches);
+criterion_main!(kzg_gpu_benches);
--- a/primitives/benches/kzg_gpu.rs
+++ b/primitives/benches/kzg_gpu.rs
@@ -82,21 +82,12 @@ where
             &log_degree,
             |b, _log_degree| {
                 b.iter(|| {
-                    // step-by-step commitment on gpu
-                    let poly_on_gpu =
-                        <UnivariateKzgPCS<E> as GPUCommit<E, C>>::load_poly_to_gpu(&p).unwrap();
-                    let msm_result_on_gpu =
-                        <UnivariateKzgPCS<E> as GPUCommit<E, C>>::commit_on_gpu(
-                            &mut srs_on_gpu,
-                            &poly_on_gpu,
-                            &stream,
-                        )
-                        .unwrap();
-                    let _comm = <UnivariateKzgPCS<E> as GPUCommit<E, C>>::load_commitment_to_host(
-                        msm_result_on_gpu,
+                    <UnivariateKzgPCS<E> as GPUCommit<E, C>>::gpu_commit_with_loaded_prover_param(
+                        &mut srs_on_gpu,
+                        &p,
                         &stream,
                     )
-                    .unwrap();
+                    .unwrap()
                 })
             },
         );
--- a/primitives/benches/kzg_gpu.rs
+++ b/primitives/benches/kzg_gpu.rs
@@ -3,18 +3,16 @@
 //!
 //! Run `cargo bench --bench kzg-gpu --features "test-srs icicle"`
 use ark_bn254::Bn254;
-use ark_ec::{
-    models::{short_weierstrass::Affine, CurveConfig},
-    pairing::Pairing,
-};
+#[cfg(feature = "icicle")]
+use ark_ec::models::{short_weierstrass::Affine, CurveConfig};
+use ark_ec::pairing::Pairing;
 use ark_poly::{univariate::DensePolynomial, DenseUVPolynomial};
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
-use jf_primitives::{
-    icicle_deps::*,
-    pcs::{
-        prelude::{PolynomialCommitmentScheme, UnivariateKzgPCS},
-        StructuredReferenceString,
-    },
+#[cfg(feature = "icicle")]
+use jf_primitives::icicle_deps::*;
+use jf_primitives::pcs::{
+    prelude::{PolynomialCommitmentScheme, UnivariateKzgPCS},
+    StructuredReferenceString,
 };
 use jf_utils::test_rng;
 
@@ -86,7 +84,8 @@ where
 }
 
 fn kzg_gpu_bn254(c: &mut Criterion) {
-    // kzg_ark::<Bn254>(c);
+    kzg_ark::<Bn254>(c);
+    #[cfg(feature = "icicle")]
     kzg_icicle::<Bn254, icicle_bn254::curve::CurveCfg>(c);
 }
 
--- a/primitives/src/pcs/univariate_kzg/mod.rs
+++ b/primitives/src/pcs/univariate_kzg/mod.rs
@@ -698,10 +698,21 @@ impl<E: Pairing> UnivariateKzgPCS<E> {
         end_timer!(conv_time);
 
         // load them on host first
-        let bases = HostOrDeviceSlice::Host(bases);
-        let scalars = HostOrDeviceSlice::Host(scalars);
+        let mut bases_on_device =
+            HostOrDeviceSlice::<'_, IcicleAffine<C>>::cuda_malloc(bases.len()).unwrap();
+        let mut scalars_on_device =
+            HostOrDeviceSlice::<'_, C::ScalarField>::cuda_malloc(scalars.len()).unwrap();
         let mut msm_result: HostOrDeviceSlice<'_, IcicleProjective<C>> =
             HostOrDeviceSlice::cuda_malloc(1).unwrap();
+
+        #[cfg(feature = "kzg-print-trace")]
+        let load_to_device_time = start_timer!(|| "Load bases and scalars to device");
+
+        bases_on_device.copy_from_host(&bases).unwrap();
+        scalars_on_device.copy_from_host(&scalars).unwrap();
+
+        #[cfg(feature = "kzg-print-trace")]
+        end_timer!(load_to_device_time);
 
         let stream = CudaStream::create().unwrap();
         let mut cfg = MSMConfig::default();
@@ -711,14 +722,20 @@ impl<E: Pairing> UnivariateKzgPCS<E> {
         #[cfg(feature = "kzg-print-trace")]
         let msm_time = start_timer!(|| "GPU-accelerated MSM");
 
-        icicle_core::msm::msm(&scalars, &bases, &cfg, &mut msm_result).unwrap();
+        icicle_core::msm::msm(&scalars_on_device, &bases_on_device, &cfg, &mut msm_result).unwrap();
 
         #[cfg(feature = "kzg-print-trace")]
         end_timer!(msm_time);
 
+        #[cfg(feature = "kzg-print-trace")]
+        let load_back_time = start_timer!(|| "Load results back to host");
+
         let mut msm_host_result = vec![IcicleProjective::<C>::zero(); 1];
         stream.synchronize().unwrap();
         msm_result.copy_to_host(&mut msm_host_result[..]).unwrap();
+
+        #[cfg(feature = "kzg-print-trace")]
+        end_timer!(load_back_time);
 
         let commitment = msm_host_result[0].to_ark().into_affine();
 
--- a/primitives/src/pcs/univariate_kzg/mod.rs
+++ b/primitives/src/pcs/univariate_kzg/mod.rs
@@ -793,10 +793,10 @@ pub(crate) mod icicle {
 
             #[cfg(feature = "kzg-print-trace")]
             let conv_time = start_timer!(|| "Type Conversion: ark->ICICLE: Group");
-            let bases: Vec<IcicleAffine<C>> =
-                parallelizable_slice_iter(&prover_param.powers_of_g[..supported_degree + 1])
-                    .map(|&p| Self::ark_affine_to_icicle(p))
-                    .collect();
+            let bases: Vec<IcicleAffine<C>> = prover_param.powers_of_g[..supported_degree + 1]
+                .par_iter()
+                .map(|&p| Self::ark_affine_to_icicle(p))
+                .collect();
             #[cfg(feature = "kzg-print-trace")]
             end_timer!(conv_time);
 
@@ -819,7 +819,8 @@ pub(crate) mod icicle {
 
             #[cfg(feature = "kzg-print-trace")]
             let conv_time = start_timer!(|| "Type Conversion: ark->ICICLE: Scalar");
-            let scalars: Vec<C::ScalarField> = parallelizable_slice_iter(&poly.coeffs()[..size])
+            let scalars: Vec<C::ScalarField> = poly.coeffs()[..size]
+                .par_iter()
                 .map(|&s| Self::ark_field_to_icicle(s))
                 .collect();
             #[cfg(feature = "kzg-print-trace")]
@@ -857,12 +858,12 @@ pub(crate) mod icicle {
 
             #[cfg(feature = "kzg-print-trace")]
             let conv_time = start_timer!(|| "Type Conversion: ark->ICICLE: Scalar");
-            let scalars: Vec<C::ScalarField> = parallelizable_slice_iter(polys)
-                .map(|poly| {
-                    parallelizable_slice_iter(&poly.coeffs()[..size])
-                        .map(|&s| Self::ark_field_to_icicle(s))
-                })
-                .flatten()
+            let scalars: Vec<C::ScalarField> = polys
+                .iter()
+                .flat_map(|poly| poly.coeffs())
+                .collect::<Vec<_>>()
+                .into_par_iter()
+                .map(|&s| Self::ark_field_to_icicle(s))
                 .collect();
             #[cfg(feature = "kzg-print-trace")]
             end_timer!(conv_time);
@@ -950,7 +951,8 @@ pub(crate) mod icicle {
 
             #[cfg(feature = "kzg-print-trace")]
             let conv_time = start_timer!(|| "Type Conversion: ICICLE->ark: Group");
-            let comms = parallelizable_slice_iter(&msm_host_result)
+            let comms = msm_host_result
+                .par_iter()
                 .map(|p| Commitment(p.to_ark().into_affine()))
                 .collect();
             #[cfg(feature = "kzg-print-trace")]
--- a/primitives/src/pcs/univariate_kzg/mod.rs
+++ b/primitives/src/pcs/univariate_kzg/mod.rs
@@ -1313,16 +1313,12 @@ mod tests {
             let p = <DensePolynomial<E::ScalarField> as DenseUVPolynomial<E::ScalarField>>::rand(
                 degree, rng,
             );
-            let comm = <UnivariateKzgPCS<E> as GPUCommit<E, C>>::gpu_commit(&ck, &p)?;
-
-            // step-by-step commitment on gpu
-            let comm2 =
+            let _comm =
                 <UnivariateKzgPCS<E> as GPUCommit<E, C>>::gpu_commit_with_loaded_prover_param(
                     &mut srs_on_gpu,
                     &p,
                     &stream,
                 )?;
-            assert_eq!(comm, comm2);
 
             let polys: Vec<_> = (0..10)
                 .map(|_| {
@@ -1331,7 +1327,8 @@ mod tests {
                     )
                 })
                 .collect();
-            let _comms = <UnivariateKzgPCS<E> as GPUCommit<E, C>>::gpu_batch_commit(&ck, &polys)?;
+            let _comms = <UnivariateKzgPCS<E> as GPUCommit<E, C>>::gpu_batch_commit_with_loaded_prover_param(
+                &mut srs_on_gpu, &polys, &stream)?;
 
             Ok(())
         }
--- a/primitives/src/pcs/univariate_kzg/mod.rs
+++ b/primitives/src/pcs/univariate_kzg/mod.rs
@@ -696,9 +696,8 @@ impl<E: Pairing> UnivariateKzgPCS<E> {
             )));
         }
 
-        // TODO: use async version? so that we can do the conversion on host meanwhile
         let mut bases_on_device =
-            HostOrDeviceSlice::<'_, IcicleAffine<C>>::cuda_malloc(supported_degree + 1).unwrap();
+            HostOrDeviceSlice::<'_, IcicleAffine<C>>::cuda_malloc(supported_degree + 1)?;
 
         #[cfg(feature = "kzg-print-trace")]
         let conv_time = start_timer!(|| "Type Conversion: ark->ICICLE: Group");
@@ -711,7 +710,7 @@ impl<E: Pairing> UnivariateKzgPCS<E> {
 
         #[cfg(feature = "kzg-print-trace")]
         let load_time = start_timer!(|| "Load group elements: CPU->GPU");
-        bases_on_device.copy_from_host(&bases).unwrap();
+        bases_on_device.copy_from_host(&bases)?;
         #[cfg(feature = "kzg-print-trace")]
         end_timer!(load_time);
 
@@ -730,9 +729,7 @@ impl<E: Pairing> UnivariateKzgPCS<E> {
         E: Pairing<G1Affine = Affine<<C as IcicleCurve>::ArkSWConfig>>,
     {
         let size = poly.degree() + 1;
-        // TODO: use async version?
-        let mut scalars_on_device =
-            HostOrDeviceSlice::<'_, C::ScalarField>::cuda_malloc(size).unwrap();
+        let mut scalars_on_device = HostOrDeviceSlice::<'_, C::ScalarField>::cuda_malloc(size)?;
 
         #[cfg(feature = "kzg-print-trace")]
         let conv_time = start_timer!(|| "Type Conversion: ark->ICICLE: Scalar");
@@ -744,7 +741,7 @@ impl<E: Pairing> UnivariateKzgPCS<E> {
 
         #[cfg(feature = "kzg-print-trace")]
         let load_time = start_timer!(|| "Load scalars: CPU->GPU");
-        scalars_on_device.copy_from_host(&scalars).unwrap();
+        scalars_on_device.copy_from_host(&scalars)?;
         #[cfg(feature = "kzg-print-trace")]
         end_timer!(load_time);
 
@@ -768,9 +765,9 @@ impl<E: Pairing> UnivariateKzgPCS<E> {
         C::BaseField: ArkConvertible<ArkEquivalent = <C::ArkSWConfig as CurveConfig>::BaseField>,
         E: Pairing<G1Affine = Affine<<C as IcicleCurve>::ArkSWConfig>>,
     {
-        let mut msm_result = HostOrDeviceSlice::<'_, IcicleProjective<C>>::cuda_malloc(1).unwrap();
+        let mut msm_result = HostOrDeviceSlice::<'_, IcicleProjective<C>>::cuda_malloc(1)?;
 
-        let stream = CudaStream::create().unwrap();
+        let stream = CudaStream::create()?;
         let mut cfg = MSMConfig::default();
         cfg.ctx.stream = &stream;
         cfg.is_async = true; // non-blocking
@@ -778,7 +775,7 @@ impl<E: Pairing> UnivariateKzgPCS<E> {
         #[cfg(feature = "kzg-print-trace")]
         let msm_time = start_timer!(|| "GPU-accelerated MSM");
 
-        icicle_core::msm::msm(poly_on_gpu, prover_param_on_gpu, &cfg, &mut msm_result).unwrap();
+        icicle_core::msm::msm(poly_on_gpu, prover_param_on_gpu, &cfg, &mut msm_result)?;
 
         #[cfg(feature = "kzg-print-trace")]
         end_timer!(msm_time);
@@ -803,12 +800,10 @@ impl<E: Pairing> UnivariateKzgPCS<E> {
 
         // Since `commit_on_gpu()` is conducting the MSM in async way, we need to
         // synchronize it first.
-        stream.synchronize().unwrap();
+        stream.synchronize()?;
 
         let mut msm_host_result = vec![IcicleProjective::<C>::zero(); 1];
-        commitment_on_gpu
-            .copy_to_host(&mut msm_host_result[..])
-            .unwrap();
+        commitment_on_gpu.copy_to_host(&mut msm_host_result[..])?;
 
         #[cfg(feature = "kzg-print-trace")]
         end_timer!(load_time);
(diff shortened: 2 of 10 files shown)


alxiong:
Current benchmark doesn't look right, the difference is only 5x, we should expect at least >10x speedup. In initial exploration, we observed 200x differences!

```
MSM with arkworks/12    time:   [20.128 ms 20.157 ms 20.189 ms]
MSM with arkworks/13    time:   [35.529 ms 35.653 ms 35.778 ms]
MSM with arkworks/14    time:   [67.961 ms 68.246 ms 68.623 ms]
MSM with arkworks/15    time:   [119.05 ms 119.43 ms 119.77 ms]
MSM with arkworks/16    time:   [216.39 ms 216.61 ms 216.91 ms]
MSM with arkworks/17    time:   [435.28 ms 435.69 ms 436.10 ms]
MSM with arkworks/18    time:   [820.43 ms 821.17 ms 822.05 ms]
MSM with arkworks/19    time:   [1.5854 s 1.5876 s 1.5902 s]
MSM with arkworks/20    time:   [2.9161 s 2.9178 s 2.9199 s]
MSM with arkworks/21    time:   [5.9400 s 5.9642 s 5.9849 s]
MSM with arkworks/22    time:   [11.604 s 11.664 s 11.716 s]

MSM with ICICLE/12      time:   [15.503 ms 15.512 ms 15.523 ms]
MSM with ICICLE/13      time:   [17.806 ms 17.834 ms 17.862 ms]
MSM with ICICLE/14      time:   [22.151 ms 22.192 ms 22.222 ms]
MSM with ICICLE/15      time:   [30.990 ms 31.176 ms 31.352 ms]
MSM with ICICLE/16      time:   [51.316 ms 51.570 ms 51.919 ms]
MSM with ICICLE/17      time:   [89.319 ms 89.740 ms 90.149 ms]
MSM with ICICLE/18      time:   [160.70 ms 161.77 ms 163.22 ms]
MSM with ICICLE/19      time:   [318.50 ms 319.18 ms 319.94 ms]
MSM with ICICLE/20      time:   [635.62 ms 637.99 ms 640.69 ms]
MSM with ICICLE/21      time:   [1.2574 s 1.2616 s 1.2680 s]
MSM with ICICLE/22      time:   [2.5118 s 2.5236 s 2.5354 s]
```

primitives/src/pcs/univariate_kzg/mod.rs
```diff
@@ -652,6 +655,81 @@ where
...
+            .collect();
+        let scalars: Vec<C::ScalarField> = poly.coeffs()[..size]
+            .iter()
+            .map(|&s| C::ScalarField::from_ark(s))
+            .collect();
+        #[cfg(feature = "kzg-print-trace")]
+        end_timer!(conv_time);
+
+        // load them on host first
+        let bases = HostOrDeviceSlice::Host(bases);
```
chancharles92, line 702: Do we need to convert and load SRS bases everytime during committing? Can we just load it once and reuse it?

chancharles92: Another point: in VID, the polynomial degree is much lower. Thus we also won't need to upload too many `powers_of_g` elements to GPUs.

primitives/src/pcs/univariate_kzg/mod.rs
```diff
@@ -652,6 +655,81 @@ where
...
+                poly.degree(),
+                prover_param.powers_of_g.len()
+            )));
+        }
+
+        #[cfg(feature = "kzg-print-trace")]
+        let conv_time = start_timer!(|| "Type Conversion: arkworks->ICICLE");
+
+        let size = poly.degree() + 1;
+        let bases: Vec<IcicleAffine<C>> = prover_param.powers_of_g[..size]
```
mrain, line 690: Did some profilings on my local computer with degree 2^20, this type conversion consumes quite a lot of time.
Consider using `parallelizable_slice_iter` instead.

```
Start:   KZG10::Setup with prover degree 1048576 and verifier degree 1
··Start:   Generating powers of G
··End:     Generating powers of G ..................................................637.759ms
End:     KZG10::Setup with prover degree 1048576 and verifier degree 1 .............695.933ms
Start:   Committing to polynomial of degree 1048576
··Start:   Type Conversion: arkworks->ICICLE
··End:     Type Conversion: arkworks->ICICLE .......................................382.334ms
··Start:   GPU-accelerated MSM
··End:     GPU-accelerated MSM .....................................................38.753ms
End:     Committing to polynomial of degree 1048576  ...............................730.142ms
Start:   Opening polynomial of degree 1048576
··Start:   Computing witness polynomial
··End:     Computing witness polynomial ............................................92.564ms
··Start:   Converting polynomial coeffs to
    bigints
··End:     Converting polynomial coeffs to
    bigints .............................12.859ms
End:     Opening polynomial of degree 1048576 ......................................932.008ms
Start:   Checking evaluation
End:     Checking evaluation Result: true...........................................1.059ms
```

mrain: And the bench result after using parallelizable_slice_iter
```
MSM with ICICLE/19      time:   [42.931 ms 43.461 ms 43.881 ms]
MSM with ICICLE/20      time:   [76.738 ms 77.138 ms 77.910 ms]
MSM with ICICLE/21      time:   [143.75 ms 145.62 ms 148.23 ms]
MSM with ICICLE/22      time:   [272.33 ms 275.70 ms 279.48 ms]
```

mrain: Their type conversion is also not optimal, see
https://github.com/ingonyama-zk/icicle/blob/e6035698b5e54632f2c44e600391352ccc11cad4/wrappers/rust/icicle-core/src/field.rs#L68-L82
and
https://github.com/ingonyama-zk/icicle/blob/e6035698b5e54632f2c44e600391352ccc11cad4/wrappers/rust/icicle-core/src/field.rs#L125-L135

We can at least avoid one memory copy here.

mrain: Moreover, sad that the conversion is not trivial.
Looks like icicle stores the field elements as in plain u64 limbs, while artworks stores them in Montgomery form.
But it's not complicated, definitely we do not need those memory copies.

alxiong: yah those "conversion dominated" breakdown makes lots of sense. thanks for helping on this! 

I'll improve as you suggested!

alxiong: `par_iter` added in [250e6f4](https://github.com/EspressoSystems/jellyfish/pull/498/commits/250e6f450db94ab2b232b6d10b6abe678dca83f7), 

>Their type conversion is also not optimal

i assume you are saying that they don't need to convert to bytes then back, instead, they can just use bigint?
if so, it's the upstream that needs to be changed. plus the memory copy should be neglibile and fast

primitives/src/pcs/univariate_kzg/mod.rs
chancharles92: I see. I feel it might be better to split it at the PCS level rather than in the VID code. Because the PCS commit function itself shouldn't upload SRS everytime. Is it possible that we add another API, `load_srs_to_gpu()`, and the `commit_in_gpu` function can take `&HostOrDevicesSlice<T>` and no longer need to upload srs, and it will return an error if SRS hasn't been uploaded?

In the VID code, it will call `load_srs_to_gpu` and `commit_in_gpu` when needed.

primitives/src/pcs/univariate_kzg/mod.rs
mrain: > ^^ somehow it feels like my par_iter is not working, the conversion part still takes a long time
(I suspect it's because my CPU core is limited only 4 cores, 8 threads 🤔

Yes, it heavily depends on number of cores.

> i assume you are saying that they don't need to convert to bytes then back, instead, they can just use bigint?
if so, it's the upstream that needs to be changed. plus the memory copy should be neglibile and fast

Sure we could nail a PR in upstream repos. But I think we could have our own converter.

primitives/src/pcs/univariate_kzg/mod.rs
alxiong: your suggestion makes more sense! I'll implement that!

alxiong: a slight annoyance is lifetime due to `HostOrDeviceSlice<'a, T>` as a return parameter, I've been fighting this today. I don't want to assign `'static` for it, as we shouldn't expect the reference to live that long, we only want it to be as long as the cuda pointer being active.

I'll figure sth out, but just to share some engineering journey.

primitives/src/pcs/univariate_kzg/mod.rs
```diff
@@ -652,6 +655,164 @@ where
...
+        C::BaseField: ArkConvertible<ArkEquivalent = <C::ArkSWConfig as CurveConfig>::BaseField>,
+        E: Pairing<G1Affine = Affine<<C as IcicleCurve>::ArkSWConfig>>,
+    {
+        let size = poly.degree() + 1;
+        let mut scalars_on_device = HostOrDeviceSlice::<'_, C::ScalarField>::cuda_malloc(size)?;
+
+        #[cfg(feature = "kzg-print-trace")]
+        let conv_time = start_timer!(|| "Type Conversion: ark->ICICLE: Scalar");
+        let scalars: Vec<C::ScalarField> = parallelizable_slice_iter(&poly.coeffs()[..size])
+            .map(|&s| C::ScalarField::from_ark(s))
```
mrain, line 737: use `from_bytes_le(s.into_bigint().to_bytes_le()` instead.

mrain: If we hardcode these field types, maybe we should use `from_limbs` directly.

mrain: If we do not use generics, we may directly use, e.g. `icicle_bn254::curve::ScalarField::from(ark_bn254::Fr::from(1u64).0.0)`

primitives/src/pcs/univariate_kzg/mod.rs
alxiong: done in [514d479](https://github.com/EspressoSystems/jellyfish/pull/498/commits/514d479ea4a17314e09e057ef6e888bb8828cf6a)

primitives/src/pcs/univariate_kzg/mod.rs
mrain: Even faster conversion if we avoid using generics: https://github.com/EspressoSystems/jellyfish/blob/cl/type-conversion/utilities/src/conversion.rs#L310-L339

mrain: [ultimate solution](https://github.com/EspressoSystems/jellyfish/blob/9898c73931ca7d5819da99aadb9dcb49381c95ec/utilities/src/conversion.rs#L311-L365)

alxiong:
I'm so annoyed that I can't get my nix shell to work. outside it, i can compile and run icicle code, but inside I can only successfully compile, but any invocation of cuda FFI code would failed. 
```
$ nix develop .#cudaShell
$ cargo test gpu --release --features icicle
```

<details>
<summary>runtime error trace</summary>

```
---- pcs::univariate_kzg::tests::gpu_end_to_end_test stdout ----
thread 'pcs::univariate_kzg::tests::gpu_end_to_end_test' panicked at primitives/src/pcs/univariate_kzg/mod.rs:1026:14:
test failed for bn254: IcicleError("IcicleError { icicle_error_code: InternalCudaError, cuda_error: Some(cudaErrorInsufficientDriver), reason: Some(\"Runtime CUDA error.\") }")
stack backtrace:
   0:     0x555555b12366 - std::backtrace_rs::backtrace::libunwind::trace::hbee8a7973eeb6c93
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/../../backtrace/src/backtrace/libunwind.rs:104:5
   1:     0x555555b12366 - std::backtrace_rs::backtrace::trace_unsynchronized::hc8ac75eea3aa6899
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x555555b12366 - std::sys_common::backtrace::_print_fmt::hc7f3e3b5298b1083
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/sys_common/backtrace.rs:68:5
   3:     0x555555b12366 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hbb235daedd7c6190
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/sys_common/backtrace.rs:44:22
   4:     0x555555b3b720 - core::fmt::rt::Argument::fmt::h76c38a80d925a410
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/fmt/rt.rs:142:9
   5:     0x555555b3b720 - core::fmt::write::h3ed6aeaa977c8e45
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/fmt/mod.rs:1120:17
   6:     0x555555b0fbff - std::io::Write::write_fmt::h1299aa7741865f2b
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/io/mod.rs:1810:15
   7:     0x555555b12144 - std::sys_common::backtrace::_print::h5d645a07e0fcfdbb
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/sys_common/backtrace.rs:47:5
   8:     0x555555b12144 - std::sys_common::backtrace::print::h85035a511aafe7a8
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/sys_common/backtrace.rs:34:9
   9:     0x555555b13eb7 - std::panicking::default_hook::{{closure}}::hcce8cea212785a25
  10:     0x555555b13b9d - std::panicking::default_hook::hf5fcb0f213fe709a
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/panicking.rs:289:9
  11:     0x555555a8d207 - <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call::h60265c2dfa87ee34
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/alloc/src/boxed.rs:2029:9
  12:     0x555555a8d207 - test::test_main::{{closure}}::h77865bd3127078c6
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/test/src/lib.rs:138:21
  13:     0x555555b144d6 - <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call::hbc5ccf4eb663e1e5
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/alloc/src/boxed.rs:2029:9
  14:     0x555555b144d6 - std::panicking::rust_panic_with_hook::h095fccf1dc9379ee
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/panicking.rs:783:13
  15:     0x555555b14222 - std::panicking::begin_panic_handler::{{closure}}::h032ba12139b353db
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/panicking.rs:657:13
  16:     0x555555b12866 - std::sys_common::backtrace::__rust_end_short_backtrace::h9259bc2ff8fd0f76
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/sys_common/backtrace.rs:171:18
  17:     0x555555b13f80 - rust_begin_unwind
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/panicking.rs:645:5
  18:     0x5555555ace85 - core::panicking::panic_fmt::h784f20a50eaab275
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/panicking.rs:72:14
  19:     0x5555555ad393 - core::result::unwrap_failed::h03d8a5018196e1cd
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/result.rs:1649:5
  20:     0x555555793ade - jf_primitives::pcs::univariate_kzg::tests::gpu_end_to_end_test::h4a2c05e21a017c3d
  21:     0x555555915b69 - core::ops::function::FnOnce::call_once::h34aa7a3b7c3c87aa
  22:     0x555555a92adf - core::ops::function::FnOnce::call_once::h8dc6907944022cf6
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/ops/function.rs:250:5
  23:     0x555555a92adf - test::__rust_begin_short_backtrace::haae1a87433f1efb3
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/test/src/lib.rs:627:18
  24:     0x555555a91861 - test::run_test_in_process::{{closure}}::h8c7decfa7c14e152
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/test/src/lib.rs:650:60
  25:     0x555555a91861 - <core::panic::unwind_safe::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once::h19e6ff056d9d21e9
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/panic/unwind_safe.rs:272:9
  26:     0x555555a91861 - std::panicking::try::do_call::h89c848fcaa37c035
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/panicking.rs:552:40
  27:     0x555555a91861 - std::panicking::try::h57ab3dc74e2839b8
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/panicking.rs:516:19
  28:     0x555555a91861 - std::panic::catch_unwind::hfb6a1b1abc120fb9
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/panic.rs:142:14
  29:     0x555555a91861 - test::run_test_in_process::h5ae2f9875edd562d
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/test/src/lib.rs:650:27
  30:     0x555555a91861 - test::run_test::{{closure}}::h35d7300d8928a067
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/test/src/lib.rs:573:43
  31:     0x555555a58b96 - test::run_test::{{closure}}::h7525ced405d23d1b
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/test/src/lib.rs:601:41
  32:     0x555555a58b96 - std::sys_common::backtrace::__rust_begin_short_backtrace::h4e7db78ce05afad8
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/sys_common/backtrace.rs:155:18
  33:     0x555555a5dbf7 - std::thread::Builder::spawn_unchecked_::{{closure}}::{{closure}}::hcfbcb64f1a1b3482
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/thread/mod.rs:529:17
  34:     0x555555a5dbf7 - <core::panic::unwind_safe::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once::h9d89c5c4108bd689
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/panic/unwind_safe.rs:272:9
  35:     0x555555a5dbf7 - std::panicking::try::do_call::h8a4869bc94ec50c9
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/panicking.rs:552:40
  36:     0x555555a5dbf7 - std::panicking::try::h9a576f20ff81ac30
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/panicking.rs:516:19
  37:     0x555555a5dbf7 - std::panic::catch_unwind::hbcb4e3f860ef9830
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/panic.rs:142:14
  38:     0x555555a5dbf7 - std::thread::Builder::spawn_unchecked_::{{closure}}::h93c79a6be1505948
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/thread/mod.rs:528:30
  39:     0x555555a5dbf7 - core::ops::function::FnOnce::call_once{{vtable.shim}}::h426d96740c81bdaf
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/ops/function.rs:250:5
  40:     0x555555b19095 - <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once::h12de4fc57affb195
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/alloc/src/boxed.rs:2015:9
  41:     0x555555b19095 - <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once::h3c619f45059d5cf1
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/alloc/src/boxed.rs:2015:9
  42:     0x555555b19095 - std::sys::unix::thread::Thread::new::thread_start::hbac657605e4b7389
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/sys/unix/thread.rs:108:17
  43:     0x7ffff76a2333 - start_thread
  44:     0x7ffff7724efc - __clone3
  45:                0x0 - <unknown>
```
</details>

I first suspected that [`clang-sys`](https://github.com/KyleMayes/clang-sys) that icicle's `build.rs` relies on didn't get the correct env var, but even with explicit setting it to the proper clang and llvm bin and lib, I still run into the above error. 

I might need to dig deeper into what actually happens during the FFI calling phase, which object files are being used, and which dynamic library were used to built these executables etc. to pinpoint the source. 

p.s. since icicle's build script[ explicitly use `/usr/local/cuda/lib64` ](https://github.com/ingonyama-zk/icicle/blob/main/wrappers/rust/icicle-cuda-runtime/build.rs#L25), I'm also setting my `CUDA_PATH` etc to a local installation instead of installing a cudatoolkit inside a nix shell

```nix
        baseShell = with pkgs;
          clang15Stdenv.mkDerivation {
            name = "clang15-nix-shell";
            buildInputs = [
              argbash
              openssl
              pkg-config
              git
              nixpkgs-fmt

              cargo-with-nightly
              stableToolchain
              nightlyToolchain
              cargo-sort
              clang-tools_15
              clangStdenv
              llvm_15
            ] ++ lib.optionals stdenv.isDarwin
              [ darwin.apple_sdk.frameworks.Security ];

            CARGO_TARGET_DIR = "target/nix_rustc";

            shellHook = ''
              export RUST_BACKTRACE=full
              export PATH="$PATH:$(pwd)/target/debug:$(pwd)/target/release"
              # Prevent cargo aliases from using programs in `~/.cargo` to avoid conflicts with local rustup installations.
              export CARGO_HOME=$HOME/.cargo-nix

              # Ensure `cargo fmt` uses `rustfmt` from nightly.
              export RUSTFMT="${nightlyToolchain}/bin/rustfmt"

              export C_INCLUDE_PATH="${llvmPackages_15.libclang.lib}/lib/clang/${llvmPackages_15.libclang.version}/include"
              export CC="${clang-tools_15.clang}/bin/clang"
              export CXX="${clang-tools_15.clang}/bin/clang++"
              export AR="${llvm_15}/bin/llvm-ar"
              export CFLAGS="-mcpu=generic"

              # ensure clang-sys got the correct version
              export LLVM_CONFIG_PATH="${llvmPackages_15.llvm.dev}/bin/llvm-config"
              export LIBCLANG_PATH=${llvmPackages_15.libclang.lib}/lib
              export CLANG_PATH=${clang-tools_15.clang}/bin/clang

              # by default choose u64_backend
              export RUSTFLAGS='--cfg curve25519_dalek_backend="u64"'
            ''
              # install pre-commit hooks
              + self.check.${system}.pre-commit-check.shellHook;
          };

//....
        devShells = {
          # enter with `nix develop .#cudaShell`
          cudaShell = baseShell.overrideAttrs (oldAttrs: {
            # for GPU/CUDA env (e.g. to run ICICLE code)
            name = "cuda-env-shell";
            buildInputs = oldAttrs.buildInputs ++ [ cmake util-linux gcc11 ];
            # CXX is overridden to use gcc11 as icicle-curves's build scripts need them, but gcc12 is not supported
            shellHook = oldAttrs.shellHook + ''
              export CUDA_PATH=/usr/local/cuda
              export PATH="${pkgs.gcc11}/bin:$CUDA_PATH/bin:$CUDA_PATH/nvvm/bin:$PATH"
              export LD_LIBRARY_PATH="$CUDA_PATH/lib64:$LIBCLANG_PATH"
            '';
          });
        };
```

🤦 need more time on this

alxiong:
The code is ready for review, interestingly I have some naive "warmup" function that makes the benchmark more accurate, as the warmup takes a constant of ~200ms, which you won't see now:

```
Start:   Committing to polynomial of degree 6238
··Start:   Type Conversion: ark->ICICLE: Group
··End:     Type Conversion: ark->ICICLE: Group .....................................230.554µs
··Start:   Load group elements: CPU->GPU
··End:     Load group elements: CPU->GPU ...........................................78.152µs
··Start:   Type Conversion: ark->ICICLE: Scalar
··End:     Type Conversion: ark->ICICLE: Scalar ....................................54.771µs
··Start:   Load scalars: CPU->GPU
··End:     Load scalars: CPU->GPU ..................................................21.100µs
··Start:   GPU-accelerated MSM
··End:     GPU-accelerated MSM .....................................................25.347ms
··Start:   Load MSM result GPU->CPU
··End:     Load MSM result GPU->CPU ................................................10.292ms
··Start:   Type Conversion: ICICLE->ark: Group
··End:     Type Conversion: ICICLE->ark: Group .....................................13.050µs
End:     Committing to polynomial of degree 6238  ..................................37.062ms
```

There are some remaining tasks:

- [ ] Test passing for subslice (right now construction is fine, but memory dropping still panic; note without subslice things are fine)
- [ ] add `batch_commit()` API which depends on the subslice (because I don't just want exact multiple number of bases of scalars, but I want multiple poly of any degrees <= supported_degree to be committed in batch)
- [ ] update criterion benchmark code

primitives/src/pcs/univariate_kzg/mod.rs
```diff
@@ -652,6 +651,249 @@ where
...
+        fn load_poly_to_gpu<'poly>(
+            poly: &DensePolynomial<E::ScalarField>,
+        ) -> Result<HostOrDeviceSlice<'poly, C::ScalarField>, PCSError> {
+            let size = poly.degree() + 1;
+            let mut scalars_on_device = HostOrDeviceSlice::<'_, C::ScalarField>::cuda_malloc(size)?;
+
+            #[cfg(feature = "kzg-print-trace")]
+            let conv_time = start_timer!(|| "Type Conversion: ark->ICICLE: Scalar");
+            let scalars: Vec<C::ScalarField> = parallelizable_slice_iter(&poly.coeffs()[..size])
+                .map(|&s| Self::ark_field_to_icicle(s)) // TODO: update here
```
mrain, line 761: Same here.

primitives/src/pcs/univariate_kzg/mod.rs
```diff
@@ -910,4 +1152,77 @@ mod tests {
...
+            let pp = UnivariateKzgPCS::<E>::gen_srs_for_testing(rng, supported_degree)?;
+            let (full_ck, _vk) = pp.trim(supported_degree)?;
+            let mut srs_on_gpu =
+                <UnivariateKzgPCS<E> as GPUCommit<E, C>>::load_prover_param_to_gpu(
+                    full_ck,
+                    supported_degree,
+                )?;
+
+            for _ in 0..10 {
+                let degree = usize::rand(rng) % 10000;
```
mrain, line 1188: I recommend running tests for large degrees, e.g. 2^20.

primitives/src/pcs/univariate_kzg/mod.rs
```diff
@@ -910,4 +1152,77 @@ mod tests {
...
+                        <UnivariateKzgPCS<E> as GPUCommit<E, C>>::commit_on_gpu(
+                            &mut srs_on_gpu,
+                            &poly_on_gpu,
+                            &stream,
+                        )?;
+                    let comm2 = <UnivariateKzgPCS<E> as GPUCommit<E, C>>::load_commitment_to_host(
+                        msm_result_on_gpu,
+                        &stream,
+                    )?;
+                    assert_eq!(comm, comm2);
```
mrain, line 1209: Should we add a single test to compare the GPU results with the CPU results? This one could be a small instance

primitives/src/pcs/univariate_kzg/mod.rs
```diff
@@ -652,6 +651,249 @@ where
...
+                )));
+            }
+
+            let mut bases_on_device =
+                HostOrDeviceSlice::<'_, IcicleAffine<C>>::cuda_malloc(supported_degree + 1)?;
+
+            #[cfg(feature = "kzg-print-trace")]
+            let conv_time = start_timer!(|| "Type Conversion: ark->ICICLE: Group");
+            let bases: Vec<IcicleAffine<C>> = parallelizable_slice_iter(&prover_param.powers_of_g[..supported_degree + 1])
+                    .map(|&p| Self::ark_affine_to_icicle(p)) // TODO: maybe add api?
```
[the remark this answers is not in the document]
alxiong: Yes, is that a problem? It's so fast now, I don't see it as a concern. And we don't have unsafe or implicit "forced type interpretation" happening here.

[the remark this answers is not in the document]
mrain: It could be a (small) problem when we are committing large amounts of data.
E.g, for 32MB payload data would be encoded into 2^{17~18} field elements.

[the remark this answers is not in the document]
mrain: Also if we are to use it for plonk proof generation. Often you need to deal with an even larger degree.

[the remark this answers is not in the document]
mrain: BTW it's not urgent, you can leave it as an issue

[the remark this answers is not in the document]
alxiong: so you would suggest: 
load the arkwork type byte-by-byte to GPU, then let GPU directly interpret that on the GPU?

my 2 cents is the `msm()` rust bindings will have rust's type system checks anyway (meaning `msm` is expecting `IcicleAffine<IcicleCurve>`, we might as well just raw pointer cast it in CPU. 
(unless we just completely go around ICICLE's `msm` api and write our own logic using their `msm` via FFI.

primitives/src/pcs/univariate_kzg/mod.rs
alxiong: i guess we could, but the fact that I'm passing this result to verify in `open()` and `verify()` is a testament that it is correct.

alxiong: done in [ca5b494](https://github.com/EspressoSystems/jellyfish/pull/498/commits/ca5b4942c0253dcda433abb2958f52e633cbae6e)

primitives/src/pcs/univariate_kzg/mod.rs
alxiong: done in [ca5b494](https://github.com/EspressoSystems/jellyfish/pull/498/commits/ca5b4942c0253dcda433abb2958f52e633cbae6e)

primitives/src/pcs/univariate_kzg/mod.rs
```diff
@@ -910,4 +1158,91 @@ mod tests {
...
+            let supported_degree = 2usize.pow(21);
+            let pp = UnivariateKzgPCS::<E>::gen_srs_for_testing(rng, supported_degree)?;
+            let (full_ck, _vk) = pp.trim(supported_degree)?;
+            let mut srs_on_gpu =
+                <UnivariateKzgPCS<E> as GPUCommit<E, C>>::load_prover_param_to_gpu(
+                    full_ck,
+                    supported_degree,
+                )?;
+
+            // testing on large degree for profiling
```
mrain, line 1193: I'm thinking about adding a profiling feature. But no action needed right now

primitives/src/pcs/univariate_kzg/mod.rs
```diff
@@ -652,6 +651,255 @@ where
...
+    pub trait GPUCommit<E, C>
+    where
+        C: IcicleCurve + MSM<C>,
+        C::ScalarField: ArkConvertible<ArkEquivalent = E::ScalarField>,
+        C::BaseField: ArkConvertible<ArkEquivalent = <C::ArkSWConfig as CurveConfig>::BaseField>,
+        <C::ArkSWConfig as CurveConfig>::BaseField: PrimeField,
+        E: Pairing<G1Affine = Affine<<C as IcicleCurve>::ArkSWConfig>>,
+    {
+        /// The full cycle of computing poly-commit on GPU
+        fn commit_with_gpu(
```
mrain, line 671: Should we add another function, which assumes that the srs is already on GPU and takes polys as input.

primitives/src/pcs/univariate_kzg/mod.rs
alxiong: so you want something beyond arkworks' `print-trace`? 

i was on pause on this, maybe you can take a look if you find it interesting: https://github.com/EspressoSystems/jellyfish/pull/318

primitives/src/pcs/univariate_kzg/mod.rs
alxiong: this function accepts `HostOrDevice`, so srs and poly can be any one of the 4 combination states. this function shouldn't fail.

if either srs or poly are not on device, the `msm()` from ICICLE side would push it to GPU first, else it will just use the device ref we pass in.

so I'd say we don't need this another function imo

primitives/src/pcs/univariate_kzg/mod.rs
mrain: Oh `print-trace` looks good enough. I mean running this part only when `print-trace` is enabled.

alxiong: i see, okay I can see if I can add this request in my next commit. stay tuned

primitives/src/pcs/univariate_kzg/mod.rs
mrain: I mean a full cycle in VID would be, a pre-loaded SRS is already on GPU, and some interface should takes (a list of) polys and return a commitment

mrain: Current APIs are already sufficient.

I'm overthinking that, should we have a PCS struct, whose `ProverParam` is a `HostOrDeviceSlice`. We don't have to do it now.

alxiong: i don't think that's a good idea, the main concern is icicle is undergoing major breaking changes https://github.com/ingonyama-zk/icicle/pull/412, and I would expect significant changes, so fewer places we use them for now, the easier later when we upgrade.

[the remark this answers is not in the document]
alxiong: what would you suggest? I agree they are very close, I was hoping the API interface (function signature) and code doc could help distinguish their usage. 

but I'm open for better name suggestions.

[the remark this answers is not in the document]
chancharles92: How about `gpu_commit()` and `gpu_batch_commit()`?

primitives/src/pcs/univariate_kzg/mod.rs
```diff
@@ -652,6 +651,344 @@ where
...
+
+        /// Similar to [`Self::commit_with_gpu()`] but for a batch of poly of
+        /// the same degree
+        fn batch_commit_with_gpu(
+            prover_param: impl Borrow<UnivariateProverParam<E>>,
+            polys: &[DensePolynomial<E::ScalarField>],
+        ) -> Result<Vec<Commitment<E>>, PCSError> {
+            let stream = warmup_new_stream().unwrap();
+
+            let degree = polys[0].degree();
```
chancharles92, line 700: Maybe add a check that all polys are having the same degree?

primitives/src/pcs/univariate_kzg/mod.rs
```diff
@@ -652,6 +651,344 @@ where
...
+    /// Trait for GPU-accelerated PCS.commit APIs
+    pub trait GPUCommit<E, C>
+    where
+        C: IcicleCurve + MSM<C>,
+        C::ScalarField: ArkConvertible<ArkEquivalent = E::ScalarField>,
+        C::BaseField: ArkConvertible<ArkEquivalent = <C::ArkSWConfig as CurveConfig>::BaseField>,
+        <C::ArkSWConfig as CurveConfig>::BaseField: PrimeField,
+        E: Pairing<G1Affine = Affine<<C as IcicleCurve>::ArkSWConfig>>,
+    {
+        /// The full cycle of computing poly-commit on GPU
```
chancharles92, line 670: Should we also add another API that includes all steps except for loading prover params? This gives a convenient API if the user wants to commit multiple polynomials but doesn't want to design the computing flow herself.

Another option is to remove the loading_prover_params step in this API and warn the user to call `load_prover_params` before calling this API.

Either option above works for me. (I slightly prefer adding another API)

primitives/src/pcs/univariate_kzg/mod.rs
```diff
@@ -910,4 +1247,122 @@ mod tests {
...
+            <C::ArkSWConfig as CurveConfig>::BaseField: PrimeField,
+            E: Pairing<G1Affine = Affine<<C as IcicleCurve>::ArkSWConfig>>,
+            UnivariateKzgPCS<E>: GPUCommit<E, C>,
+        {
+            let rng = &mut test_rng();
+            let supported_degree = 2usize.pow(12);
+            let pp = UnivariateKzgPCS::<E>::gen_srs_for_testing(rng, supported_degree)?;
+
+            // testing on smaller degree for correctness
+            for _ in 0..10 {
```
mrain, line 1327: weird thing. My local test is failing if I change it to 0..50

primitives/src/pcs/univariate_kzg/mod.rs
```diff
@@ -910,4 +1247,122 @@ mod tests {
...
+            }
+            Ok(())
+        }
+
+        #[test]
+        fn test_gpu_e2e() {
+            test_gpu_e2e_template::<Bn254, IcicleBn254>().unwrap();
+            // testing on large degree for profiling
+            #[cfg(feature = "kzg-print-trace")]
+            gpu_profiling::<Bn254, IcicleBn254>().unwrap();
```
mrain, line 1365: Should we move it to another test fn

primitives/src/pcs/univariate_kzg/mod.rs
```diff
@@ -652,6 +651,344 @@ where
...
+                ));
+            }
+
+            let mut scalars_on_device =
+                HostOrDeviceSlice::<'_, C::ScalarField>::cuda_malloc(size * polys.len())?;
+
+            #[cfg(feature = "kzg-print-trace")]
+            let conv_time = start_timer!(|| "Type Conversion: ark->ICICLE: Scalar");
+            let scalars: Vec<C::ScalarField> = parallelizable_slice_iter(polys)
+                .map(|poly| {
```
mrain, line 828: nit: we could directly use `flat_map` here.

Also according to profiling, this is taking a long time committing 10 degree 2^20 polys.
We should have an issue for faster type conversion

mrain: I suspect that this step takes some unnecessary mem copies

mrain: Don't know why but the following code has 2x speed up
```rust
           #[cfg(feature = "kzg-print-trace")]
            let conv_time = start_timer!(|| "Type Conversion 1: ark->ICICLE: Scalar");
            let escalars: Vec<E::ScalarField> = polys
                .iter()
                .flat_map(|poly| poly.coeffs())
                .copied()
                .collect();
            let _scalars: Vec<C::ScalarField> = parallelizable_slice_iter(&escalars)
                .map(|&s| Self::ark_field_to_icicle(s))
                .collect();
            #[cfg(feature = "kzg-print-trace")]
            end_timer!(conv_time);
```

primitives/src/pcs/univariate_kzg/mod.rs
alxiong: ah, this check was already in the `load_batch_poly_to_gpu()`, do we want duplicated check here as well? (even tho it should be blazing fast, negl cost) 
I'm fine either way

primitives/src/pcs/univariate_kzg/mod.rs
alxiong: definitely, adding another api is better, we want users not to figure out how to use icicle themselves.

@mrain also suggested the same additional handy API offline. I'll add it.

primitives/src/pcs/univariate_kzg/mod.rs
alxiong: ofc, 2^20 * 50 * 32, your GPU memory exhausted

primitives/src/pcs/univariate_kzg/mod.rs
alxiong: sounds good, will do

primitives/src/pcs/univariate_kzg/mod.rs
mrain: I thought they'll be freed. NVM

alxiong: 🤔 you have a good point, lemme investigate why it's not dropping.

primitives/benches/kzg_gpu.rs
```diff
@@ -0,0 +1,120 @@
+//! This benchmark meant for MSM speed comparison between arkworks and
```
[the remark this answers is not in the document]
alxiong: my bad, fixing it now. forgot to update.

primitives/src/pcs/univariate_kzg/mod.rs
alxiong: check added in [ae6a9ca](https://github.com/EspressoSystems/jellyfish/pull/498/commits/ae6a9cadc4fa54aef719af0cce53e5a8b075bbfb)

primitives/src/pcs/univariate_kzg/mod.rs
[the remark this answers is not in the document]
alxiong: renamed in [ae6a9ca](https://github.com/EspressoSystems/jellyfish/pull/498/commits/ae6a9cadc4fa54aef719af0cce53e5a8b075bbfb)

primitives/src/pcs/univariate_kzg/mod.rs
alxiong: moved to new test in [ae6a9ca](https://github.com/EspressoSystems/jellyfish/pull/498/commits/ae6a9cadc4fa54aef719af0cce53e5a8b075bbfb)

primitives/src/pcs/univariate_kzg/mod.rs
alxiong: here's what I observe so far by watching GPU memory usage:

- at startup, before I load anything onto it, as long as i create a cuda stream, `260 MB` of memory will be used
- after SRS is loaded, another 66 MB (which is expected for 2^20) will be used
- after a single commit, at the end of it, there's another 2MB dangling extra usage (I expect everything should be cleaned except the SRS). apparently scalars are freed, cuz they will be 32 MB if not, the MSM result is just ~100B, and I have no clue so far why there's extra 2MB increase. Either somewhere in my code or ICICLE MSM code, we fail to clean up.

primitives/src/pcs/univariate_kzg/mod.rs
alxiong: confirm that this is a problem, scalar conversion time scales more than linearly with the batch_size. there's definitely some unnecessary memory copying.

alxiong: implemented your `flat_map` idea in [7e5862c](https://github.com/EspressoSystems/jellyfish/pull/498/commits/7e5862c2d161bacca594216404465f94cae04a8f)! 
conversion time for scalars drop from 250ms to 108ms
(and MSM took 282ms, so roughly 1/3 of that, and roughly 1/4 of overall runtime)

the single instance takes 5.8 ms. so we are still suffer a super-linear cost (but much better now!)

alxiong: more experiment: a single MSM of 2^23, and batch MSM 8 * 2^20
turns out the scalar conversion of latter is exactly 2x of the former, all the rest is the same. 

this should give the definitive reason: the latter have exactly 1 extra memory copy which cause the 2x runtime. and this copy happens during the two `.collect()`. 

unfortunately, rayon's `ParallelIterator` can only be created from collections or slices, not directly from `Iterator` from std. 

IMO, this experiment demonstrates that we have squeezed the performance as much as we could without violating Rust's type checking. and we should accept this and close #507 now. 
@mrain

mrain: Check this: https://docs.rs/rayon/latest/rayon/iter/trait.ParallelBridge.html

There is still room for improvement. I can investigate it later.

primitives/benches/kzg_gpu.rs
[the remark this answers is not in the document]
mrain: I'm also thinking about a bench for gpu_batch_commit. 
But we can do it along in VID integration.

primitives/src/pcs/univariate_kzg/mod.rs
mrain: This is just # of iterations not batch_size right? And the degree below is nowhere near 2^20.

chancharles92 (approved):
LGTM

PR #516, opened 2024-03-14
chore: overhead free type conversion for `load_poly_to_gpu`

mrain:
## Description

_Update: this also fixes the API changes when `gpu_vid` is enabled._

This PR avoids a memory copy in `load_poly_to_gpu`. However it contains unsafe rust code.

Before:
```
Start:   KZG10::Setup with prover degree 4194304 and verifier degree 1
...
Start:   Type Conversion: ark->ICICLE: Scalar
End:     Type Conversion: ark->ICICLE: Scalar ......................................17.305ms
...
```

After:

```
Start:   KZG10::Setup with prover degree 4194304 and verifier degree 1
...
Start:   Type Conversion: ark->ICICLE: Scalar
End:     Type Conversion: ark->ICICLE: Scalar ......................................160ns
...
```

It doesn't apply to affine point conversion because of two different underlying struct reprs.

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/primitives/src/pcs/univariate_kzg/mod.rs
+++ b/primitives/src/pcs/univariate_kzg/mod.rs
@@ -810,16 +810,18 @@ pub(crate) mod icicle {
 
             #[cfg(feature = "kzg-print-trace")]
             let conv_time = start_timer!(|| "Type Conversion: ark->ICICLE: Scalar");
-            let scalars: Vec<<Self::IC as IcicleCurve>::ScalarField> = poly.coeffs()[..size]
-                .par_iter()
-                .map(|&s| Self::ark_field_to_icicle(s))
-                .collect();
+            // We assume that two types use the same underline repr.
+            let scalars = unsafe {
+                poly.coeffs()[..size]
+                    .align_to::<<Self::IC as IcicleCurve>::ScalarField>()
+                    .1
+            };
             #[cfg(feature = "kzg-print-trace")]
             end_timer!(conv_time);
 
             #[cfg(feature = "kzg-print-trace")]
             let load_time = start_timer!(|| "Load scalars: CPU->GPU");
-            scalars_on_device.copy_from_host(&scalars)?;
+            scalars_on_device.copy_from_host(scalars)?;
             #[cfg(feature = "kzg-print-trace")]
             end_timer!(load_time);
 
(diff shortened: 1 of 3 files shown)


alxiong:
This is pretty impressive speed-up, but you pushed the instance to be 2^22 to make make the gap meaningful, for much smaller instances (which is our case), the gap is so small, (correct?)

- I have no objection of using this unsafe, since it's very limited scope, and easy to reason about. As long as we add some tests to constantly watching for the "same underlying repr" assumption is correct, I'm comfortable. (ICICLE is using tones of unsafe anyway.)
- I vote we turn this into a draft PR and archive it (instead of merging it). and use it in the future when the instance size become so large that conversion time is a bottleneck. 


wdyt? @chancharles92 @mrain

primitives/src/pcs/univariate_kzg/mod.rs
```diff
@@ -1401,6 +1407,26 @@ mod tests {
             test_gpu_e2e_template::<Bn254>().unwrap();
         }
 
+        fn test_gpu_ark_conversion_template<E: Pairing>()
```
ggutoski, line 1410: I like this test. But can we do more? Earlier code comment says:
> We assume that two types use the same underline repr.

In that case, we should be able to assert that the bytes are exactly the same. That way, if an incompatibility is introduced then the test will detect it immediately, no?

mrain: This is a bit annoying but doable: [c82e230](https://github.com/EspressoSystems/jellyfish/pull/516/commits/c82e230c7294f8218df2da73304b8bfe6976261c)

ggutoski: Thank you! I don't fully understand this new test but here's what I see:
1. make a slice of random new ark scalars in `scalars`
2. convert them into a IC scalars slice `ic_scalars` via `from_ark`.
3. copy the slice into a new GPU slice called `d_scalars`.
4. transform the GPU slice to Montgomery form on the GPU
5. copy the GPU scalars out of GPU and back into `ic_scalars`
6. these `ic_scalars` (in Montgomery form) should be byte-for-byte identical to the original `scalars`. Check this via `align_to`

Not sure why you need to copy to/from GPU but AFIACT this test does what it's supposed to.

ggutoski (approved):

PR #528, opened 2024-03-21
chore: cleanup dependencies

mrain:
## Description

- cleanup and re-organize dependencies
- use const generics instead of `typenum` 

Remaining problem:
- still cannot pass no_std check here: `./scripts/check_no_std.sh`

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,19 +13,23 @@ documentation = "https://jellyfish.docs.espressosys.com"
 repository = "https://github.com/EspressoSystems/jellyfish"
 
 [workspace.dependencies]
+ark-bls12-377 = { version = "0.4.0", default-features = false, features = [ "curve" ] }
+ark-bls12-381 = { version = "0.4.0", default-features = false, features = [ "curve" ] }
+ark-bn254 = { version = "0.4.0", default-features = false, features = [ "curve" ] }
+ark-bw6-761 = { version = "0.4.0", default-features = false }
 ark-ec = { version = "0.4.0", default-features = false }
-ark-ff = { version = "0.4.0", default-features = false, features = ["asm"] }
+ark-ff = { version = "0.4.0", default-features = false, features = [ "asm" ] }
 ark-poly = { version = "0.4.0", default-features = false }
 ark-serialize = { version = "0.4.0", default-features = false }
 ark-std = { version = "0.4.0", default-features = false }
 derivative = { version = "2", features = ["use_core"] }
-digest = { version = "0.10.1", default-features = false }
+digest = { version = "0.10", default-features = false }
 displaydoc = { version = "0.2.3", default-features = false }
 hashbrown = "0.14.3"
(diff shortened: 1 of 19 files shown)


Cargo.toml
```diff
@@ -13,5 +13,24 @@ documentation = "https://jellyfish.docs.espressosys.com"
...
+displaydoc = { version = "0.2.3", default-features = false }
+hashbrown = "0.14.3"
+merlin = { version = "3.0.0", default-features = false }
+num-bigint = { version = "0.4", default-features = false }
+rand_chacha = { version = "0.3.1", default-features = false }
+serde = { version = "1.0", default-features = false, features = [ "derive", "rc" ] }
+sha2 = { version = "0.10", default-features = false }
+sha3 = { version = "0.10", default-features = false }
+itertools = { version = "0.12", default-features = false }
+tagged-base64 = { version = "0.4" }
```
alxiong, line 36: we have to be a little conscious about bringing these depenedices up to workspace level since they will be built and compiled even if i only need a specific workspace member crate that doesn't depend on them (thus making the compilation slower)

specifically, I think the curve crates could be pushed down?
(maybe debatable, the only reason our generic code still depend on these curves is because we are implementing specialized [rescue parameters](https://github.com/EspressoSystems/jellyfish/tree/main/primitives/src/rescue/rescue_constants) for them, which IMO should be hidden behind feature flags)

maybe for now, this is not so painful and we could afford importing them all the time. but just keep this caveat in mind if we were to add even more curves support in the future.

primitives/src/merkle_tree/internal.rs
```diff
@@ -137,29 +135,24 @@ impl<T: NodeValue> MerkleCommitment<T> for MerkleTreeCommitment<T> {
 #[derivative(Eq, Hash, PartialEq)]
 #[serde(bound = "E: CanonicalSerialize + CanonicalDeserialize,
              I: CanonicalSerialize + CanonicalDeserialize,")]
-pub struct MerkleProof<E, I, T, Arity>
+pub struct MerkleProof<E, I, T, const ARITY: usize>
```
alxiong, line 138: i believe the [naming convention](https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case) for types are UpperCamelCase, even with constant generic

mrain: lint or fmt was complaining. Looks like most people are using uppercase, e.g. https://www.reddit.com/r/rust/comments/11medef/rust_const_generics_for_proof_types/

alxiong: I see, you will get the following linter warning instead: 
```
  |
1 | trait Foo<E, const Arity: usize> 
  |                          ^^^^^ help: convert the identifier to upper case: `ARITY`
  |
  = note: `#[warn(non_upper_case_globals)]` on by default
```

alxiong (approved):
LGTM

PR #540, opened 2024-04-01
fix: optimize solidity transcript

philippecamacho:
## Description

Related to https://github.com/EspressoSystems/espresso-sequencer/issues/1006.
Simplify the solidity transcript computation in order to save gas in the light client contract.

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/plonk/src/transcript/solidity.rs
+++ b/plonk/src/transcript/solidity.rs
@@ -60,22 +60,17 @@ impl<F> PlonkTranscript<F> for SolidityTranscript {
     where
         E: Pairing,
     {
-        // 1. state = keccak256(state|transcript|0) || keccak256(state|transcript|1)
-        let input0 = [self.state.as_ref(), self.transcript.as_ref(), &[0u8]].concat();
-        let input1 = [self.state.as_ref(), self.transcript.as_ref(), &[1u8]].concat();
+        // 1. state = keccak256(state|transcript))
+        let input = [self.state.as_ref(), self.transcript.as_ref()].concat();
 
         let mut hasher = Keccak256::new();
-        hasher.update(&input0);
+        hasher.update(&input);
         let buf0 = hasher.finalize();
 
-        let mut hasher = Keccak256::new();
-        hasher.update(&input1);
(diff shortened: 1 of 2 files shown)


plonk/src/transcript/solidity.rs
```diff
@@ -60,22 +60,17 @@ impl<F> PlonkTranscript<F> for SolidityTranscript {
...
-        let buf0 = hasher.finalize();
+        hasher.update(&input);
+        let buf = hasher.finalize();
 
-        let mut hasher = Keccak256::new();
-        hasher.update(&input1);
-        let buf1 = hasher.finalize();
-
-        self.state.copy_from_slice(&[buf0, buf1].concat());
+        self.state.copy_from_slice(&[buf].concat());
```
ggutoski, line 70: I don't think `concat` is doing anything here. Why not just
```rust
self.state = hasher.finalize();
```

plonk/src/transcript/solidity.rs
```diff
@@ -60,22 +60,17 @@ impl<F> PlonkTranscript<F> for SolidityTranscript {
...
 
         // 2. challenge: sample field from random bytes.
-        let challenge = E::ScalarField::from_le_bytes_mod_order(&self.state[..48]);
+        let challenge = E::ScalarField::from_be_bytes_mod_order(&self.state);
```
ggutoski, line 73: Is there a reason you switched from `le` to `be`?

plonk/src/transcript/solidity.rs
philippecamacho: Thanks! Fixed in 66391ebe. 
Cannot due directly `self.state = hasher.finalize();` due to the types involved.

plonk/src/transcript/solidity.rs
philippecamacho: Yes, in solidity uint256 are represented using big endian. This will avoid reverting the order of the bits and thus save gas.

ggutoski (approved):

PR #595, opened 2024-05-29
chore: bump icicle to v2.3.1

mrain:
no issue related.

### This PR: 

- bump icicle deps to v2.3.1
  - `HostOrDeviceSlice` is now a trait. Occurrences are changed to concrete types now.
  - now `load_poly_to_gpu` has one memory copy instead of free as before
- add GPU stress test for memory leakage/fragmentation

### Key places to review:

 

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/pcs/src/lib.rs
+++ b/pcs/src/lib.rs
@@ -419,7 +419,10 @@ pub mod icicle_deps {
         curve::{Affine as IcicleAffine, Curve as IcicleCurve, Projective as IcicleProjective},
         msm::{MSMConfig, MSM},
     };
-    pub use icicle_cuda_runtime::{memory::HostOrDeviceSlice, stream::CudaStream};
+    pub use icicle_cuda_runtime::{
+        memory::{DeviceSlice, DeviceVec, HostOrDeviceSlice, HostSlice},
+        stream::CudaStream,
+    };
 
     /// curve-specific types both from arkworks and from ICICLE
     /// including Pairing, CurveCfg, Fr, Fq etc.
@@ -435,8 +438,8 @@ pub mod icicle_deps {
     /// Create a new stream and warmup
     pub fn warmup_new_stream() -> anyhow::Result<CudaStream> {
         let stream = CudaStream::create().map_err(|e| anyhow!("{:?}", e))?;
-        let _warmup_bytes = HostOrDeviceSlice::<'_, u8>::cuda_malloc_async(1024, &stream)
-            .map_err(|e| anyhow!("{:?}", e))?;
+        let _warmup_bytes =
+            DeviceVec::<u8>::cuda_malloc_async(1024, &stream).map_err(|e| anyhow!("{:?}", e))?;
         Ok(stream)
     }
 }
--- a/pcs/src/univariate_kzg/mod.rs
+++ b/pcs/src/univariate_kzg/mod.rs
@@ -1305,8 +1305,8 @@ mod tests {
         #[cfg(feature = "kzg-print-trace")]
         use crate::icicle_deps::warmup_new_stream;
         use crate::{
-            icicle_deps::{curves::*, IcicleCurve},
-            pcs::univariate_kzg::icicle::GPUCommittable,
+            icicle_deps::{curves::*, warmup_new_stream, IcicleCurve},
+            univariate_kzg::icicle::GPUCommittable,
         };
         use core::mem::size_of;
         use icicle_core::traits::{ArkConvertible, MontgomeryConvertible};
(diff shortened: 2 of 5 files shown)


alxiong:
thanks @mrain for taking the initiative to upgrade our icicle dep! 👍

vid/Cargo.toml
```diff
@@ -38,7 +38,7 @@ ark-ed-on-bls12-377 = "0.4.0"
 ark-ed-on-bls12-381 = "0.4.0"
 ark-ed-on-bn254 = "0.4.0"
 criterion = "0.5.1"
-jf-pcs = { git = "https://github.com/EspressoSystems/jellyfish", tag = "0.4.5", features = ["test-srs"] }
+jf-pcs = { path = "../pcs", features = ["test-srs"] }
```
mrain, line 41: Use path b/c some updates are not reflected in the github yet.

ggutoski: Do you intend to switch back before merge?

pcs/src/lib.rs
```diff
@@ -435,8 +438,8 @@ pub mod icicle_deps {
     /// Create a new stream and warmup
     pub fn warmup_new_stream() -> anyhow::Result<CudaStream> {
         let stream = CudaStream::create().map_err(|e| anyhow!("{:?}", e))?;
-        let _warmup_bytes = HostOrDeviceSlice::<'_, u8>::cuda_malloc_async(1024, &stream)
-            .map_err(|e| anyhow!("{:?}", e))?;
+        let _warmup_bytes =
+            DeviceVec::<u8>::cuda_malloc_async(1024, &stream).map_err(|e| anyhow!("{:?}", e))?;
```
alxiong, line 442: i think we can forgo this function in favor of out-of-box: https://github.com/ingonyama-zk/icicle/blob/621676bd41659bc101f29dc7ac54f29c7cdaa92b/wrappers/rust/icicle-cuda-runtime/src/device.rs#L44

vid/Cargo.toml
alxiong: i don't think he can. One of the perils of separate versioning is multi-step upgrade here. 
changing back need another PR: a new tag will be added once this PR is merged, then next one would switch back.

vid/src/advz/test.rs
```diff
@@ -48,6 +48,28 @@ fn commit_only_timer() {
     let _ = advz.commit_only(payload_random);
 }
 
+#[ignore]
+#[test]
+#[cfg(feature = "gpu-vid")]
+/// Stress test with varied payload sizes for GPU memory
+/// leakage/fragmentation.
```
alxiong, line 55: how does this test leakage or fragmentation?

iiuc, this only put lots of running instances, and in case we didn't clean up (after use) the memory, it would overflow the memory?

but how does it test fragmentation?

pcs/src/univariate_kzg/mod.rs
```diff
@@ -965,67 +964,56 @@ pub(crate) mod icicle {
...
-            icicle_bn254::curve::ScalarField::from(f.0 .0)
+            unsafe { ark_std::mem::transmute(f.into_bigint()) }
         }
 
-        // NOTE: we are directly using montgomery form, different from default!
         fn ark_affine_to_icicle(p: ark_bn254::G1Affine) -> icicle_bn254::curve::G1Affine {
             icicle_bn254::curve::G1Affine {
-                x: icicle_bn254::curve::BaseField::from(p.x.0 .0),
-                y: icicle_bn254::curve::BaseField::from(p.y.0 .0),
+                x: unsafe { ark_std::mem::transmute(p.x.into_bigint()) },
```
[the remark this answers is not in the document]
alxiong: and did we add test for this unsafe transmute anywhere?

also can you add comment that even though arkworks uses u64 limb (as opposed to u32 in icicle), they are both little-endian encoded, thus agree on a byte-level, which is why this unsafe transmute is correct. 

^^ am I right?

vid/src/advz.rs
```diff
@@ -63,7 +63,7 @@ pub type AdvzGPU<'srs, E, H> = AdvzInternal<
     E,
     H,
     (
-        HostOrDeviceSlice<'srs, IcicleAffine<<UnivariateKzgPCS<E> as GPUCommittable<E>>::IC>>,
+        DeviceVec<IcicleAffine<<UnivariateKzgPCS<E> as GPUCommittable<E>>::IC>>,
         CudaStream,
```
alxiong, line 67: do we strictly need `type AdvzGPU`? the `'srs` lifetime can be removed at least?

pcs/src/univariate_kzg/mod.rs
```diff
@@ -886,40 +889,36 @@ pub(crate) mod icicle {
...
-        /// - if you pass in `HostOrDeviceSlice::Host`, then `icicle_core::msm`
-        ///   will push them onto GPU first; if they are already on GPU, i.e.
-        ///   `HostOrDeviceSlice::Device`, then msm will be executed directly
         /// - the result is also temporarily on GPU, you can use
         ///   `Self::load_commitments_to_host()` to load back to host CPU.
         /// - this function is async/non-blocking, thus returns a CudaStream
         ///   handle
         /// - default implementation assume normal(non-montgomery) affine for
         ///   bases and scalars, consider overwrite this function if you want
         ///   otherwise
```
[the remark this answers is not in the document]
mrain: No this is always correct. The generic implementation assumes that the scalars are normal. Only for `Bn254` we take montgomery scalars. Now they have to be normal in all cases.

vid/src/advz/test.rs
mrain: Suggested by Salman. It really depends on the GPU memory management. Hypothesis is that if you make many GPU memory allocations and they are not cleaned up fast enough, then there're potentially fragmentations that prevent you from allocating even if the GPU memory is not full.

pcs/src/univariate_kzg/mod.rs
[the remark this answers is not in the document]
mrain: Yes, the test is here: https://github.com/EspressoSystems/jellyfish/pull/595/files#diff-1b7c209306eb56d628e64ac6bc03978824af57767468dc3c618243860d04a6e3L1415

vid/src/advz.rs
mrain: Sure we could remove the lifetime specifier. I'm not sure what do you mean, `AdvzGPU` takes additional type generics for on GPU srs and CudaStream.

vid/src/advz/test.rs
ggutoski: plz put this explanation in a comment 🙏

vid/src/advz.rs
alxiong: > AdvzGPU takes additional type generics for on GPU srs and CudaStream.

i forgot about this. nvm. maybe just remove the lifetime.

pcs/src/lib.rs
alxiong: this is probably the last unaddressed comment? @mrain

mrain: They are already at v4.0.0 and there're many breaking changes. Probably we could directly jump there by spending some more time if necessary.

Issue #605, opened 2024-06-11
[bug]: incorrect lagrange evaluation when zeta is roots of unity


alxiong:
Thanks to Shresth and CommonPrefix team for catching this bug (for an audit of the corresponding plonk verifier contract).

Currently the logic for computing lagrange evaluation is wrong when random challenge `zeta` is roots of unity (which results in `vanishing_eval` to be zero. In this case, $\mathcal{L}_1(\zeta) = 1$ if $\zeta$ is the first roots-of-unity, and zero otherwise. but our current code would simply panic due to division by zero. 

```rust
    pub(crate) fn evaluate_lagrange_1_and_n(
        &self,
        zeta: &E::ScalarField,
        vanish_eval: &E::ScalarField,
    ) -> (E::ScalarField, E::ScalarField) {
        let divisor =
            E::ScalarField::from(self.domain.size() as u32) * (*zeta - E::ScalarField::one());
        let lagrange_1_eval = *vanish_eval / divisor;
        let divisor =
            E::ScalarField::from(self.domain.size() as u32) * (*zeta - self.domain.group_gen_inv);
        let lagrange_n_eval = *vanish_eval * self.domain.group_gen_inv / divisor;
        (lagrange_1_eval, lagrange_n_eval)
    }
```

this also affects `evaluate_pi_poly()` that internally uses `lagrange_i_eval`. 
We should further double check and fix prover-side code.

Note: we didn't catch this in test because:
- we didn't have property test on lagrange poly eval
- `zeta` is random challenge, the probability of it being one of roots of unity is negligible

Thus this issue should also aim to improve tests to guard against this bug

Note2: this is a hard-to-exploit bug (unless our FS is also broken), so current downstream relying on jellyfish shouldn't panic, but the PR that fix this bug should issue a new release and recommend all downstream to migrate over soon.

cc @mrain @chancharles92

alxiong:
comments needed, 
I wanted to directly replace our usage with arkworks' API, unfortunately, they only have [`EvalDomain::evaluate_all_lagrange_coefficients()`](https://docs.rs/ark-poly/latest/ark_poly/domain/trait.EvaluationDomain.html#method.evaluate_all_lagrange_coefficients)

maybe I'll just mimic write a `fn evaluate_lagrange_coeff_at(domain)` for now, and submit a PR upstream for future arkworks to include this API directly in `struct EvalDomain`, okay?

PR #611, opened 2024-06-21
fix: correct lagrange coefficient computation

alxiong:
closes: #605 

### This PR: 

Executed this plan: https://github.com/EspressoSystems/jellyfish/issues/605#issuecomment-2180773751
and updated all previously lagrange coeffs computation

### This PR does not:

Update anything about recursive circuit, since the occurrence of the challenge `zeta` being one of the domain elements is negligibly small, not sure if we want to burden our recursion circuit to include the extra branch condition.  

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/plonk/src/lagrange.rs
+++ b/plonk/src/lagrange.rs
@@ -1,0 +1,224 @@
+//! Utilities for Lagrange interpolations, evaluations, coefficients for a
+//! polynomial
+
+use ark_ff::FftField;
+use ark_poly::{EvaluationDomain, Radix2EvaluationDomain};
+use ark_std::{ops::Range, vec, vec::Vec};
+
+// TODO: (alex) include these APIs upstream in arkworks directly
+
+/// A helper trait for computing Lagrange coefficients of an evaluation domain
+///
+/// from arkworks:
+/// Evaluate all Lagrange polynomials at tau to get the lagrange coefficients.
+/// Define the following as
+/// - H: The coset we are in, with generator g and offset h
+/// - n: The size of the coset H
+/// - Z_H: The vanishing polynomial for H. Z_H(x) = prod_{i in n} (x - hg^i) =
+///   x^n - h^n
+/// - v_i: A sequence of values, where v_0 = 1/(n * h^(n-1)), and v_{i + 1} = g
+///   * v_i
+///
+/// We then compute L_{i,H}(tau) as `L_{i,H}(tau) = Z_H(tau) * v_i / (tau - h *
+/// g^i)`
+#[allow(dead_code)]
+pub(crate) trait LagrangeCoeffs<F: FftField> {
+    /// Returns the first coefficient: `L_{0, Domain}(tau)`
+    fn first_lagrange_coeff(&self, tau: F) -> F;
+    /// Returns the last coefficient: `L_{n-1, Domain}(tau)`
+    fn last_lagrange_coeff(&self, tau: F) -> F;
+    /// Return a list of coefficients for `L_{range, Domain}(tau)`
+    fn lagrange_coeffs_for_range(&self, range: Range<usize>, tau: F) -> Vec<F>;
+}
+
+impl<F: FftField> LagrangeCoeffs<F> for Radix2EvaluationDomain<F> {
+    // L_0(tau) = Z_H(tau) * g^0 / (n * h^(n-1) *(tau - h * g^0))
+    // with g^0 = 1
+    // special care when tau in H, as both numerator and denominator is zero
+    fn first_lagrange_coeff(&self, tau: F) -> F {
+        let offset = self.coset_offset();
+        let z_h_at_tau = self.evaluate_vanishing_polynomial(tau);
+        if z_h_at_tau.is_zero() {
+            if tau == F::one() * offset {
+                F::one()
+            } else {
+                F::zero()
+            }
+        } else {
+            let denominator = self.size_as_field_element()
+                * offset.pow([self.size() as u64 - 1])
+                * (tau - offset);
+            z_h_at_tau * denominator.inverse().unwrap()
+        }
+    }
+
+    // L_n-1(tau) = Z_H(tau) * g^-1 / (n * h^(n-1) * (tau - h * g^-1))
+    // with g^n-1 = g^-1
+    fn last_lagrange_coeff(&self, tau: F) -> F {
+        let offset = self.coset_offset();
+        let z_h_at_tau = self.evaluate_vanishing_polynomial(tau);
+        if z_h_at_tau.is_zero() {
+            if tau == self.group_gen_inv() * offset {
+                F::one()
+            } else {
+                F::zero()
+            }
+        } else {
+            let denominator = self.size_as_field_element()
+                * offset.pow([self.size() as u64 - 1])
+                * (tau - offset * self.group_gen_inv());
+            z_h_at_tau * self.group_gen_inv() * denominator.inverse().unwrap()
+        }
+    }
+
+    // similar to `EvaluationDomain::evaluate_all_lagrange_coefficients()`
+    //
+    // # Panic
+    // if `range` exceeds the `self.size()`
+    fn lagrange_coeffs_for_range(&self, range: Range<usize>, tau: F) -> Vec<F> {
+        if range.end > self.size() {
+            panic!("Out of range: domain size smaller than range.end");
+        }
+        let size = range.end - range.start;
+        let z_h_at_tau = self.evaluate_vanishing_polynomial(tau);
+        let offset = self.coset_offset();
+        let group_gen = self.group_gen();
+        let group_start = group_gen.pow([range.start as u64]);
+
+        if z_h_at_tau.is_zero() {
+            // In this case, we know that tau = hg^i, for some value i.
+            // Then i-th lagrange coefficient in this case is then simply 1,
+            // and all other lagrange coefficients are 0.
+            // Thus we find i by brute force.
+            let mut u = vec![F::zero(); size];
+            let mut omega_i = offset * group_start;
+            for u_i in u.iter_mut().take(size) {
+                if omega_i == tau {
+                    *u_i = F::one();
+                    break;
+                }
+                omega_i *= &group_gen;
+            }
+            u
+        } else {
+            // In this case we have to compute `Z_H(tau) * v_i / (tau - h g^i)`
+            // for i in start..end
+            // We actually compute this by computing (Z_H(tau) * v_i)^{-1} * (tau - h g^i)
+            // and then batch inverting to get the correct lagrange coefficients.
+            // We let `l_i = (Z_H(tau) * v_i)^-1` and `r_i = tau - h g^i`
+            // Notice that since Z_H(tau) is i-independent,
+            // and v_i = g * v_{i-1}, it follows that
+            // l_i = g^-1 * l_{i-1}
+
+            let group_gen_inv = self.group_gen_inv();
+            let start = range.start as u64;
+
+            // v_0_inv = n * h^(n-1)
+            let v_0_inv = self.size_as_field_element() * offset.pow([self.size() as u64 - 1]);
+            let mut l_i = z_h_at_tau.inverse().unwrap() * v_0_inv * group_gen_inv.pow([start]);
+
+            let mut negative_cur_elem = -offset * group_gen.pow([start]);
+            let mut lagrange_coefficients_inverse = vec![F::zero(); size];
+            for coeff in lagrange_coefficients_inverse.iter_mut() {
+                let r_i = tau + negative_cur_elem;
+                *coeff = l_i * r_i;
+                // Increment l_i and negative_cur_elem
+                l_i *= &group_gen_inv;
+                negative_cur_elem *= &group_gen;
+            }
+            ark_ff::fields::batch_inversion(lagrange_coefficients_inverse.as_mut_slice());
+            lagrange_coefficients_inverse
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use ark_bls12_381::Fr;
+    use ark_std::{rand::Rng, One, UniformRand, Zero};
+
+    /// Test that for points in the domain, coefficients are computed correctly
+    #[test]
+    fn test_in_domain_lagrange_coeff() {
+        let mut rng = jf_utils::test_rng();
+        for domain_log_size in 4..9 {
+            let domain_size = 1 << domain_log_size;
+            let domain = Radix2EvaluationDomain::<Fr>::new(domain_size).unwrap();
+            let coset_domain = domain.get_coset(Fr::GENERATOR).unwrap();
+            for (i, (x, coset_x)) in domain.elements().zip(coset_domain.elements()).enumerate() {
+                if i == 0 {
+                    assert_eq!(domain.first_lagrange_coeff(x), Fr::one());
+                    assert_eq!(domain.last_lagrange_coeff(x), Fr::zero());
+                    assert_eq!(coset_domain.first_lagrange_coeff(coset_x), Fr::one());
+                    assert_eq!(coset_domain.last_lagrange_coeff(coset_x), Fr::zero());
+                }
+                if i == domain.size() - 1 {
+                    assert_eq!(domain.last_lagrange_coeff(x), Fr::one());
+                    assert_eq!(domain.first_lagrange_coeff(x), Fr::zero());
+                    assert_eq!(coset_domain.last_lagrange_coeff(coset_x), Fr::one());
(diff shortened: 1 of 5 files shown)


plonk/src/lagrange.rs
```diff
@@ -0,0 +1,224 @@
...
+                * (tau - offset * self.group_gen_inv());
+            z_h_at_tau * self.group_gen_inv() * denominator.inverse().unwrap()
+        }
+    }
+
+    // similar to `EvaluationDomain::evaluate_all_lagrange_coefficients()`
+    //
+    // # Panic
+    // if `range` exceeds the `self.size()`
+    fn lagrange_coeffs_for_range(&self, range: Range<usize>, tau: F) -> Vec<F> {
```
mrain, line 78: Is it worth another fn `lagrange_coeff_at`?

plonk/src/lagrange.rs
```diff
@@ -0,0 +1,224 @@
...
+
+impl<F: FftField> LagrangeCoeffs<F> for Radix2EvaluationDomain<F> {
+    // L_0(tau) = Z_H(tau) * g^0 / (n * h^(n-1) *(tau - h * g^0))
+    // with g^0 = 1
+    // special care when tau in H, as both numerator and denominator is zero
+    fn first_lagrange_coeff(&self, tau: F) -> F {
+        let offset = self.coset_offset();
+        let z_h_at_tau = self.evaluate_vanishing_polynomial(tau);
+        if z_h_at_tau.is_zero() {
+            if tau == F::one() * offset {
```
chancharles92, line 42: How about putting this check before computing z_h_at_tau? Then we can save a eval_vanish_poly ops when tau = h. And for other tau \in H, we can now directly return zero without if-else condition.

plonk/src/lagrange.rs
```diff
@@ -0,0 +1,224 @@
...
+        }
+    }
+
+    // L_n-1(tau) = Z_H(tau) * g^-1 / (n * h^(n-1) * (tau - h * g^-1))
+    // with g^n-1 = g^-1
+    fn last_lagrange_coeff(&self, tau: F) -> F {
+        let offset = self.coset_offset();
+        let z_h_at_tau = self.evaluate_vanishing_polynomial(tau);
+        if z_h_at_tau.is_zero() {
+            if tau == self.group_gen_inv() * offset {
```
chancharles92, line 61: Similar here.

plonk/src/lagrange.rs
```diff
@@ -0,0 +1,224 @@
...
+        let z_h_at_tau = self.evaluate_vanishing_polynomial(tau);
+        if z_h_at_tau.is_zero() {
+            if tau == F::one() * offset {
+                F::one()
+            } else {
+                F::zero()
+            }
+        } else {
+            let denominator = self.size_as_field_element()
+                * offset.pow([self.size() as u64 - 1])
```
chancharles92, line 49: It's a bit wasteful to recompute `offset.pow([self.size()])` in each API call.

plonk/src/lagrange.rs
```diff
@@ -0,0 +1,224 @@
...
+            // We let `l_i = (Z_H(tau) * v_i)^-1` and `r_i = tau - h g^i`
+            // Notice that since Z_H(tau) is i-independent,
+            // and v_i = g * v_{i-1}, it follows that
+            // l_i = g^-1 * l_{i-1}
+
+            let group_gen_inv = self.group_gen_inv();
+            let start = range.start as u64;
+
+            // v_0_inv = n * h^(n-1)
+            let v_0_inv = self.size_as_field_element() * offset.pow([self.size() as u64 - 1]);
```
chancharles92, line 117: Same here for `offset.pow[self.size]`.

plonk/src/lagrange.rs
alxiong: we never use "specific index" api, so i didn't include it. we are also not intending this to be used outside this scope.

alxiong:
@chancharles92 one thing I need to get your opinion:

in `verifier.rs::evaluate_pi_poly()`

```rust
        // TODO: (alex) deal with merged circuit later

        if circuit_is_merged {
            let n = self.domain.size();
            for (i, val) in pub_input.iter().skip(len).enumerate() {
                let lagrange_n_minus_i = vanish_eval_div_n * self.domain.element(n - i - 1)
                    / (*z - self.domain.element(n - i - 1));
                result += lagrange_n_minus_i * val;
            }
        }
```

I have left a todo, where the current `trait LagrangeCoeffs` doesn't handle this logic blackboxly. So I left it untouched.
Do you have any suggestion?

plonk/src/lagrange.rs
```diff
@@ -0,0 +1,224 @@
...
+            let mut l_i = z_h_at_tau.inverse().unwrap() * v_0_inv * group_gen_inv.pow([start]);
+
+            let mut negative_cur_elem = -offset * group_gen.pow([start]);
```
chancharles92, line 120: Usually `inverse()` vs `pow()`, which is more expensive? Alternatively, we can replace `group_gen_inv.pow()` and `group_gen.pow()` with one pow and one inverse.

plonk/src/proof_system/prover.rs
```diff
@@ -783,11 +783,11 @@ impl<E: Pairing> Prover<E> {
         let n = pk.domain_size();
         let m = self.quot_domain.size();
         let domain_size_ratio = m / n;
-        let n_field = E::ScalarField::from(n as u64);
-        let lagrange_n_coeff =
-            self.domain.group_gen_inv / (n_field * (eval_point - self.domain.group_gen_inv));
-        let lagrange_1_coeff =
-            E::ScalarField::one() / (n_field * (eval_point - E::ScalarField::one()));
+        let vanish_eval = self.domain.evaluate_vanishing_polynomial(eval_point);
```
chancharles92, line 786: Would this be more expensive than the previous code?

plonk/src/lagrange.rs
```diff
@@ -0,0 +1,224 @@
...
+            let denominator = self.size_as_field_element()
+                * offset.pow([self.size() as u64 - 1])
+                * (tau - offset);
+            z_h_at_tau * denominator.inverse().unwrap()
+        }
+    }
+
+    // L_n-1(tau) = Z_H(tau) * g^-1 / (n * h^(n-1) * (tau - h * g^-1))
+    // with g^n-1 = g^-1
+    fn last_lagrange_coeff(&self, tau: F) -> F {
```
chancharles92, line 57: minor: there are many repetitive computations between `first_lagrange_coeff()` and `last_lagrange_coeff()`. When we need both coeffs, is it better to call `first_and_langrange_coeff()` or consider it as a special case of `lagrange_coeff_for_range()` API ? (Since {n, 1} is also a special range.)

chancharles92:
> @chancharles92 one thing I need to get your opinion:
> 
> in `verifier.rs::evaluate_pi_poly()`
> 
> ```rust
>         // TODO: (alex) deal with merged circuit later
> 
>         if circuit_is_merged {
>             let n = self.domain.size();
>             for (i, val) in pub_input.iter().skip(len).enumerate() {
>                 let lagrange_n_minus_i = vanish_eval_div_n * self.domain.element(n - i - 1)
>                     / (*z - self.domain.element(n - i - 1));
>                 result += lagrange_n_minus_i * val;
>             }
>         }
> ```
> 
> I have left a todo, where the current `trait LagrangeCoeffs` doesn't handle this logic blackboxly. So I left it untouched. Do you have any suggestion?

How about just adding a simple handle when `*z = self.domain.element(n - i - 1)`?

plonk/src/lagrange.rs
alxiong: > How about putting this check before computing z_h_at_tau? Then we can save a eval_vanish_poly ops when tau = h

sure, I'll do that.

>  And for other tau \in H, we can now directly return zero without if-else condition.

do you suggest we loop over all elements in the group and find matches? the easiest would still be evaluating the vanishing poly (which would still be a if-else)

plonk/src/lagrange.rs
alxiong: iiuc, `inverse` is much more expensive, thus the trick is to try to do `batch_inversion`. 

> we can replace group_gen_inv.pow() and group_gen.pow() with one pow and one inverse.

true, depending on the value of `start` index, the cost of computing one value as pow or as the inverse of the other value might cross over.

I'll use your suggestion

plonk/src/lagrange.rs
alxiong: > consider it as a special case of lagrange_coeff_for_range() API ? (Since {n, 1} is also a special range.)

this is a little hard. 

but I agree that we could probably add a `first_and_last_langrange_coeff()` to save some repeated computation

plonk/src/lagrange.rs
alxiong: true, for us, this won't be a problem, as offset is 1, this would trivially return without any exponentiation.

plonk/src/lagrange.rs
alxiong: on a second thought, the only overlapped computation is `vanish_eval` (the `offset.pow()` is negligible, as we has offset=1), maybe we leave it as it is?

plonk/src/proof_system/prover.rs
alxiong: i think so, because we `* vanish_eval` inside `lagrange_1/n_coeff()`, then divide it outside.. 🤔

plonk/src/lagrange.rs
alxiong: done in [2249534](https://github.com/EspressoSystems/jellyfish/pull/611/commits/224953442d24b72679f615db5b3dbb66aa1f72db)

plonk/src/proof_system/prover.rs
alxiong: here's the cheaper alternative: [4fdde7c](https://github.com/EspressoSystems/jellyfish/pull/611/commits/4fdde7c42e366f27f828b1be4c124a579bc0380f) 

lmk which do you prefer

alxiong:
> How about just adding a simple handle when *z = self.domain.element(n - i - 1)?

just realized that we can use `lagrange_for_range` for this, see [f8f2d5f](https://github.com/EspressoSystems/jellyfish/pull/611/commits/f8f2d5fd56412acc66dcb13db90f2b75fb63c98c)

plonk/src/lagrange.rs
chancharles92: I thought coset_offset is h? And h^{n-1} seems to be expensive. Or did I miss sth?

alxiong: It is. But for our code base, h=1, we didn't utilize the coset from EvalDomain, we defines our own k somewhere I recall, and multiply them manually.

plonk/src/lagrange.rs
```diff
@@ -0,0 +1,228 @@
...
+    /// Returns the first coefficient: `L_{0, Domain}(tau)`
+    fn first_lagrange_coeff(&self, tau: F) -> F;
+    /// Returns the last coefficient: `L_{n-1, Domain}(tau)`
+    fn last_lagrange_coeff(&self, tau: F) -> F;
+    /// Return a list of coefficients for `L_{range, Domain}(tau)`
+    fn lagrange_coeffs_for_range(&self, range: Range<usize>, tau: F) -> Vec<F>;
+}
+
+impl<F: FftField> LagrangeCoeffs<F> for Radix2EvaluationDomain<F> {
+    // L_0(tau) = Z_H(tau) * g^0 / (n * h^(n-1) *(tau - h * g^0))
```
mrain, line 35: very nit: space on both side of *: `(n * h^(n-1) * (tau - h * g^0))`

plonk/src/lagrange.rs
```diff
@@ -0,0 +1,228 @@
...
+    // with g^0 = 1
+    // special care when tau in H, as both numerator and denominator is zero
+    fn first_lagrange_coeff(&self, tau: F) -> F {
+        let offset = self.coset_offset();
+        if tau == F::one() * offset {
```
mrain, line 40: nit: redundant multiplication

plonk/src/lagrange.rs
```diff
@@ -0,0 +1,228 @@
...
+                * (tau - offset);
+            z_h_at_tau * denominator.inverse().unwrap()
+        }
+    }
+
+    // L_n-1(tau) = Z_H(tau) * g^-1 / (n * h^(n-1) * (tau - h * g^-1))
+    // with g^n-1 = g^-1
+    fn last_lagrange_coeff(&self, tau: F) -> F {
+        let offset = self.coset_offset();
+        if tau == self.group_gen_inv() * offset {
```
mrain, line 61: minor: we could early return if `tau` is either `offset` or `offset * g^{-1}`. Feels slightly weird that only one condition is considered in each case.

plonk/src/proof_system/verifier.rs
```diff
@@ -821,23 +820,6 @@ where
         self.domain.evaluate_vanishing_polynomial(*zeta)
     }
 
-    /// Evaluate the first and the last lagrange polynomial at point `zeta`
-    /// given the vanishing polynomial evaluation `vanish_eval`.
-    #[inline]
-    pub(crate) fn evaluate_lagrange_1_and_n(
```
mrain, line 827: Since we are replacing this fn. Maybe it's better to have (an additional) `first_and_last_lagrange_coeffs`?

plonk/src/lagrange.rs
alxiong: we will use https://docs.rs/ark-poly/latest/ark_poly/domain/trait.EvaluationDomain.html#tymethod.coset_offset_pow_size / offset instead to save the pow of high exponent.

plonk/src/proof_system/prover.rs
alxiong: after discussion, we will revert this change and go for the more readable option

plonk/src/lagrange.rs
alxiong: why? the offset in the general case is not one. (in our case where we only use the roots of unity primary subgroup, it is 1)

plonk/src/lagrange.rs
alxiong: i'll consider both first and last conditions in a new API `first_and_last` as you suggested

plonk/src/lagrange.rs
alxiong: done in [caaebf0](https://github.com/EspressoSystems/jellyfish/pull/611/commits/caaebf0e3f9ffb95de1df42ffa1cb97edb69fd8a)

plonk/src/lagrange.rs
alxiong: changed in [caaebf0](https://github.com/EspressoSystems/jellyfish/pull/611/commits/caaebf0e3f9ffb95de1df42ffa1cb97edb69fd8a)

plonk/src/proof_system/prover.rs
alxiong: changed back in [7f27a78](https://github.com/EspressoSystems/jellyfish/pull/611/commits/7f27a78c19f146bfd35167ec2719732f274fdb85)

plonk/src/lagrange.rs
alxiong: added in [70ef5f5](https://github.com/EspressoSystems/jellyfish/pull/611/commits/70ef5f545728e2ef95abf2443abf002a1814f84d)

plonk/src/proof_system/verifier.rs
alxiong: see [70ef5f5](https://github.com/EspressoSystems/jellyfish/pull/611/commits/70ef5f545728e2ef95abf2443abf002a1814f84d)

plonk/src/lagrange.rs
alxiong: fixed all your comments in [70ef5f5](https://github.com/EspressoSystems/jellyfish/pull/611/commits/70ef5f545728e2ef95abf2443abf002a1814f84d)

plonk/src/lagrange.rs
mrain: I mean this could be in all 3 APIs.

alxiong: wait, i don't understand then. can you elaborate?

mrain: oh no worry. my misunderstanding

mrain (approved):

plonk/src/lagrange.rs
chancharles92: Are we gonna using the implementation that saves repeated computation? If we're just calling first_lagrange_coeff() and last_lagrange_coeff() naively, maybe there is no need to add one more API.

chancharles92 (approved):

PR #556, opened 2024-04-16
feat!: crates refactor

mrain:
## Description

We carve out many primitives from `jf-primitives` repo, so that upstream users could selectively upgrade each component without breaking others.

Tasks:
- [x] break `jf-primitives` into smaller crates
- [x] specify CODEOWNERS
- [x] fix all tests
- [x] crate CHANGELOG for each crate
- [x] documentations
- [x] checking combinations of feature flags.
- [x] update the scripts and github workflows 

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [ ] Targeted PR against correct branch (main)
- [ ] Linked to GitHub issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [ ] Wrote unit tests
- [ ] Updated relevant documentation in the code
- [ ] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md`
- [ ] Re-reviewed `Files changed` in the GitHub PR explorer

PR #639, opened 2024-07-25
chore: use tag for inter-dep among member crates

alxiong:
### This PR: 

I encounter type compatibility error when trying to point to previously tagged `jf-plonk-v0.5.0`; where I realized that we are still using relative path `path = "../xx"` for inter-depenecy among member crates, which points to the head instead of a tagged version. 

Since we made the decision in #556 to have independent version management/bumping among crates, we forgot to reflect that in our dependency declaration and this PR fixed that. 

- [ ] re-tag `jf-plonk-v0.5.0` to the latest `main` once this PR is merged.

### DIFF
--- a/aead/Cargo.toml
+++ b/aead/Cargo.toml
@@ -24,7 +24,7 @@ serde = { workspace = true }
 
 [dev-dependencies]
 bincode = "1.3"
-jf-utils = { path = "../utilities" }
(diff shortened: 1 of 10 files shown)


mrain (approved):

alxiong:
I notice that when we tag `0.4.5` we forgot to bump the project root `workspace.version`: https://github.com/EspressoSystems/jellyfish/blob/0.4.5/Cargo.toml#L6

let's be more careful next time.

PR #643, opened 2024-07-31
Add constructor for jf_vdf::MinRootElement

zilong-dai:
Can't create MinRootElement Object when use jf-vdf crate


mrain:
VDF crate is now deprecated in #683 because its security issue and we have no plan to implement the verification part.

PR #665, opened 2024-08-21
Diff Tracking: Sp1 patches

alxiong:
**Not meant to merge into `main` !**

Keep this as a draft PR to track the changes made.

### DIFF
--- a/pcs/src/univariate_kzg/succinct.rs
+++ b/pcs/src/univariate_kzg/succinct.rs
@@ -1,0 +1,67 @@
+use ark_ec::pairing::Pairing;
+use ark_ff::{BigInt, BigInteger, PrimeField, Zero};
+
+#[derive(Clone, Copy, Debug)]
+pub struct BnAffinePoint(pub BigInt<4>, pub BigInt<4>);
+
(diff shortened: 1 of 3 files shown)


pcs/src/univariate_kzg/succinct.rs
```diff
@@ -0,0 +1,67 @@
...
+        infinity: false,
+    }
+}
+
+pub fn msm(p: &[G1Affine], s: &[ScalarField]) -> G1Affine {
+    let mut iter = p.iter().zip(s).filter(|(_, s)| !s.is_zero());
+    let mut result = {
+        let (p, s) = iter.next().unwrap();
+        let mut p = wrap_g1affine(p);
+        bn254_double_and_add(&mut p, s);
```
alxiong, line 24: may i ask why we didn't use `AffinePoint` struct and the `::msm()` directly from `sp1-lib` as suggested in https://github.com/EspressoSystems/zkrollup-integration/issues/21#issuecomment-2292645127?

mrain: I saw this after I had these working implementations. I don't think it's necessary to follow their APIs.

Issue #663, opened 2024-08-20
Fix large shares when small payload and large multiplicity


akonring:
In many settings it may be advantageous to configure the VID scheme to use large multiplicity.
However, the current implementation of the ADVZ scheme will force polynomials to be of size `multiplicity * recovery_threshold` even for small payloads.
Instead, the `disperse` function should use the lowest value for `multiplicity` such that `payload_len` $$\leq$$ `multiplicity * recovery_threshold`.

ggutoski:
There are lots of ways to solve this problem, but all of them are bad ergonomics. After some discussion we decided the least bad solution is what's suggested by OP: silently use the smallest multiplicity needed to fit the entire payload into a single polynomial. We'll rename `multiplicity` -> `max_multiplicity` and update docs to explain that this arg allows the caller to specify only the *maximum* multiplicity that can be used for dispersal.

PR #667, opened 2024-08-23
feat: let multiplicity depend on payload size

akonring:
closes: #663 

### Key places to review:

 

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/vid/src/advz.rs
+++ b/vid/src/advz.rs
@@ -183,6 +183,8 @@ where
         // https://github.com/EspressoSystems/jellyfish/issues/339
         if chunk_size as usize != eval_domain.size() {
             return Err(VidError::Argument(format!(
+                // TODO GUS FIX THIS ERROR CHECK & MESSAGE
+                // TODO GUS EXPLAIN WHY WE ENFORCE POWER OF TWO
                 "recovery_threshold {} currently unsupported, round to {} instead",
                 chunk_size,
                 eval_domain.size()
@@ -858,7 +860,7 @@ where
     fn polynomial_internal<I>(
         domain_ref: &Radix2EvaluationDomain<KzgPoint<E>>,
         chunk_size: usize,
-        coeffs: I,
+        coeffs: I, // TODO GUS rename to evals, they're not coefficients!!
     ) -> KzgPolynomial<E>
     where
         I: Iterator,
@@ -876,6 +878,7 @@ where
         // then we were not given the correct number of coeffs. In that case
         // coeffs.len() could be anything, so there's nothing to sanity check.
         if pre_fft_len == chunk_size {
+            // TODO GUS don't panic, return internal error instead
             assert_eq!(coeffs_vec.len(), pre_fft_len);
         }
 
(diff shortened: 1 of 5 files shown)


vid/src/advz.rs
```diff
@@ -741,12 +762,12 @@ where
     fn evaluate_polys(
         &self,
         polys: &[DensePolynomial<<E as Pairing>::ScalarField>],
+        code_word_size: usize,
```
ggutoski, line 765: I wonder whether this arg should be `multiplicity` instead of `code_word_size`. Then we refactor the multiplication  `num_storage_nodes * multiplicity` inside this function instead of relying on the caller to do it each time. Also, at first glance it seems the purpose of most/all use of `code_word_size` outside this function is to prepare an arg for this function or `bytes_to_polys` or `polynomial`. In that case it's a no-brainier.

vid/src/advz.rs
```diff
@@ -858,16 +882,35 @@ where
...
+        let recovery_threshold = self.recovery_threshold;
+        if recovery_threshold * multiplicity < elems {
+            // payload is large. no change in multiplicity needed.
+            return multiplicity;
+        }
+
+        // payload is small: choose m such that 0 < m < multiplicity
+        // and m is the power of two that fits the payload best.
+        let mut m = multiplicity;
+        loop {
```
ggutoski, line 906: ok I know this loop has only logarithmic complexity but I'm pretty sure we can replace it with a call to `ilog2` or something.

ggutoski (commented):
Great start! Seems that you've completed changes to production code; it remains only to write tests + doc. I have only a couple nits. As discussed, I'll take it from here. Thank you!

vid/src/advz.rs
ggutoski: This loop has a bug: the `if` condition is always met so the loop it always breaks immediately with `m = multiplicity`. Fixed in https://github.com/EspressoSystems/jellyfish/pull/667/commits/17d55fba72b2cb4739fb417525c57ac41e9523e0.

ggutoski:
Status as of https://github.com/EspressoSystems/jellyfish/pull/667/commits/17d55fba72b2cb4739fb417525c57ac41e9523e0 :

The new test `max_multiplicity` fails with
```
thread 'advz::test::max_multiplicity' panicked at vid/src/advz.rs:882:13:
assertion `left == right` failed
  left: 4096
 right: 4
```
Here's the code:
https://github.com/EspressoSystems/jellyfish/blob/17d55fba72b2cb4739fb417525c57ac41e9523e0/vid/src/advz.rs#L876-L883

The problem is that we're still calling FFT with the original `self.eval_domain`, which auto-magically resizes the polynomial degree to the size of `self.eval_domain`, which was originally set in the `Advz` constructor according to `max_multiplicity`. In the above test failure a small payload of size `4` field elements is resized to a large payload of size `4096`.

Any fix for this issue must also address the above problem. But that's a bigger task. We now need the eval domain to change size at `disperse`-time to accommodate a small payload.

# Options
- Construct small domains on-the-fly as needed at `disperse`-time: if at `disperse`-time we find that `self.eval_domain` is too big then construct a new small domain, otherwise use `self.eval_domain`.
  - It would be nice of arkworks `EvaluationDomain` has a convenient way to construct a subdomain from a given domain. But I don't see such a method. Fortunately the code to construct a new domain is not difficult.
- Perhaps we should delete the field `eval_domain` from struct `Advz`. Instead, we *always* construct this domain on-the-fly with the correct size at `disperse`-time. But that's a nontrivial change so I prefer to discuss with others.

vid/src/advz.rs
akonring: Yes, I had initially just passed in `multiplicity` as argument to the functions but changed it again since IMO fx
`let all_storage_node_evals = self.evaluate_polys(&polys, code_word_size as usize)?;` provides slightly better readability than just passing in the `multiplicity`.

vid/src/advz.rs
akonring: Yes, you are absolutely right. Thanks for spotting this. This should have been `m=1` and then left-shift (`m<<1`) but your approach is definitely superior and more readable, imo.

ggutoski:
closing so that I can open a new PR for this branch so as to better facilitate review from others.

Issue #671, opened 2024-08-29
Eliminate unnecessary merkle proofs in ADVZ shares


ggutoski:
This issue drawn from https://github.com/EspressoSystems/jellyfish/pull/657#discussion_r1720327956

Currently each share now has `multiplicity` distinct merkle paths instead of just 1. That's a waste of bandwidth and it breaks backwards compatibility with jellyfish tag `0.4.5`.

We should have only a *single* merkle proof for *all* evals.

ggutoski:
The fix #657 breaks backwards compatibility with jellyfish tag `0.4.5` for serialization of ADVZ's `Share` struct because this field
https://github.com/EspressoSystems/jellyfish/blob/7d71dbeff14f1a501b0b0dc391f1dffa1b8374fb/vid/src/advz.rs#L313
...became this
https://github.com/EspressoSystems/jellyfish/blob/7cd4f76b0844d4efdf9179fcdd7d4cae44bf7400/vid/src/advz.rs#L314

It seems impossible restore backwards compatibility. Any fix for this issue that restores `evals_proof` to its original type must change how the merkle tree is built, in which case future versions of jf-vid will disagree with past versions on how to build the merkle tree.

Given that we must break backwards compatibility somewhere, we should fix this issue and break how the merkle tree is built. The only alternative is to not fix this issue, in which case we continue to unnecessarily bloat share size by including unneeded merkle paths.

Issue #654, opened 2024-08-15
VID: verify consistency with the merkle root for all the evaluations in a VID share


akonring:
Discussion of issue here: https://github.com/EspressoSystems/jellyfish/pull/653#issue-2466614905 

We don't check consistency with the merkle root for all the evaluations in a share (with multiplicity). The share only contains the merkle path for the first evaluation.

If we consider doing KZG eval batching in the future, we should represent the VID share such that consistency check with the merkle root can be done for all evaluations still.

PR #678, opened 2024-09-05
fix!: eliminate unnecessary merkle proofs in ADVZ shares

ggutoski:
closes: #671 

### This PR: 
- `Advz::evals_proof` is no longer a `Vec`
- structure of the merkle tree is changed accordingly. Each leaf in the merkle tree is a vec of polynomial evaluations as before, except now that vec looks like: https://github.com/EspressoSystems/jellyfish/blob/689555e07860c5ed6e54b467c76d9a1776f0f7b5/vid/src/advz.rs#L702-L715
- delete test `sad_path_verify_share_with_multiplicity` as it is a regression for #654 , which is no longer relevant as there is now only one merkle path in each share. 

### Key places to review:

 

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/vid/src/advz.rs
+++ b/vid/src/advz.rs
@@ -478,13 +478,17 @@ where
             return Ok(Err(()));
         }
 
+        // the magic pseudorandom scalar used for aggregation
         let pseudorandom_scalar = Self::pseudorandom_scalar(common, commit)?;
 
-        // Compute aggregate polynomial [commitment|evaluation]
-        // as a pseudorandom linear combo of [commitments|evaluations]
-        // via evaluation of the polynomial whose coefficients are
-        // [commitments|evaluations] and whose input point is the pseudorandom
-        // scalar.
+        // Compute an aggregate polynomial [commitment|evaluation] as a
+        // pseudorandom linear combo of [commitments|evaluations].
+        // To this end use function `polynomoial_eval` where each "coefficient"
+        // is actually a [commitment|evaluation] and the input point is
+        // `pseudorandom_scalar`.
+        //
+        // - Here: aggregate polynomial commitment in `aggregate_poly_commit`.
+        // - Below: aggregate polynomial evals in `aggregate_eval`.
         let aggregate_poly_commit = KzgCommit::<E>::from(
             polynomial_eval(
                 common
@@ -496,13 +500,12 @@ where
             .into(),
         );
 
-        // verify aggregate proofs
-        //
-        // some boilerplate needed to accommodate builds without `parallel`
-        // feature.
+        // convenience quantities
         let multiplicities = Vec::from_iter((0..multiplicity_usize));
         let polys_len = common.poly_commits.len();
         let multi_open_domain = self.multi_open_domain(multiplicity)?;
+
+        // verify aggregate proofs
         let verification_iter = parallelizable_slice_iter(&multiplicities).map(|m| {
             let evals_iter = evals.iter().skip(*m).step_by(multiplicity_usize);
             let aggregate_eval =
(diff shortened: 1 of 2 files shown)


vid/src/advz.rs
```diff
@@ -515,35 +496,25 @@ where
...
-                    .iter()
-                    .map(FieldMultiplier),
-                pseudorandom_scalar,
-            );
+        let verification_iter = parallelizable_slice_iter(&multiplicities).map(|m| {
+            let evals_iter = evals.iter().skip(*m).step_by(multiplicity_usize);
+            let aggregate_eval =
+                polynomial_eval(evals_iter.map(FieldMultiplier), pseudorandom_scalar);
+            let domain_index = usize::try_from(share.index).map_err(vid)?
+                + m * usize::try_from(self.num_storage_nodes).map_err(vid)?;
```
akonring, line 511: nit: we could use "convenience quantities" here as well.

vid/src/advz/test.rs
```diff
@@ -254,51 +263,9 @@ fn verify_share_with_multiplicity() {
...
-        num_storage_nodes: 20,
-        max_multiplicity: 32, // payload fitting into a single polynomial
-        payload_len: 8200,
-    };
-    let (mut advz, payload) = advz_init_with::<Bn254>(advz_params);
-
-    let disperse = advz.disperse(payload).unwrap();
-    let (shares, common, commit) = (disperse.shares, disperse.common, disperse.commit);
-    for (i, share) in shares.iter().enumerate() {
-        // corrupt the last evaluation of the share
```
akonring, line 275: Do we maybe want to keep some version of this test? Seems important to exercise the sad path for `multiplicity > 1`

vid/src/advz.rs
```diff
@@ -715,27 +689,41 @@ where
...
 
-            // distribute evals from each poly among the storage nodes
+            // Populate a Vec of polynomial evaluations for all storage nodes:
+            //
+            // The `i`th item is a Vec of `polys.len() * multiplicity`
+            // polynomial evaluations.
             //
-            // perf warning: runtime is quadratic in payload_size
-            let mut all_storage_node_evals = vec![Vec::with_capacity(polys.len()); code_word_size];
+            // Evaluations for storage node `i` are ordered as follows. Define
```
akonring, line 706: Great doc!

akonring (approved):
Nice.
Don't fully understand the motivation for moving to a different "share evaluation layout".
But they seem equivalent and the docs added are great and needed. Thanks.

vid/src/advz.rs
ggutoski: Thanks!

> Don't fully understand the motivation for moving to a different "share evaluation layout".
But they seem equivalent and the docs added are great and needed. Thanks.

The only "motivation" is I had to rewrite this part from scratch and this is the first thing I coded. I don't know whether there's a reason to prefer one layout over another. But if you have a reason then perhaps we could change it.

vid/src/advz.rs
ggutoski: done in https://github.com/EspressoSystems/jellyfish/pull/678/commits/7a37893155ec163efa6a8e5553a3559609848c70

vid/src/advz/test.rs
ggutoski: done in https://github.com/EspressoSystems/jellyfish/pull/678/commits/52e7c17e87bf4b7c88ef0703163684dfb054fa14. All tests now use nontrivial multiplicity. (Set in `advz_init`.)

akonring (approved):

PR #680, opened 2024-09-10
chore: bump jf-vid version, update changelog

ggutoski:
- bump `jf-vid` package to `0.2.0`
- update CHANGELOG for `jf-vid`. Include only nontrivial PRs relevant to jf-vid.
- See discussion below on version number vs git tags: https://github.com/EspressoSystems/jellyfish/pull/680#discussion_r1752504845

### DIFF
--- a/vid/CHANGELOG.md
+++ b/vid/CHANGELOG.md
@@ -6,3 +6,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## 0.1.0
 
 - Initial release. Verifiable information dispersal trait definition and implementations.
+
+## 0.2.0
+
+- [#650](https://github.com/EspressoSystems/jellyfish/pull/650) perf: VID ADVZ verify_share use parallelism over multiplicity
+- [#653](https://github.com/EspressoSystems/jellyfish/pull/653) fix: VID ADVZ check consistency of multiplicity in verify_share
+- [#657](https://github.com/EspressoSystems/jellyfish/pull/657) fix: check eval_proof for all evaluations
+- [#670](https://github.com/EspressoSystems/jellyfish/pull/670) feat: multiplicity depend on payload size
+- [#674](https://github.com/EspressoSystems/jellyfish/pull/674) fix: Delete field Share::evals, extract data from eval_proofs instead
+- [#678](https://github.com/EspressoSystems/jellyfish/pull/678) fix!: eliminate unnecessary merkle proofs in ADVZ shares
--- a/vid/CHANGELOG.md
+++ b/vid/CHANGELOG.md
@@ -3,15 +3,50 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## 0.1.0
+# Unreleased
 
-- Initial release. Verifiable information dispersal trait definition and implementations.
+## Breaking changes
 
-## 0.2.0
+## Added
 
-- [#650](https://github.com/EspressoSystems/jellyfish/pull/650) perf: VID ADVZ verify_share use parallelism over multiplicity
-- [#653](https://github.com/EspressoSystems/jellyfish/pull/653) fix: VID ADVZ check consistency of multiplicity in verify_share
-- [#657](https://github.com/EspressoSystems/jellyfish/pull/657) fix: check eval_proof for all evaluations
-- [#670](https://github.com/EspressoSystems/jellyfish/pull/670) feat: multiplicity depend on payload size
-- [#674](https://github.com/EspressoSystems/jellyfish/pull/674) fix: Delete field Share::evals, extract data from eval_proofs instead
-- [#678](https://github.com/EspressoSystems/jellyfish/pull/678) fix!: eliminate unnecessary merkle proofs in ADVZ shares
+## Deprecated
+
+## Removed
+
+## Fixed
+
+## Performance
+
+## Security
+
+# 0.2.0 (2024-09-11)
+
+## Breaking changes
+
+- [#670](https://github.com/EspressoSystems/jellyfish/pull/670) ADVZ multiplicity depend on payload size
+  - Largely a performance improvement, but technically a breaking change because ADVZ no longer guarantees to use the exact `multiplicity` specified by the caller. Instead, this arg is now `max_multiplicity`; dispersal might use a smaller multiplicity for small payloads.
+- [#674](https://github.com/EspressoSystems/jellyfish/pull/674) ADVZ delete field Share::evals, extract data from eval_proofs instead
+  - Change serialization of ADVZ shares.
+- [#678](https://github.com/EspressoSystems/jellyfish/pull/678) ADVZ eliminate unnecessary merkle proofs in ADVZ shares
+  - Change serialization of ADVZ shares.
+
+## Fixed
+
+- [#653](https://github.com/EspressoSystems/jellyfish/pull/653) ADVZ check consistency of multiplicity in verify_share
+
+## Performance
+
+- [#650](https://github.com/EspressoSystems/jellyfish/pull/650) ADVZ verify_share use parallelism over multiplicity
+- [#670](https://github.com/EspressoSystems/jellyfish/pull/670) ADVZ multiplicity depend on payload size
+  - Also a breaking change.
+
+## Security
+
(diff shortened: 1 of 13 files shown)


vid/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "jf-vid"
-version = "0.1.0"
+version = "0.2.0"
```
ggutoski, line 3: # Version numbers vs git tags

[Copied from internal discussion.]

Currently each package in the jellyfish workspace (eg jf-vid, jf-pcs, etc) specifies its own version number (eg `0.1.0`, etc). Moreover, some jellyfish packages depend on other jellyfish packages. Such a dependency is currently specified in Cargo.toml like so:
```toml
jf-pcs = { git = "https://github.com/EspressoSystems/jellyfish", tag = "0.4.5" }
```
External workspaces (eg espresso-sequencer) are similar, except they might also specify a version number:
```toml
jf-pcs = { version = "0.1.0", git = "https://github.com/EspressoSystems/jellyfish", tag = "0.4.5" }
```
I think this is not best practice. According to [The Cargo Book](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#the-role-of-the-version-key):

> The `version` key does *not* affect which commit is used when Cargo retrieves the `git` dependency, but Cargo checks the version information in the dependency’s `Cargo.toml` file against the `version` key and raises an error if the check fails.

Thus, no user of a jellyfish package (either inside or outside the jellyfish workspace) ever actually uses the package's version number! (Except possibly as a sanity check against a git commit.)

Also, it's awkward to bump a version number of a package in the jellyfish workspace. Any jellyfish package (eg `jf-vid`) with a dependency on another jellyfish package (eg. `jf-pcs`) must specify a git tag for an older commit of `jf-pcs`. Thus, the jellyfish workspace is not internally consistent: packages are not guaranteed to build against a single, fixed jellyfish commit. Any user of `jf-vid` must necessarily have at least two jellyfish git tags in its dependency graph. :man_facepalming: 

Opinions? @alxiong @mrain

alxiong: > Thus, no user of a jellyfish package (either inside or outside the jellyfish workspace) ever actually uses the package's version number!

it's true that version number is ignored for now, but when we eventually publish these crates, the version number will matter. Before publishing, we sync the tag name with the version number as `git = "", tag = ""` pins to a specific version.

>  Thus, the jellyfish workspace is not internally consistent: packages are not guaranteed to build against a single, fixed jellyfish commit. 

not true? we always created tag from some commit on `main` (that commit is the single commit that is guaranteed to be built successfully). 
e.g. https://github.com/EspressoSystems/jellyfish/releases/tag/jf-plonk-v0.5.1, after adding this tag, we didn't force downstream to update all `jf-*` to point to this tag (if they do, it would still compile), but only `jf-plonk` to point to it. because the `jf-plonk` at this commit is pointing to some earlier tag of other crates `jf-*`.

if we update `jf-crhf` and cause `jf-plonk` to break, we would bump versions for both.


> Any user of jf-vid must necessarily have at least two jellyfish git tags in its dependency graph.

that's expected outcome when we decide to use separate version bumps for every package/crates. I don't see this as a problem -- it happens in your nix/store as well. Different tools depend on different versions of `curl`, thus you have multiple versions of `curl` under the same nix shell environment. 

I also disagree with the definition of "internally consistency" implied. so

vid/CHANGELOG.md
```diff
@@ -6,3 +6,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## 0.1.0
 
 - Initial release. Verifiable information dispersal trait definition and implementations.
+
+## 0.2.0
+
+- [#650](https://github.com/EspressoSystems/jellyfish/pull/650) perf: VID ADVZ verify_share use parallelism over multiplicity
```
alxiong, line 12: nit: 

- put newer version on top (0.2.0 before 0.1.0), reverse chronological 
- remove `feat/feat/..` prefix, those are for commit naming, here we just succinctly describe the changes
- i'd personally divide these into "breaking", "bugs", "non-breaking" such as [this](https://github.com/EspressoSystems/jellyfish/blob/main/plonk/CHANGELOG.md#breaking-changes-1); our [older version](https://github.com/EspressoSystems/jellyfish/blob/main/CHANGELOG_OLD.md#added) is even more granular, but maybe unnecessary

ggutoski: done in https://github.com/EspressoSystems/jellyfish/pull/680/commits/93c0ade3d1ee1df0aff364683be549982e4926b6

vid/Cargo.toml
ggutoski: > Suggestion: add `version` key to all `jf-*` dependencies in the jellyfish repo. This suggestion is more consistent with idiomatic Rust, and will make it easier to switch from git commit tags to version numbers when jellyfish crates eventually get published to a registry.

done in https://github.com/EspressoSystems/jellyfish/pull/680/commits/dc7130c09b7de6baf3c5c0a21c75e67879e10a11

alxiong: i'm a little mixed with "adding version key", it reads confusing. `version = 0.4.4, tag = 0.4.5`
this is the idiomatic way?

ggutoski: That's only in `jf-utils` package. Unlike all other packages, `jf-utils` inherits its version from the workspace:
https://github.com/EspressoSystems/jellyfish/blob/7d71dbeff14f1a501b0b0dc391f1dffa1b8374fb/utilities/Cargo.toml#L4
...and the workspace at git tag `0.4.5` declares its version to be `0.4.4` 🤦 :
https://github.com/EspressoSystems/jellyfish/blob/7d71dbeff14f1a501b0b0dc391f1dffa1b8374fb/Cargo.toml#L6

I propose that `jf-utils` do not inherit its version from the workspace. Instead, we should reset its version to `0.1.0` like all the other packages in this repo.

# Question

What should we do with the workspace-level version? It seems kinda useless now because each package explicitly sets its own version. Options:

1. Set it to `0.4.6` in so that it matches a future new commit tag for this repo. But this version is meaningless because no package actually uses it.
2. Set it to `0.1.0`, since that's the version number for most packages in this workspace. If we do this, do we then revisit all packages currently set to `0.1.0` and change them to inherit from the workspace? Seems obfuscated for no real benefit.
3. Delete `version` key entirely from the workspace.

alxiong: > I propose that jf-utils do not inherit its version from the workspace. Instead, we should reset its version to 0.1.0 like all the other packages in this repo.

i agree. cc @mrain 

> Delete version key entirely from the workspace.

I choose this one. since RustCrypto does so. https://github.com/RustCrypto/hashes/blob/master/Cargo.toml, separate version management for each member.

ggutoski: Deleted version key from workspace in https://github.com/EspressoSystems/jellyfish/pull/680/commits/ba7ffb72e51bb4311663bf61a724dbc7059e2620.

On further consideration, I think we should keep jf-utils at 0.4.4. This package has existed since the beginning, so to reset its version number now would be bad. There is precedent: jf-plonk also kept its 0.4.4 version number after the restructuring of this workspace. (jf-plonk is now at 0.5.1 whereas jf-utils is still at 0.4.4.)

Unfortunately, the stupidity remains with `version = 0.4.4, tag = 0.4.5` in downstream Cargo.toml files. 😕 

# Proposal

Now that package version numbers are completely decoupled from the workspace, we should stop naming jellyfish tags according to semver because it creates confusion. I don't know what tag name we should use instead. It's hard to pick a good name for tags because we need a new tag each time *any* of the packages bumps its version. One idea is to create a separate tag for each package, so that downstream Cargo.toml files look like
```toml
jf-rescue = { version = "0.1.0", tag = "jf-rescue-0.1.0", git = ...
jf-utils = { version = "0.4.4", tag = "jf-utils-0.4.4", git = ...
```
Initially, most of these tags will point to a single commit. For example, both `jf-rescue-0.1.0` and `jf-utils-0.4.4` will point to the commit currently tagged as `0.4.5`. Yes, this means lots of redundant tags. But the result is much more readable and the usage pattern is intuitive.

Opinions? @alxiong @mrain

alxiong: > One idea is to create a separate tag for each package, 

agree, that's exactly what we've been doing: https://github.com/EspressoSystems/jellyfish/tags (and what RustCrypto does: https://github.com/RustCrypto/hashes/tags)

alxiong (approved):

Issue #649, opened 2024-08-07
[security] MinRoot delay function is "vulnerable"


mrain:
Existing attacks break the delay property of MinRoot function. However, currently it costs dozens of millions to launch such an attack.

Since we do not care about verifiability, we could simply switch to a hash chain for our delay function.

Ref: 
- https://ethresear.ch/t/statement-regarding-the-public-report-on-the-analysis-of-minroot/16670
- https://crypto.ethereum.org/events/minrootanalysis2023.pdf

PR #683, opened 2024-09-13
revert!: deprecate VDF crate

mrain:
closes: #649 

### This PR: 
Deprecates the VDF crate because the original implemented MinRoot is insecure. Also we didn't and won't implement the verification part.

### Key places to review:

 

---

write a little note why.

- [x] Wrote unit tests


vdf/src/hashchain.rs
```diff
@@ -0,0 +1,88 @@
...
+impl VDF for HashChain {
+    type PublicParameter = HashChainParam;
+    type Proof = [u8; 32];
+    type Input = [u8; 32];
+    type Output = [u8; 32];
+
+    fn setup<R: ark_std::rand::CryptoRng + ark_std::rand::RngCore>(
+        difficulty: u64,
+        prng: Option<&mut R>,
+    ) -> Result<Self::PublicParameter, VDFError> {
```
ss-es, line 48: Is it possible to make `setup` and `eval` infallible (not return a `Result`)? It looks like they can't fail in either implementation

not really a big deal: we can of course always handle the failure in HotShot

vdf/src/hashchain.rs
```diff
@@ -0,0 +1,88 @@
...
+        Ok(HashChainParam { difficulty })
+    }
+
+    fn eval(
+        pp: &Self::PublicParameter,
+        input: &Self::Input,
+    ) -> Result<(Self::Output, Self::Proof), VDFError> {
+        let mut output = *input;
+        for _ in 0..pp.difficulty {
+            output = sha3::Keccak256::digest(&input).into();
```
ss-es, line 58: maybe I'm not reading this right, but should this be `digest(&output)`?

mrain: stupid me! fixed in [ffc603b](https://github.com/EspressoSystems/jellyfish/pull/683/commits/ffc603be8be262428dee883c35b5ac8520f9b619)

vdf/src/hashchain.rs
mrain: I prefer to keep it. Error may occur when we are implementing Verifiable DF.

alxiong (commented):
1. why are we keeping the minroot? afaik, no one, not even ourselves, is using it
2. why sh3 hashchain? is it being used or planned anywhere?

mrain:
> 1. why are we keeping the minroot? afaik, no one, not even ourselves, is using it

No particular reason. Maybe we'll comeback to it sometime later (very not likely)

> 2. why sh3 hashchain? is it being used or planned anywhere?

I just picked a random hash function. Do you have other suggestions or would you prefer a somewhat generic implementation cc @ss-es

alxiong (approved):
I agree with the removal.

PR #685, opened 2024-09-18
feat: add keccak256 merkle tree

alxiong:
### This PR: 

- introduce macro rules `impl_mt_hash_256!` for implementing boilerplate for node and digest for MT that has 32 bytes output size
- impl for both Sha3 and Keccak256 under this macro
- expose a MT and lightweight MT for keccak256

@mrain do you remember why we pick rate 3 for MT? is that an assumption for our current MT internals?

---

write a little note why.

- [x] ~Wrote unit tests~

### DIFF
--- a/merkle_tree/src/prelude.rs
+++ b/merkle_tree/src/prelude.rs
@@ -22,7 +22,7 @@ use ark_serialize::{
     CanonicalDeserialize, CanonicalSerialize, Compress, Read, SerializationError, Valid, Validate,
     Write,
 };
-use ark_std::{marker::PhantomData, vec::Vec};
+use ark_std::{fmt, marker::PhantomData, vec::Vec};
 use jf_rescue::{crhf::RescueCRHF, RescueParameter};
 use sha3::{Digest, Keccak256, Sha3_256};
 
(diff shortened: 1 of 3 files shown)


mrain:
> @mrain do you remember why we pick rate 3 for MT? is that an assumption for our current MT internals?

Rescue hash has a rate of 3.

@sveitser I think it's better to instantiate a MT in the application since it's simple enough. If we do it here, people may think that we are enforcing some default instantiations while we are not.

alxiong:
> Rescue hash has a rate of 3.

but our Sha3MT doesn't have to use rate of 3, right? also our implementation is generic of ARITY. I don't know why our Sha3Tree also choose Arity of 3

> @sveitser I think it's better to instantiate a MT in the application since it's simple enough. If we do it here, people may think that we are enforcing some default instantiations while we are not.

Softly disagree, the fact that our API is generic already offers flexibility. `prelude::*` offers good out-of-box options that can be safely ignored. 
I don't see downside of adding `Keccak256MerkleTree` in jellyfish actually.

mrain:
> but our Sha3MT doesn't have to use rate of 3, right? also our implementation is generic of ARITY. I don't know why our Sha3Tree also choose Arity of 3

I just picked a random number because it was supposed to be only an example. If we are making it public we'd better make it generic to ARITY.

> Softly disagree, the fact that our API is generic already offers flexibility. `prelude::*` offers good out-of-box options that can be safely ignored. I don't see downside of adding `Keccak256MerkleTree` in jellyfish actually.

We could go either way.

merkle_tree/src/prelude.rs
```diff
@@ -52,71 +52,90 @@ pub type RescueLightWeightMerkleTree<F> = LightWeightMerkleTree<F, RescueHash<F>
...
+/// # Usage
+/// `impl_mt_hash_256!(Sha3_256, Sha3Node, Sha3Digest)` will
+/// - introduce `struct Sha3Node` for internal node in a merkle tree that use
+///   Sha3_256 as hash
+/// - introduce `struct Sha3Digest` which implements `DigestAlgorithm`
+macro_rules! impl_mt_hash_256 {
+    ($hasher:ident, $node_name:ident, $digest_name:ident) => {
+        /// Internal node for merkle tree
+        #[derive(Default, Eq, PartialEq, Clone, Copy, Debug, Ord, PartialOrd, Hash)]
+        pub struct $node_name(pub(crate) [u8; 32]);
```
mrain, line 67: nit: we could have some nice `Display`/`Debug` for this type to print a hex string.

mrain (dismissed):
LGTM. I think we could include this change in version `0.2.0`, and make a new tag after merge.

merkle_tree/src/prelude.rs
alxiong: done in [95584ac](https://github.com/EspressoSystems/jellyfish/pull/685/commits/95584ac7a41782747bbf3439316e97a7ad688988)

mrain (approved):
LGTM!

Issue #642, opened 2024-07-29
Ergonomics of jf-merkle-tree


mrain:
- [ ] https://github.com/EspressoSystems/jellyfish/blob/main/merkle_tree/src/lib.rs#L232-L236: verification should be taking `Self::Commitment` instead of a `NodeValue`.
- [ ] [MAYBE NOT] https://github.com/EspressoSystems/jellyfish/blob/main/merkle_tree/src/internal.rs#L101: `MerkleTreeCommitment` should have an API to verify a `MerkleProof`. Same for `MerkleProof` struct. 
- [ ] #658

Issue #658, opened 2024-08-16
`MerkleProof` should not include position or leaf value


ggutoski:
This issue is an opinionated suggestion. Others may disagree.

`MerkleProof` struct contains:
- 2 copies of index position
- leaf value

These items should not be in the merkle proof. Instead, `MerkleTree::verify` should take them as separate args. If the leaf value is large then the size of the merkle proof is unnecessarily large.

First copy of index position is `pos` field in `MerkleProof struct:
https://github.com/EspressoSystems/jellyfish/blob/92714a4cc509fac07b8e8fc321fc0271c5dbe6b6/merkle_tree/src/internal.rs#L139-L150

Digging into `MerklePath` we see

https://github.com/EspressoSystems/jellyfish/blob/92714a4cc509fac07b8e8fc321fc0271c5dbe6b6/merkle_tree/src/internal.rs#L84

where `MerkleNode` is an enum

https://github.com/EspressoSystems/jellyfish/blob/92714a4cc509fac07b8e8fc321fc0271c5dbe6b6/merkle_tree/src/internal.rs#L25

with `Leaf` variant

https://github.com/EspressoSystems/jellyfish/blob/92714a4cc509fac07b8e8fc321fc0271c5dbe6b6/merkle_tree/src/internal.rs#L37-L47

where we find a second copy of `pos` and the leaf value `elem`.

If, as I suggest, we remove these occurrences of `pos` then `MerkleProof` becomes just a wrapper for `MerklePath`. In that case we should eliminate the wrapper and just use `MerklePath`.

I suggest that `MerklePath` be an opaque struct. No need to expose the fact that it's a `Vec<MerkleNode>`.

mrain:
I tend to keep it in the original form. But this comment is valuable:
> If the leaf value is large then the size of the merkle proof is unnecessarily large.

ggutoski:
Forgot to mention: currently `MerkleTree::verify` takes a redundant arg `pos`. The current implementation merely checks `pos` against one of its copies inside `MerkleProof`:

https://github.com/EspressoSystems/jellyfish/blob/92714a4cc509fac07b8e8fc321fc0271c5dbe6b6/merkle_tree/src/macros.rs#L81-L90

Instead `verify` args should include both `pos` and `elem` or neither of these items. Seems strange to take only one but not the other. Naturally, my suggestion is to include both and then remove them from `MerkleProof`.

mrain:
[...]

PR #692, opened 2024-10-15
feat: refactor Merkle proof for ergonomics

mrain:
closes: #642, #658

### This PR: 
- Remove leaf information (index, element) from the `MerkleProof` struct.

### Key places to review:

 

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/merkle_tree/src/internal.rs
+++ b/merkle_tree/src/internal.rs
@@ -7,7 +7,7 @@
 use super::{
     DigestAlgorithm, Element, Index, LookupResult, MerkleCommitment, NodeValue, ToTraversalPath,
 };
-use crate::{errors::MerkleTreeError, VerificationResult};
+use crate::{errors::MerkleTreeError, prelude::MerkleTree, VerificationResult};
 use alloc::sync::Arc;
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 use ark_std::{borrow::Borrow, format, iter::Peekable, string::ToString, vec, vec::Vec};
@@ -46,7 +46,7 @@ pub enum MerkleNode<E: Element, I: Index, T: NodeValue> {
         elem: E,
     },
     /// The subtree is forgotten from the memory
-    ForgettenSubtree {
+    ForgottenSubtree {
         /// Merkle hash value of this forgotten subtree
         #[serde(with = "canonical")]
         value: T,
@@ -70,18 +70,15 @@ where
                 elem: _,
             } => *value,
             Self::Branch { value, children: _ } => *value,
-            Self::ForgettenSubtree { value } => *value,
+            Self::ForgottenSubtree { value } => *value,
         }
     }
 
     #[inline]
     pub(crate) fn is_forgotten(&self) -> bool {
-        matches!(self, Self::ForgettenSubtree { .. })
+        matches!(self, Self::ForgottenSubtree { .. })
     }
 }
-
-/// A merkle path is a bottom-up list of nodes from leaf to the root.
-pub type MerklePath<E, I, T> = Vec<MerkleNode<E, I, T>>;
 
 /// A merkle commitment consists a root hash value, a tree height and number of
 /// leaves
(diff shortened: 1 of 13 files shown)


ggutoski:
> Change the verifications APIs to take a Merkle commitment instead of a simple root digest.

Why is this needed?

mrain:
> Why is this needed?

I think it looks nicer.

ggutoski:
> And make sure that users won't miss this check: https://github.com/EspressoSystems/zkrollup-integration/blob/89e86d0ea3f6ea25b92e310d131d177512b0ff80/sp1/program/src/main.rs#L82

Sorry but I don't understand this comment.

mrain:
> Sorry but I don't understand this comment.

This is an additional check that the height information is correct. With previous API users may forget. to do so.
But I'm not quite sure whether it is necessary now. cc @alxiong 

> Anything specific to review with greater care? Seems like most changes are boilerplate: renaming variables, accommodating the removal of (index, element) from merkle proof, etc.

True. Most of them are boilerplate. Changes in `forget_internal()` and `remember_internal()` may require additional care. But they are all covered by unit tests. And this is also kind of internal audit.

alxiong:
> >  Sorry but I don't understand this comment.
>
> This is an additional check that the height information is correct. With previous API users may forget. to do so.
But I'm not quite sure whether it is necessary now. cc @alxiong

This is to ensure correct tree height is being used? to prevent extension attack (i.e. find a root collision via much larger tree) 

meanwhile, I agree it's unclear from API design that why we need this check outside, instead of being part of `BlockMerkleTree::verify()` 🤔

mrain:
> meanwhile, I agree it's unclear from API design that why we need this check outside, instead of being part of BlockMerkleTree::verify() 🤔

Therefore it's checked inside `verify()` now.
My point is, as we are supplying the leaf information to `verify()`, is the height check still necessary to prevent the extension attack?

merkle_tree/src/internal.rs
```diff
@@ -131,52 +128,76 @@ impl<T: NodeValue> MerkleCommitment<T> for MerkleTreeCommitment<T> {
...
+                        let mut data = values.clone();
+                        data.insert(*branch, val);
+                        H::digest(&data)
+                    }
+                },
+            )?;
+        if computed_root == commitment.digest() {
+            Ok(Ok(()))
+        } else {
+            Ok(Err(()))
```
ggutoski, line 195: return an error message here?

merkle_tree/src/internal.rs
```diff
@@ -131,52 +128,76 @@ impl<T: NodeValue> MerkleCommitment<T> for MerkleTreeCommitment<T> {
...
         }
+    } else {
```
ggutoski, line 197: nit: in order to avoid excessive indentation consider coding like this:

```rust
if (error condition) {
  return error;
}

// remaining code
```

merkle_tree/src/internal.rs
```diff
@@ -131,52 +128,76 @@ impl<T: NodeValue> MerkleCommitment<T> for MerkleTreeCommitment<T> {
...
+            .try_fold(
+                init,
+                |val, (branch, values)| -> Result<T, MerkleTreeError> {
+                    if values.len() == 0 {
+                        Ok(T::default())
+                    } else {
+                        let mut data = values.clone();
+                        data.insert(*branch, val);
+                        H::digest(&data)
+                    }
```
ggutoski, line 189: Took me a while to figure out what's going on here. Perhaps some comments would help. eg.
- `values` is an arity-size list of hashes comprising the present node of the merkle tree
- `branch` is an integer index into that list indicating where in the arity-size list of hashes the current hash (`val`) should be inserted.
- It's unfortunate that you do an arity-size clone + insert for each step on the merkle path. Lots of touching the heap here. I suppose it's not a bottleneck because arity is always small.

merkle_tree/src/internal.rs
mrain: `VerificationResult = Result<(), ()>` so no.

merkle_tree/src/internal.rs
mrain: Changed in [11e0f4d](https://github.com/EspressoSystems/jellyfish/pull/692/commits/11e0f4debbc8142925649d187f24fd0703c95934)

No heap allocation now. Just a bunch of memory copies.

merkle_tree/src/internal.rs
mrain: Chaged in [11e0f4d](https://github.com/EspressoSystems/jellyfish/pull/692/commits/11e0f4debbc8142925649d187f24fd0703c95934)

merkle_tree/src/internal.rs
```diff
@@ -131,52 +128,76 @@ impl<T: NodeValue> MerkleCommitment<T> for MerkleTreeCommitment<T> {
...
-        &self.pos
-    }
-
-    /// Return the element associated with this `MerkleProof`. None if it's a
-    /// non-membership proof.
-    pub fn elem(&self) -> Option<&E> {
-        match self.proof.first() {
-            Some(MerkleNode::Leaf { elem, .. }) => Some(elem),
-            _ => None,
+    if proof.len() == commitment.height() {
```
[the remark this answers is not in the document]
mrain: I mean the height check may not be necessary now.

merkle_tree/src/internal.rs
ggutoski: Right! I got confused by another error earlier in this function:
```rust
return Err(MerkleTreeError::InconsistentStructureError(
            "Inconsistent Merkle tree height.".to_string(),
        ));
```
but of course the return type is a nested `Result`. 😳

alxiong: nit: we should use type alias to help readability here and save from typing this cryptic type everywhere.

alxiong (commented):
I agree with the trait changes.

- Why are we completely removing `namespaced_merkle_tree`?

mrain:
> * Why are we completely removing `namespaced_merkle_tree`?

Because it's no longer in use and takes too much efforts to refactor.

merkle_tree/src/internal.rs
```diff
@@ -390,162 +411,157 @@ where
...
-                    elem: elem.clone(),
-                }];
-                (
-                    Arc::new(MerkleNode::ForgettenSubtree { value: *value }),
-                    LookupResult::Ok(elem, proof),
-                )
-            },
+            MerkleNode::Leaf { value, pos, elem } => (
+                Arc::new(MerkleNode::ForgottenSubtree { value: *value }),
+                LookupResult::Ok(elem.clone(), MerkleTreeProof(vec![])),
```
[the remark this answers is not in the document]
mrain: It didn't implement `Default`

merkle_tree/src/light_weight.rs
```diff
@@ -110,7 +110,7 @@ where
 #[cfg(test)]
 mod mt_tests {
     use crate::{
-        internal::{MerkleNode, MerkleProof},
+        internal::{MerkleNode, MerkleTreeProof},
```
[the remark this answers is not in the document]
mrain: [808d83c](https://github.com/EspressoSystems/jellyfish/pull/692/commits/808d83c66161bccea6a3454cd1dff844adb53983):  interesting that clippy didn't complain.

merkle_tree/src/internal.rs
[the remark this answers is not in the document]
ggutoski: My understanding of your comments is: now that the caller supplies the element data, there is no need to check tree height because such a check is necessary only if we assume the caller can find hash collisions. This makes sense to me. We should delete the check.

In fact, I see no reason why `commitment` needs to be a `MerkleTreeCommitment`. Really all we need is the hash digest, so we should take only that as an arg instead of `MerkleTreeCommitment`.

alxiong:
>>   Why are we completely removing namespaced_merkle_tree?
>
> Because it's no longer in use and takes too much efforts to refactor.

wait, so we are not going to use them? are we sticking to using a normal MT with a Namespace table to delineate the boundary of different namespaces? 
I thought might be useful to keep a NMT, if the effort is high, we can do it in a separate PR?

ggutoski:
> > wait, so we are not going to use them? are we sticking to using a normal MT with a Namespace table to delineate the boundary of different namespaces? I thought might be useful to keep a NMT, if the effort is high, we can do it in a separate PR?
> 
> No we are not. The namespace table now is just a plain vector with offsets @ggutoski
> 
> I think it's better to bring back NMT on request.

I was not even aware we have a NMT implementation. 😳 We don't need one IMO. We can revive it later if we find that we do need NMT.

merkle_tree/src/internal.rs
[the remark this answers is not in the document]
mrain: done in [e2e5c33](https://github.com/EspressoSystems/jellyfish/pull/692/commits/e2e5c3377fe094cf2dc84dd004ea1b53a6a9e7f4)

mrain:
Also do you think we should merge #685 together? 
cc @alxiong @ggutoski

merkle_tree/src/internal.rs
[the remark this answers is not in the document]
ggutoski: wow so much deleted code! 🎉 🥳

alxiong:
> Also do you think we should merge https://github.com/EspressoSystems/jellyfish/pull/685 together?

we should keep them as two PRs, but yeah, we can rebase that PR after this is merge and then got that one in as well.

ggutoski (approved):

PR #711, opened 2024-12-12
chore: pin the nightly toolchain

mrain:
closes: #XXXX
`cargo +nightly check -p jf-pcs` runs forever on cargo nightly version >1.84.0

### This PR: 
Pin to rust nightly toolchain `nightly-2024-06-25`

### Key places to review:

 

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,7 +37,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: nightly
+          toolchain: nightly-2024-06-25
           override: false
           default: false
           components: rustfmt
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -22,34 +22,25 @@ jobs:
       - name: Checkout Repository
         uses: actions/checkout@v4
 
-      - name: Install stable toolchain
+      - name: Install nightly toolchain
         uses: actions-rs/toolchain@v1
         with:
-          profile: default
-          toolchain: stable
```
alxiong, line 29: i still think we should keep both. 
stable for crates, nightly for format and unstable options.

mrain: but the `rust-toolchain.toml` just override it

alxiong: does it support multiple toolchains?

alxiong: can we pls use `cargo +nightly-YYYY-MM-DD fmt` instead of using `rust-toolchain` and forcing the compiler of normal code to be nightly. We really only use nightly for fmt and for unstable option of report time in CI. 
(thanks for exploring this possibility tho)

imo, it's bad to force jellyfish user to be on nightly. (some downstream repo only support stable rust)

alxiong (approved):

PR #760, opened 2025-03-07
Fix incorrect method signatures in `Vrf` trait

vipocenka:
This PR fixes method signatures in the `Vrf` trait by correcting unnecessary `&mut self` usage and ensuring consistency across functions.

### Changes  
- Removed `&self` from `param_gen`, making it a static method.  
- Changed `proof_to_hash` and `evaluate` to use `&self` instead of `&mut self`.  
- Fixed `verify` to use `&self` instead of `&mut self`.  
- Cleaned up redundant `&mut self` in method definitions.  

### Why?  
These changes improve API consistency and eliminate unnecessary mutability constraints, making the `Vrf` trait easier to use and integrate.

### Key places to review  
- `vrf/src/lib.rs`, method signatures in the `Vrf` trait.  

### Closes  
Closes #XXXX (replace with issue number if applicable)

### DIFF
--- a/vrf/src/lib.rs
+++ b/vrf/src/lib.rs
@@ -61,7 +61,6 @@ pub trait Vrf {
     //   `S::param_gen::<StdRng>(None)`
     // where `StdRng` is redundant.
     fn param_gen<R: CryptoRng + RngCore>(
-        &self,
         prng: Option<&mut R>,
     ) -> Result<Self::PublicParameter, Self::Error>;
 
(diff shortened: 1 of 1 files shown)


vipocenka:
Hi @alxiong, @mrain

Just a friendly reminder about PR #760: it fixes the `Vrf` trait method signatures by:

- Removing unnecessary `&mut self` from `param_gen`, making it a static method  
- Updating `proof_to_hash`, `evaluate`, and `verify` to use `&self` instead of `&mut self`  
- Cleaning up redundant mutability in the trait definition  

All checks have passed except the Conventional Commit title lint. Could someone with write access please review and approve (and update the PR title to match the spec)? Thanks! 🙏

alxiong:
It obviously won't compile, you only change the trait and never update the implementation, the CI didn't really run the build and test (don't know why).

You can run cargo test --workspace and your branch surely will err.

What I don't get, is do you have a specific use case or an instantiation that needs a &self and find &mut self completely unacceptable? If so, I'll reconsider the design.

(Our BLS VRF clearly show a case where &mut is warranted, sure you can say let's change to interior mutability, but trait definition is supposed to be generalization of instantiations, and we defined it by abstracting out from our only inplementor. Until shown insufficient with more implementations of VRF, I don't see a strong case changing it)

vipocenka:
Hi @alxiong,

I’m not confident enough to implement the necessary changes to the trait implementations right now, so I’m going to close this PR. I plan to revisit it later when I have a better understanding of the codebase and Rust’s interior mutability patterns.

Thanks for your patience and all the feedback!

PR #774, opened 2025-04-11
fix(merkle-tree): add domain separation for merkle tree hash

mrain:
closes: #772 

### Key places to review:

 

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/merkle_tree/src/prelude.rs
+++ b/merkle_tree/src/prelude.rs
@@ -36,26 +36,30 @@ pub struct RescueHash<F: RescueParameter> {
     phantom_f: PhantomData<F>,
 }
 
-fn rescue_hash_leaf_domain_separator<F: PrimeField>() -> F {
+/// domain separator of algebraic hash, for the leaf node
+fn leaf_hash_dom_sep<F: PrimeField>() -> F {
     F::one()
 }
-fn rescue_hash_internal_domain_separator<F: PrimeField>() -> F {
+
+/// domain separator of algebraic hash, for the internal node
+fn internal_hash_dom_sep<F: PrimeField>() -> F {
     F::zero()
 }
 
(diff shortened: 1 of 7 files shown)


merkle_tree/src/prelude.rs
```diff
@@ -65,13 +68,19 @@ where
...
         Ok(FixedLenPoseidon2Hash::<F, S, INPUT_SIZE, 1>::evaluate(input)?[0])
     }
 
     fn digest_leaf(pos: &I, elem: &F) -> Result<F, MerkleTreeError> {
-        let mut input = [F::default(); INPUT_SIZE];
+        let mut input = vec![F::zero(); INPUT_SIZE + 1];
+        input[0] = F::one();
         input[INPUT_SIZE - 1] = F::from(pos.clone());
         input[INPUT_SIZE - 2] = *elem;
         Ok(FixedLenPoseidon2Hash::<F, S, INPUT_SIZE, 1>::evaluate(input)?[0])
```
alxiong, line 86: i don't understand. how can we not increase the input size to the hash function? why not `FixedLenPoseidon2Hash::<F, S, INPUT_SIZE + 1, 1>::evaluate()`

alxiong (changes requested):
- instead of directly plug in 0 and 1 as the domain separator, i'd like to mark them as constant. like `const LEAF_DOM_SEP: F = F::one()`. it will be clearer.
- can we add a unit test for this extension attack?

merkle_tree/src/prelude.rs
mrain: "generic parameters may not be used in const operations\nconst parameters may only be used as standalone arguments, i.e. `INPUT_SIZE`"

alxiong (approved):
In principle approval, @mrain please double check

PR #784, opened 2025-05-13
[Proposal] Integrate Fuzzing into Jellyfish

jparr721:
Howdy gents. 

Tobias and I have had some initial luck with getting libFuzzer integrated and want to spend some time looking at this further (i.e. a more comprehensive test harness, etc). We wanted to solicit your feedback as this would be a lot of work, and if there's no hope of it ever merging we might consider some other approaches.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-members = ["aead", "commitment", "crhf", "elgamal", "merkle_tree", "pcs", "plonk", "poseidon2", "prf", "relation", "rescue", "signature", "utilities", "vid", "vrf"]
+members = ["aead", "commitment", "crhf", "elgamal", "fuzz", "merkle_tree", "pcs", "plonk", "poseidon2", "prf", "relation", "rescue", "signature", "utilities", "vid", "vrf"]
 resolver = "2"
 
 [workspace.package]
--- a/fuzz/.gitignore
+++ b/fuzz/.gitignore
@@ -1,0 +1,4 @@
+target
+corpus
+artifacts
+coverage
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -11,12 +11,25 @@ cargo-fuzz = true
 arbitrary = { version = "1.4.1", features = ["derive"] }
 libfuzzer-sys = "0.4"
 jf-merkle-tree = { path = "../merkle_tree" }
+jf-rescue = { path = "../rescue" }
 sha2 = "0.10"
 rand = "0.9.1"
+num-bigint = { workspace = true }
+ark-bls12-377 = { version = "0.4.0", default-features = false, features = [ "curve" ] }
+ark-bls12-381 = { version = "0.4.0", default-features = false, features = [ "curve" ] }
+ark-bn254 = { version = "0.4.0", default-features = false, features = [ "curve" ] }
+hashbrown = "0.14.3"
 
 [[bin]]
 name = "fuzz_merkle_tree"
 path = "fuzz_targets/fuzz_merkle_tree.rs"
+test = false
+doc = false
+bench = false
+
+[[bin]]
+name = "fuzz_universal_merkle_tree"
+path = "fuzz_targets/fuzz_universal_merkle_tree.rs"
 test = false
 doc = false
 bench = false
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+name = "jf-fuzz"
+version = "0.1.0"
+publish = false
+edition = "2021"
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+arbitrary = { version = "1.4.1", features = ["derive"] }
+libfuzzer-sys = "0.4"
+jf-merkle-tree = { path = "../merkle_tree" }
+sha2 = "0.10"
+rand = "0.9.1"
+
+[[bin]]
+name = "fuzz_merkle_tree"
+path = "fuzz_targets/fuzz_merkle_tree.rs"
+test = false
+doc = false
+bench = false
--- a/fuzz/fuzz_targets/fuzz_merkle_tree.rs
+++ b/fuzz/fuzz_targets/fuzz_merkle_tree.rs
@@ -1,20 +1,22 @@
 #![no_main]
 
 use arbitrary::{Arbitrary, Unstructured};
-use jf_merkle_tree::{hasher::HasherMerkleTree, MerkleTreeScheme};
+use jf_merkle_tree::{
+    hasher::HasherMerkleTree, universal_merkle_tree::UniversalMerkleTree, MerkleTreeScheme,
+};
 use libfuzzer_sys::fuzz_target;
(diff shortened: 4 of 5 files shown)


alxiong:
I'm highly supportive on this effort. I'd say add fuzzing in CI as well.

alxiong:
okay, i fixed the rayon iterator issue in CI. but wasm32 target build will fail. I feel like including `fuzz` as a workspace crate is causing too much trouble, may i suggest we remove it? 
instead run fuzz by `cd fuzz && cargo fuzz run xxx`? 
we can document this in README or add a `justfile` for it? 

wdyt? @mrain

mrain:
> okay, i fixed the rayon iterator issue in CI. but wasm32 target build will fail. I feel like including `fuzz` as a workspace crate is causing too much trouble, may i suggest we remove it? instead run fuzz by `cd fuzz && cargo fuzz run xxx`? we can document this in README or add a `justfile` for it?
> 
> wdyt?

That's a good idea. We don't mean to export it.

Also don't forget to update the nix env

jparr721:
CI fixed. @0xkato to take over adding tests, I will circle back

fuzz/fuzz_targets/mt/fuzz_extension_attack.rs
```diff
@@ -0,0 +1,51 @@
...
+    test_extension_attack_helper::<Fr254>(&mut u);
+    test_extension_attack_helper::<Fr377>(&mut u);
+    test_extension_attack_helper::<Fr381>(&mut u);
+});
+
+fn test_extension_attack_helper<F: RescueParameter>(u: &mut Unstructured) {
+    if let Ok(input) = ExtensionAttackInput::arbitrary(&mut u) {
+        let forged_val = F::from(input.forged_val);
+        let attack_pos = input.attack_pos;
+        let forged_pos = attack_pos * 3 + 2;
```
alxiong, line 31: instead of using hardcoded value, let's either add explanatory code doc on top of this whole function or name `const arity = 3` 

especially explain the `digest_leaf` computation on `H(0 || pos || val)` where `0` is leaf_node_domain_separator... to help readers understand why `+2`

fuzz/fuzz_targets/mt/fuzz_extension_attack.rs
```diff
@@ -0,0 +1,51 @@
...
+
+        let data = [F::zero(), F::from(forged_pos), forged_val];
+        let squeezed = RescueCRHF::<F>::sponge_no_padding(&data, 1).unwrap();
+        let val = squeezed[0];
+
+        let elems = vec![val; attack_pos as usize + 1];
```
alxiong, line 37: should we:
1. construct this leaves vector with `Vec::with_capacity(next_multiple_of(attack_pos, arity))`
2. only mutate the `attack_pos` to be the `val`
3. add some sanity check on the constructed `mt`'s `height`, `num_leaves` etc.

fuzz/fuzz_targets/mt/fuzz_light_mt_lookup.rs
```diff
@@ -0,0 +1,46 @@
...
+
+    if let Ok(input) = MerkleTreeArbitraryInput::arbitrary(&mut unstructured) {
+       let elems: Vec<Fr254> = input.elems.iter().map(|x| Fr254::from(*x)).collect();
+       if let Ok(mt) = RescueLightWeightMerkleTree::<Fr254>::from_elems(None, elems) {
+
+        if let Some(random_lookup) = (0..input.elems.len())
+            .into_iter()
+            .collect::<Vec<_>>()
+            .choose(&mut thread_rng())
+            .cloned()
```
alxiong, line 33: is this an overly-complicated way of doing `rng.gen_range(0..len)`?

fuzz/fuzz_targets/mt/fuzz_light_mt_lookup.rs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            let commitment = mt.commitment();
+            if let jf_merkle_tree::LookupResult::Ok(elem, proof) = mt.lookup(random_lookup as u64) {
```
alxiong, line 36: Q1: i can't think of a reason why the result here won't be `LookupResult::Ok`, we didn't prune it or explicitly forget it. 
Q2: putting it behind `if let` indicate that we want to skip the failure cases? should we panic in those case instead. I think we should `?` to panic.

fuzz/fuzz_targets/mt/fuzz_merkle_tree.rs
```diff
@@ -0,0 +1,40 @@
...
+use jf_merkle_tree::{
+    hasher::HasherMerkleTree, universal_merkle_tree::UniversalMerkleTree, MerkleTreeScheme,
+};
+use libfuzzer_sys::fuzz_target;
+use rand::seq::IndexedRandom;
+use sha2::Sha256;
+
+#[derive(Arbitrary, Debug)]
+struct MerkleTreeArbitraryInput {
+    // height: usize,
```
alxiong, line 13: why not?

fuzz/fuzz_targets/mt/fuzz_merkle_tree.rs
```diff
@@ -0,0 +1,40 @@
...
+fuzz_target!(|data: &[u8]| {
+    let mut unstructured = Unstructured::new(data);
+    if let Ok(input) = MerkleTreeArbitraryInput::arbitrary(&mut unstructured) {
+        let mt = HasherMerkleTree::<Sha256, usize>::from_elems(None, input.elems.clone()).unwrap();
+        let input_elems_len = input.elems.len();
+        if let Some(random_lookup) = (0..input_elems_len)
+            .into_iter()
+            .collect::<Vec<_>>()
+            .choose(&mut rand::rng())
+            .cloned()
```
alxiong, line 26: same here, why not `rand::rng().random_rang()`?

fuzz/fuzz_targets/mt/fuzz_test_light_mt_serde.rs
```diff
@@ -0,0 +1,48 @@
...
+}
+
+fuzz_target!(|data: &[u8]| {
+    let mut unstructured = Unstructured::new(data);
+    if let Ok(input) = MerkleTreeArbitraryInput::arbitrary(&mut unstructured) {
+        if input.elems.is_empty() {
+            return;
+        }
+
+        let elems: Vec<Fr254> = input.elems.iter().map(|x| Fr254::from(*x)).collect();
```
alxiong, line 26: importing `ark_std::UniformRand`, then you can invoke `Fr254::rand(&mut rng)` which will produce a uniform field element (256bits) rather than converting from a single limb u64. 

we can use arbitrary input's as seed to a SeedableRng which is further used to generate a random field element if Arbtrary doesn't work well with `Fr::rand()` api.

fuzz/fuzz_targets/mt/fuzz_universal_merkle_tree.rs
alxiong, line 1: so what are we testing here? just universal tree construction? or maybe this is just WIP?

PR #806, opened 2025-08-09
Fix logical error in VerifyingKey merge operation

anim001k:
### Description

Fixed a logical inconsistency in `VerifyingKey::merge()` where the method first validates that both keys have equal `num_inputs`, but then incorrectly sums them in the result.

### DIFF
--- a/plonk/src/proof_system/structs.rs
+++ b/plonk/src/proof_system/structs.rs
@@ -808,7 +808,7 @@ impl<E: Pairing> VerifyingKey<E> {
 
         Ok(Self {
             domain_size: self.domain_size,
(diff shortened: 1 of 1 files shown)


alxiong:
I understand your initial intent, but you misunderstood our intended usage. 
We first create a "mergeable circuit" which is already twice as large, before invoking this `merge()` method, thus the logic is correct. 

A usage is seen at https://github.com/EspressoSystems/jellyfish/blob/4ba71c5d381667ab7c02e64537494502b3e582ea/plonk/src/proof_system/snark.rs#L1630-L1642

and for more details, you can see Sec 3.2 of https://eprint.iacr.org/2022/802.pdf#subsection.3.2

Issue #805, opened 2025-08-08
Use commitment to verify proofs (patch `v0.4.5`)


sveitser:
Currently we use the digest and forgo a check for the proof height.

Release new version and make sure it's integrated into espresso-network.

PR #816, opened 2025-08-14
refactor!: merkle proof verify accept commitment instead of digest node

alxiong:
Closes: #805 

### This PR: 

- update the `MerkleTreeScheme` trait so that the `.verify()` function accept the commitment instead of digest root node, and we check the proof length against the `commitment.height()` to prevent extension 
- update the implementation and unit tests to use the new api
- update `vid` crate to use the new api

### This PR does not:

- [ ] tag a new version to both `jf-merkle-tree` and `jf-vid` + update CHANGELOG?

---

write a little note why.

- [x] Targeted PR against correct branch (`release-*`)
- [x] Wrote unit tests

### DIFF
--- a/merkle_tree/src/namespaced_merkle_tree/proof.rs
+++ b/merkle_tree/src/namespaced_merkle_tree/proof.rs
@@ -9,8 +9,9 @@ use super::{
     hash::NamespacedHash, BindNamespace, Element, InnerTree, Namespace, NamespaceProof, Namespaced,
 };
 use crate::{
-    errors::MerkleTreeError, internal::MerkleProof, DigestAlgorithm, MerkleTreeScheme, NodeValue,
-    VerificationResult,
+    errors::MerkleTreeError,
+    internal::{MerkleProof, MerkleTreeCommitment},
+    DigestAlgorithm, MerkleTreeScheme, NodeValue, VerificationResult,
 };
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 use ark_std::{string::ToString, vec::Vec};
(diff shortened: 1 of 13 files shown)


merkle_tree/src/namespaced_merkle_tree/proof.rs
```diff
@@ -108,8 +109,17 @@ where
...
             }
             // Verify the boundary proof
-            if <InnerTree<E, H, T, N, ARITY>>::verify(root, boundary_proof.index(), boundary_proof)?
-                .is_err()
+            // Create a commitment from the root hash with proper height calculation
+            let proof_height = boundary_proof.proof.len().checked_sub(1).ok_or_else(|| {
+                MerkleTreeError::InconsistentStructureError("Empty boundary proof".to_string())
+            })?;
+            let commitment = MerkleTreeCommitment::new(*root, proof_height, 0); // size won't be checked
+            if <InnerTree<E, H, T, N, ARITY>>::verify(
```
sveitser, line 117: I don't quite understand this code but it seems to be in line with how things work now.

sveitser (approved):
LGTM. This is a breaking API change though. I'm fine with that for espresso-network but in case other parties use this version of jf this isn't ideal. Not sure how we could avoid it but semver doesn't really work if we have to make breaking changes to release branches. >.<

merkle_tree/src/namespaced_merkle_tree/proof.rs
alxiong: So basically we used to accept root node, now we only accept commitment, so we construct a commitment on the fly with num_leaves left unfilled since we don't need it for verification. 

It looks slightly hacky, I agree. But I don't want to expose another  "_unchecked" or special api for now.

PR #823, opened 2025-09-02
fix: missing non-canonical check in unpack gadget

alxiong:
Shout out to Least Authority for responsible disclosure. 

### This PR: 

See `non_canonical_field()` in unit test for demonstration of a possible attack. 

After discussing with @mrain, even though my current fix will lead to the cost of `cmp` gates to double, without lookup gates in turboplonk, there's not too much we can improve. 

We do plan to add new set of APIs (in a future release) with the assumption that wire variables are already in the `[0, p/2)` range, by only decomposing `MODULUS_BIT_SIZE-1` bits, (directly truncating the MSB as it is assumed to be zero). This will simplify lots of our comparison logic, and avoid the cost of `canonicalization_check` sub-circuit. 

We believe these new APIs are useful because in many scenarios, the values we need to compute and sum and ensure less than some threshold, these values are almost always `< (p-1)/2`. Such as the total voting weights in our light client circuit. 

We think the best course of action is to patch the vulnerability and release a new tag (`jf-relation-v0.5.0`) ASAP, bump all dependencies. Leave the new APIs for the next non-breaking minor bump. 

cc @chancharles92 @philippecamacho @0xkato 

---

write a little note why.

- [x] Wrote unit tests


mrain (approved):

chancharles92 (approved):

mrain:
We are still waiting for arkworks v0.5.0 to resolve RUSTSEC-2025-0055, see https://github.com/EspressoSystems/jellyfish/pull/827#issuecomment-3253810297

Given that we are not using `tracing-subscriber` to log anything here, I feel like we should ignore the cargo audit warning and proceed first. Will come back to fix it once arkworks updates their crates.

WDYT cc @alxiong @philippecamacho @akonring @sveitser

mrain:
@alxiong let's move forward and temporarily ignore RUSTSEC-2025-0055 here. We are not using tracing-subscribers on hotshot/sequencer/prover. Let's fix those issues first.

alxiong:
Agree, let's merge this and issue a new tag by EOD

PR #839, opened 2025-10-01
chore!: revert the `spongefish` migration for `jf-rescue`

mrain:
closes: #XXXX

### Key places to review:

 

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,0 +1,9 @@
+[advisories]
+ignore = [
+    # WARNING: crate `derivative` unmaintained.
+    "RUSTSEC-2024-0388",
+    # WARNING: crate `paste` unmaintained.
+    "RUSTSEC-2024-0436",
+    # `tracing-subscriber` doesn't rule out poisoning logs. But we are not using it.
+    "RUSTSEC-2025-0055",
+]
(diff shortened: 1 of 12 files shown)


alxiong (approved):
LGTM

alxiong:
Today I gave it a second thought, 
should we really revert on `main` for `jf-rescue`? why couldn't `jf-relation-v0.5.0-compat` simply depends on an older version of `jf-resuce` (before the spongefish update?) patched with ark0.5 cherry-picked over? we can release as `jf-rescue-v0.2.0-patch1` and use it for backward-compat purposes. 

I think the decision should come down to whether in the long term, do we want to switch from state XOR mode to overwrite mode (as `spongefish`'s `DuplexSponge` uses).

> Specifically, we consider the duplex sponge in overwrite mode rather than canonical duplex sponges “in XOR mode” as it is simpler. Our analysis can be adapted for related constructions such as XOR mode, and different padding functions.
>
> footnote 4 of https://eprint.iacr.org/2025/536.pdf

given that we have gone through an external audit with the XORing mode, it should be fine to keep using that. The only downside is that we don't get the same provable guarantees from the spongefish paper. 

In short, just documenting my thinking process, no further action required. For now, I think our current revert is fine.

PR #891, opened 2026-06-26
feat: make bls12-381 curve optional

jbearer:
This adds a feature flag (default off) to opt into compilation of the bls12-381 curve. This curve requires a dependency on the `blst` crate. Due to the way this crate binds to compiled C and assembly, it cannot be built on certain targets (in particular, I'm targetting RISC-V for use with Succinct Labs' SP1 VM). With this flag, it is possible to compile the `jf-signature` crate without this dependency, and it now compiles using SP1.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,7 +52,7 @@ jobs:
           token: ${{ github.token }}
 
       - name: Check Bench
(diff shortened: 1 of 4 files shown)


mrain (dismissed):
LGTM. nit: we could make `bls12-381` feature default, as suggested by gemini.

mrain (approved):

